PR #4, opened 2015-12-18
Fix minor doc error

quicksnap:
Another slight change. This is correct, yes?


quicksnap:
Also, I'm really grateful @gaearon tweeted about this library, because it's really well done! Thanks so much. Very exciting.

yelouafi:
Thanks again! Glad you like it :)

Issue #9, opened 2015-12-22
Hot reducer loading is broken in examples/real-world


billyjanitsch:
Repro: start the server, change the [default error message](https://github.com/yelouafi/redux-saga/blob/master/examples/real-world/reducers/index.js#L16) to 'test'. Redux complains about an invalid reducer.

The problem is that Babel 6 changed how default exports are stored:

```js
// a.js
export default 'hello'

// b.js, Babel 5
const hi = require('./a')
assert(hi === 'hello')

// b.js, Babel 6
const hi = require('./a')
assert(hi !== 'hello')
assert(hi.default === 'hello')
```

So you need to add `.default` [here](https://github.com/yelouafi/redux-saga/blob/master/examples/real-world/store/configureStore.dev.js#L19) in order to require the reducer itself.

gaearon:
Would you like to send a PR? ;-)

billyjanitsch:
I elected not to because:

1. Mixing webpack-resolved `require`s with Babel-transformed `export`s is kind of a hack to get around lack of support for conditional imports. My proposed solution requires knowledge of Babel internals and could break in future versions; maybe there's a more robust way to solve the problem.
2. Given this fragility, I'm not sure how much documentation @yelouafi would want along with a fix.

This is the sort of situation where, as a lib author, I'd prefer if someone just pointed out the problem and let me figure out the best solution. However, if you're content with my proposal and don't think anything like a `// HACK` is necessary, I'll gladly submit a PR. :)

gaearon:
It's not going to break. Babel 6 broke this on purpose to align with ES6 semantics. It won't change further.

billyjanitsch:
Didn't realize that, thanks for the clarification! :) PR: #11

yelouafi:
Thanks!

PR #11, opened 2015-12-23
Fix hot reducer loading in examples/real-world

billyjanitsch:
As raised in #9, hot reducer loading is currently broken. The problem is that Babel 6 changed how default exports are stored:

```js
// a.js
export default 'hello'

// b.js, Babel 5
const hi = require('./a')
assert(hi === 'hello')

// b.js, Babel 6
const hi = require('./a')
assert(hi !== 'hello')
assert(hi.default === 'hello')
```


yelouafi:
Thans!

PR #12, opened 2015-12-23
fix is.generator test

### LINKED ISSUE #10: redux-saga + webpack + uglifyjs: Saga must be a Generator function
First I would like to say that I am exiting about redux-saga.

I faced problem with redux-saga + webpack + uglifyjs combination in production build. For some reason sagas are "too well" uglified and don't pass the check utils.is.generator(saga). In result it throws and an error "Saga must be a Generator function"

I found a workaround. Just simple use such options:
new webpack.optimize.UglifyJsPlugin({
        mangle: {except: ['GeneratorFunction'] }
})

Hope it helps someone. Is there any better way to avoid such errors?

yelouafi:
fix issue with minified bundles (regenerator runtime's Generator function names got mangled by uglify)

should fix #10

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -7,12 +7,15 @@ export function check(value, predicate, error) {
     throw new Error(error)
 }
 
+function* sampleGen() {}
+const genConstructor = sampleGen.constructor
+
 export const is = {
   undef     : v => v === null || v === undefined,
   func      : f => typeof f === 'function',
   array     : Array.isArray,
   promise   : p => p && typeof p.then === 'function',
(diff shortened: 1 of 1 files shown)


yelouafi:
>Can we just call the function and assert that it returns something with .next()?

Yes. I think this is the best thing we can do

yelouafi:
@bgerm I released a new patch version on npm with the fix suggested by @gaearon. Could you try and confirm if it's ok now ?

bgerm:
Thanks, @yelouafi.  Works great.

yelouafi:
Great! I'll close the issue. Thanks!

PR #17, opened 2015-12-25
support cancelling child subroutines when parent routine is cancelled

aikoven:


yelouafi:
thanks @aikoven for your contribution.

For subroutine cancellation I think auto cancellation of child subroutines isn't a good idea. IMO it'd be better to let the developer explicitly have control of when to trigger the cancellation.

```javascript
function* subtask2() {
    try { ... } 
    catch(err) { 
       doCleanup 
    }
}

function* subtask() {
   const subtask = yield fork(subtask2) 
   ...
    try {
              ...
    } catch(err) {
      needToDoBeforeChildcancellation()
      subtask.cancel() // explicit cancellation of subtasks
    } 
}

function mainTask() {
   const task = yield fork(subtask)
   ...
   task.cancel()
}
```

(As a side note I think your code doesnt handle the case if a Saga forked more than one subroutine)

I think also we should also throw a more specific error, something like `SagaCancellationError`. This way we can know if the catch block was triggered by a cancellation or other error

aikoven:
That's a different case: I covered subroutines that were invoked using `call`, not `fork`. I agree that forked subtasks should be cancelled by hand, because they live in separate 'process'. But if `call`ed subroutine is in progress when parent routine is cancelled, then there is no way to actually cancel subroutine.

aikoven:
Also it seems to be nontrivial to make a custom `Error` extension, see http://stackoverflow.com/questions/1382107/whats-a-good-way-to-extend-error-in-javascript
On the other hand we don't need stack trace for these errors, so we can probably go with just plain class.

aikoven:
Just noticed that my code actually cancels `fork`ed subroutines, not only `call`ed. Will fix

yelouafi:
>But if called subroutine is in progress when parent routine is cancelled, then there is no way to actually cancel subroutine.

Thanks for drawing my attention on this. I was too focused on forked tasks.

There are some more issues i'd like to discuss

1- taking into account that a task can be blocked on multiple `call`s. for example

```javascript
function* task() {
  const [r1, r2] = yield [call(subtask1), call(subtask2)]
}
```

we need to cancel both subtasks if `task` is cancelled while waiting for the 2 subtasks to complete.

2- making task cancellation declarative. so instead of `task.cancel()` we write `yield cancel(task)`. I think it makes more sens since `fork` and `join` are already declaratives

3- Do you think we should automatically cancel `call`ed subtasks that have lost in a `race`. for example

```javascript
function* task() {
  ...
   const {timeout} = yield race({
      timeout: call(delay, 1000),
      subtask: call(subtask)
    });
}
```
if `timeout` wins the race, should we automatically cancel the `call(subtask)` ?

aikoven:
1. Good point, I will make another commit to fix that
2. I guess that makes sense too
3. In my auth example I have following code:
```js
const authLoopTask = yield fork(authorizeLoop, storedToken);

const {signOutAction} = yield race({
  signOutAction: take(SIGN_OUT),
  authLoop: join(authLoopTask)
});

if (signOutAction) {
  authLoopTask.cancel();
  // ...
}
```
`authLoop` here is an infinite cycle that ends only when cancelled. I use `fork` here, not `call`, but if we did as you propose, it would be shorter:
```js
const {signOutAction} = yield race({
  signOutAction: take(SIGN_OUT),
  authLoop: call(authorizeLoop, storedToken)
});

if (signOutAction) {
  // ...
}
```
So probably it's a good idea, but I need to think a bit more about this.
Maybe it would be convenient to make promise returned from `proc()` cancellable.

aikoven:
@yelouafi Done. 

I decided not to go with cancellable promises because there is no convention and it would result in additional complications with making chained promises cancellable as well.
So I just cancel all subroutine that is still running after race resolves.

yelouafi:
@aikoven great work! thanks :+1: .

My only doubt is on the automatic race cancellation. we should cancel the cancellable/competing effects. I looked quickly, but it seems like you're cancelling subroutines of the Saga that is running the actual effects.

yelouafi:
@aikoven 
Sorry. It seems like your code is correct after all. If the Saga that is running the actual race advances; all pending subroutines should be cancelled as we're no longer blocked on them.

aikoven:
You are right, my code will go wrong if there was another `call`ed subroutine outside of `race`, e.g.:

```js
yield [
  call(subroutine1),
  race({
    subroutine2: call(subroutine2),
    subroutine3: call(subroutine3),
  })
]
```
If race finishes before `subroutine1`, then it will be cancelled

yelouafi:
>You are right, my code will go wrong if there was another called subroutine outside of race, e.g.:

Oops! merged too soon!

aikoven:
I made a fix. I can make another PR, or you can revert the merge 😀

yelouafi:
I reverted the merge. Not sure however if this will reactivate the PR

PR #20, opened 2015-12-26
Add missing npm install

SimonSelg:
The instructions of the 'real word example' where missing an 'npm install'.


yelouafi:
thanks!

PR #26, opened 2016-01-05
[reopened] support cancelling child subroutines when parent routine is cancelled

aikoven:
New PR in place of https://github.com/yelouafi/redux-saga/pull/17

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,7 @@
 import { is, remove } from './utils'
 import proc from './proc'
-export { take, put, race, call, cps, fork, join } from './io'
+export { SagaCancellationException } from './proc';
+export { take, put, race, call, cps, fork, join, cancel } from './io'
 
 export const SAGA_NOT_A_GENERATOR_ERROR = "Saga must be a Generator function"
 
--- a/src/io.js
+++ b/src/io.js
@@ -5,6 +5,7 @@ export const CALL_FUNCTION_ARG_ERROR = "io.call first argument must be a functio
 export const CPS_FUNCTION_ARG_ERROR = "io.cps first argument must be a function"
 export const FORK_ARG_ERROR = "io.fork first argument must be a generator function or an iterator"
 export const JOIN_ARG_ERROR = "io.join argument must be a valid task (a result of io.fork)"
+export const CANCEL_ARG_ERROR = "io.cancel argument must be a valid task (a result of io.fork)"
 
 const IO    = Symbol('IO')
 
@@ -15,6 +16,7 @@ const CALL  = 'CALL'
 const CPS   = 'CPS'
 const FORK  = 'FORK'
 const JOIN  = 'JOIN'
+const CANCEL  = 'CANCEL'
 
 const effect = (type, payload) => ({ [IO]: true, [type]: payload })
 
@@ -67,6 +69,13 @@ export function join(taskDesc) {
   return effect(JOIN, taskDesc)
 }
 
+export function cancel(taskDesc) {
+  if(!taskDesc[TASK])
+    throw new Error(CANCEL_ARG_ERROR)
+
+  return effect(CANCEL, taskDesc)
+}
+
 export const as = {
   take  : effect => effect && effect[IO] && effect[TAKE],
   put   : effect => effect && effect[IO] && effect[PUT],
@@ -74,5 +83,6 @@ export const as = {
   call  : effect => effect && effect[IO] && effect[CALL],
   cps   : effect => effect && effect[IO] && effect[CPS],
   fork  : effect => effect && effect[IO] && effect[FORK],
-  join  : effect => effect && effect[IO] && effect[JOIN]
+  join  : effect => effect && effect[IO] && effect[JOIN],
+  cancel: effect => effect && effect[IO] && effect[CANCEL]
 }
--- a/src/proc.js
+++ b/src/proc.js
@@ -3,35 +3,38 @@ import { as, matcher } from './io'
 
 export const NOT_ITERATOR_ERROR = "proc first argument must be an iterator"
 
-export class SagaCancellationException {}
+export class SagaCancellationException {
+  constructor(message, stack=[]) {
+    this.message = message
+    this.stack = stack
+  }
+}
+
+export const CANCEL = Symbol('cancel')
 
 export default function proc(iterator, subscribe=()=>()=>{}, dispatch=()=>{}) {
 
   check(iterator, is.iterator, NOT_ITERATOR_ERROR)
 
-  let deferredInput, deferredEnd
+  let deferredInputs = [], deferredEnd
   const canThrow = is.throw(iterator)
 
   const endP = new Promise((resolve, reject) => deferredEnd = {resolve, reject})
 
   const unsubscribe = subscribe(input => {
-    if(deferredInput && deferredInput.match(input))
-      deferredInput.resolve(input)
+    for (let deferredInput of deferredInputs) {
+      if(deferredInput.match(input))
+        deferredInput.resolve(input)
+    }
   })
 
-  const subroutines = []
-
   iterator._isRunning = true
-  iterator._cancel = cancel
   next()
 
   return endP
 
   function next(arg, isError) {
-    if (!iterator._isRunning)
-      return
     //console.log('next', arg, isError)
-    deferredInput = null
     try {
       if(isError && !canThrow)
         throw arg
@@ -39,7 +42,9 @@ export default function proc(iterator, subscribe=()=>()=>{}, dispatch=()=>{}) {
 
       if(!result.done) {
         //console.log('yield', name, result.value)
-        runEffect(result.value).then(next, err => next(err, true))
+        const currentEffect = runEffect(result.value)
+        endP[CANCEL] = currentEffect[CANCEL]
+        currentEffect.then(next, err => next(err, true))
       } else {
         //console.log('return', name, result.value)
         iterator._isRunning = false
@@ -53,23 +58,23 @@ export default function proc(iterator, subscribe=()=>()=>{}, dispatch=()=>{}) {
       iterator._error = err
       unsubscribe()
       deferredEnd.reject(err)
-    }
-  }
 
-  function cancel(err) {
-    for (let subroutine of subroutines) {
-      subroutine._cancel(err)
+      if (err instanceof SagaCancellationException) {
+        throw err
+      }
     }
-    subroutines.length = 0
-
-    next(err, true)
   }
 
   function runEffect(effect) {
+    let deferred, isRunning = true
+    const ret = new Promise((resolve, reject) => {
+      deferred = {resolve, reject}
+    })
+
     let data
-    return (
-        is.array(effect)           ? Promise.all(effect.map(runEffect))
-      : is.iterator(effect)        ? runSubroutine(effect)
+    const effectPromise = (
+        is.array(effect)           ? runParallelEffect(effect)
+      : is.iterator(effect)        ? proc(effect, subscribe, dispatch)
 
       : (data = as.take(effect))   ? runTakeEffect(data)
       : (data = as.put(effect))    ? runPutEffect(data)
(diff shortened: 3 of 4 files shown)


yelouafi:
Sorry, But I'll have to raise some issues again

- for parallel effects `[call(saga1), call(saga2), effect]`. If `effect` is rejected the whole `[...]` effect will be rejected. The saga will progress (through a try/catch or by throwing to its parent) with the rejected effect alone, se we have also to cancel the other calls inside the array 

- In `race({c1: someEffect, c2: join(task)})`. if c1 wins `c2` should also be cancelled IMO. But the actual code will skip it because the promise result (`task._done`) doesn't define a _cancel method

- It's not only related to subroutines, for example if we have

```javascript
yield race({
  c1: call(saga1),
  c2: [
    call(saga2),
    call(saga3)
  ] 
})
``` 

If `c1` wins the race, we should cancel the 2 calls inside `c2`. But since the promise returned by the parallel effect (`Pormise.all(...)`) doesn't have a `_cancel` method it will be skipped and the inner calls will continue to execute. 

IMO, we're takling the problem from the wrong angle, i.e. we're too focused on the cancellation of sagas. we should think in a more generic way: the cancellation of effects themeselves. Effects are composables. We can exploit this to make cancellation also composables;

Let's start with a simple assumption: all effects are cancellables (e.g. their promise results define a _cancel method). We have to specify 3 things

1- Semantics of cancellation for each effect type
2- Semantics of automatic cancellation
3- Semantics of manual cancellation

## Semantics of cancellation for each effect type

- cancelling raw promises, as well as results of normal function calls (call and cps) is noop (until promises support cancellation, we can simply set `promise._cancel = () = {}` ) 

- cancelling the result of a put effect is noop

- cancelling the result of a take effect can be achieved by simply setting the `deferredInput` to null (we could also support multiple concurrent/parallel takes by allowing an array of deferred inputs).

- cancelling the result of a fork effect is noop

- cancelling the result of a join effect is trivial since the end promise of a forked task should already have a _cancel method, idem for raw iterators (we could set this directly in the body of `proc`)

- cancelling a parallel effect `[...]` means cancelling all sub-effects on the array

- cancelling a race effect means cancelling all the competing sub-effects

- cancelling a saga call result means:  1- cancelling the current effect result where the saga is blocked (could be anything, we don't care), and 2- throwing a SagaCancellationException 

- cancelling the result of cancel effect is noop (we cannot undo the effect of a cancellation)

## Semantics of automatic cancellation

Automatic cancellation can be triggered in tho following cases

- in a race effect, cancel all sub-effects except the winner

- in a parallel effect, if a sub-effect is rejected, cancel all other sub-effects


## Semantics of manual cancellation

- a manual cancellation can only be triggered via `yield cancel(...)`

- `yield cancel` can only be called on tasks created via `fork`

Unless I miss something, I think the above approach will allow us to tackle the problem in a generic way. It's composable so we can relay on recursion to handle all the edge cases

What do you think ?

aikoven:
Sounds very reasonable!

So we'll need a `_cancel` method attached to promise returned from `proc` as well as promise returned from each `run*Effect`. 

We can also make automatic cancellation of parallel / race effects simpler: just cancel the whole parallel / race effect after it completes. Race winner subroutine will already have `_isRunning = false` so `_cancel` becomes noop and we won't have to filter losers from winner.

yelouafi:
>We can also make automatic cancellation of parallel / race effects simpler: just cancel the whole parallel / race effect after it completes. Race winner subroutine will already have _isRunning = false so _cancel becomes noop and we won't have to filter losers from winner.

Yes, that makes things simpler. This adds another rule to the semantics above

- cancelling an already resolved/rejected effect is noop

Perhaps `_isRunning` should be defined in a generic way

```javascript
runEffect(effect) {
  const promise = (...)
  const done = () => promise._isRunning = false

  promise.then(done, done)
  promise._isRunning = true
  return promise
}

cancelEffectResult(promise) {
  if(promise._isRunning)
     promise._cancel()
}
```

in Promise.all/Promise.race blocks, child promises complete always before the parent promise, and their then blocks are also ran before completing the parent; so the order is Ok (done will be executed before completing the parent all/race promise)

aikoven:
One more thing: if we cancel `call`ed routine, then exception will be thrown in it, and if we don't catch it, then it will propagate to parent routine. Should we prevent propagation?

yelouafi:
I think we should not prevent propagation. All cancellables Sagas should catch cancellation exceptions to do any cleanup. If the canceled Saga omit this; then the parent could handle the exception (for example, if one cleanup is relevant to mulitple child called sagas, the parent saga can handle the cancelltion exception).

Otherwise, we should IMO propagate the exception as JavaScript does with normal errors

aikoven:
If we start two subroutines without handlers in parallel, then if first one
fails then the second will be cancelled. Which exception should be pulled
to parent?
Ср, 6 янв. 2016 г. в 21:19, Yassine Elouafi <notifications@github.com>:

aikoven:
I guess we need to define semantics for thrown exceptions.

First of all, the only way to cancel effect is to cancel routine that is currently blocked on it.

Suppose that `main` routine is blocked on `yield call(child)`. Then cancelling `main` effectively cancels `call` effect which in turn throws exception inside `child` routine. But if `child` routine caught `SagaCancellationException` and did not rethrow it, then it won't be thrown in `main` routine. I see two ways here: warn user when they don't rethrow, or just always forcibly propagate exception to parent routine. I would go with latter as less error-prone, but maybe there is a valid case with catching and not rethrowing.

As for parallel/race effects: when they finish, all losers' resolution is ignored, so any cancellation exceptions should be ignored as well.

The remaining case is `join`. I think it should work the same as `call`: cancelling joined subroutine, then propagating to main routine.

yelouafi:
>If we start two subroutines without handlers in parallel, then if first one fails then the second will be cancelled. Which exception should be pulled to parent?

I'd say the synchronous flow of things suggests to throw the cancellation exception, because it was raised while proc is trying to handle the resolution of the parallel effects. for example looks at this simplified code of proc

```javascript
proc(iterator) {

 const endPromise = ...
 endPromise._cancel = cancel

  cancel() {
    next(CancellationException(), true)
  }

  next(arg, isErr) {
    try {
       const result = isErr ? iterator.throw(arg) : iterator.next(arg)
       runEffect(result.value).then(next, err => next(err, true))
   } catch(err) {
      reject(endPromise)
   }
  }

  runParallelEffects(effects) {
    const promises = effects.map(runEffect)
    const result = Promise.all(promises)
    const done = isErr => val => {
      promises.forEach(p => p._cancel())
      return isErr ? Promise.reject(val) : val
    }
    return result.then(done(), done(true))
  }
}
```

the code `result.then(done, done)` will run before resuming the parent promise with the failed `result`. If result fails with some subeffect, the code will run the _cancel method on `promises` consecutively, which will trigger `iterator.throw` on consecutive subroutines (via next(CanellationException, isErr)), if a subroutine in the list omitted to catch the exception caused by iterator.throw then the exception will propagate to the parent (which is running the actual runParallelEffects) and will throw that Cancellation exception inside it. the uncaught exception will then propagate up to next.

Now what if a developer already defined a try/catch block inside a main routine

```javascript
function* main() {
   try {
     yield put(requestStart())
     const results = yield [call(subroutine), someEffect]
     yield put(requestSuccess(results))
   } catch(err) {
     yield put(requestFailure(err))
   }
}
```

if `someEffect` failed, then we will cancel `call(subroutine)`; if subroutine omitted to define cleanup code then we should propagate the exception to warn the developer and abort subroutine as well as main (the same as normal JS errors abort all code up to the top of the call chain).

The issue is that `catch(err)` is catch-all block and will prevent the Cancellation exception from propagating up; In languages like Java we don't have this kind of problem because we can define multiple catches per Exception type, but this is not possible in JS.

So IMO, until JS supports this feature the solution is for the developer to put a guard in his code

```javascript
// reausable guard
function rethrowCancellation(err) {
  if(err instance of CancellationException)
    throw err
}

function* main() {
   try {
     yield put(requestStart())
     const results = yield [call(subroutine), someEffect]
     yield put(requestSuccess(results))
   } catch(err) {
     rethrowCancellation(err)
     yield put(requestFailure(err))
   }
}
```

This kind of explicit solution adds a little burden to the developer, but IMO makes also the code intention more obvious

>Suppose that main routine is blocked on yield call(child). Then cancelling main effectively cancels call effect which in turn throws exception inside child routine. But if child routine caught SagaCancellationException and did not rethrow it, then it won't be thrown in main routine.

But child don't need to rethrow cancellation, If I'm not mistaken, cancelling main involves 2 operations as defined by our semantical model

1- first cancel the current effect we're blocked in
2- throw a CancellationException inside main iterator

if child catch its cancellation, then good. (2) will throw a Cancellation inside main. If not, also good, the exception will propagate to main; if main defines the `rethrowCancellation(err)` guard then it'll propagate and warn the developer (Did I miss something ?)

>As for parallel/race effects: when they finish, all losers' resolution is ignored, so any cancellation exceptions should be ignored as well.

Hmm.. I feel a bit uncomfortable with this silent failing. A missed cancellation catch will likely point to some bug in the program (developer omitted to define cancellation logic). Warning only on the console won't prevent the parent and its parents on the call chain from continuing, while the conventional JS mechanism is to abort all the routines in the call chain. Since w're trying to mimic the normal JS synchronous control flow, shouldn't we do the same with the propagation mechanism and propagate the *first* encountered  exception?

>The remaining case is join. I think it should work the same as call: cancelling joined subroutine, then propagating to main routine.

I agree, join should follow the call semantics

yelouafi:
>First of all, the only way to cancel effect is to cancel routine that is currently blocked on it.

IMO, it'd be clearer if you view cancellation at the effect level, not the subroutine level. a subroutine is cancelled by 1- cancelling its current effect, 2- throwing a CancellationException inside it. As for (1) it doesn't care what the current effect is: could be a race, all, another subroutine or a simple promise (noop); all it has to do is invoke _cancel on that effect. Composition will look more natural this way, and recursion will take care of all edge cases for us.

yelouafi:
in thinking about it, perhaps a `@@redux-saga/Cancellable` symbol is better than a `_cancel` property. So w'd now that a result is cancelable if it defines that symbol

aikoven:
> if someEffect failed, then we will cancel call(subroutine); if subroutine omitted to define cleanup code then we should propagate the exception to warn the developer and abort subroutine as well as main 
Then we'll end up with two exceptions: one for `someEffect` failure and the second for `CancellationException`; which one should we throw in `main`?

I think that the way user handles cancellation in subroutine should not influence main routine; that our two operations should be completely independent: we first cancel current effect and then throw exception in main routine regardless of how current effect reacted on cancellation.

I just don't see benefit in propagating `CancellationException` to parent assuming that it will be thrown in parent routine explicitly by (2).

I propose the following:
* If we have unhandled `CancellationException` then we stop parent routine immediately with exception; not via `iterator.throw()`, but by throwing it inside of `next()`
* Any other exception propagates as usual

This way we won't burden developer to write extra boilerplate for rethrowing. And in my opinion it would be easier to reason about execution flow if it was only subroutine's business to handle its own cancellation.

aikoven:
About cancelling race losers:
What if loser had some async cleanup code? Should we wait for it to complete before returning control to parent routine?

aikoven:
In regular `Promise.race` any loser promise rejection would be unhandled and I guess there is no way to warn about it other than just logging to console.

aikoven:
Maybe I'm wrong; I just think that result of cancellation of subroutine should have predictable effect on parent routine regardless of cancellation handling in subroutine.
And in my opinion it would be easier if  `CancellationException` is local — i.e. if we caught it, then it means that current routine was cancelled from outside, not that some child routine was cancelled.

yelouafi:
>I propose the following:

If we have unhandled CancellationException then we stop parent routine immediately with exception; not via iterator.throw(), but by throwing it inside of next()
Any other exception propagates as usual
This way we won't burden developer to write extra boilerplate for rethrowing. And in my opinion it would be easier to reason about execution flow if it was only subroutine's business to handle its own cancellation.

Sounds reasonable. 

In regular Promise.race any loser promise rejection would be unhandled and I guess there is no way to warn about it other than just logging to console.

Sounds also reasonable, we definitely throw away all losers: they are no longer part of our control flow

>And in my opinion it would be easier if CancellationException is local — i.e. if we caught it, then it means that current routine was cancelled from outside, not that some child routine was cancelled.

Agree, it makes more sens.

yelouafi:
>About cancelling race losers:
What if loser had some async cleanup code? Should we wait for it to complete before returning control to parent routine?

considering your solution, the response comes naturally. Any race losers are no longer part of our main control flow, any cleanup code they perform becomes independent. Like if they were detached from our main flow.

aikoven:
Ok then. 
I'm almost done, will push tomorrow.

aikoven:
Pushed. Here's how unhandled cancellation is reported:

image

Thinking about displaying effect stack in more readable way, maybe something like how you do in DevTools.

yelouafi:
Great! I'll push this into the master branch asap. Hope to make a release tonight.

Thanks again for the great work :+1:

yelouafi:
Merged. Since the master branch was ahead of the current task-cancel branch I resorted to manual merging.

I did some minor modifications: SagaCancellationException is constructed with 3 props: type (manual, race auto, parallel auto) which is a defined constant (to work with monitoring), saga which is the current saga being cancelled, and finally origin which is the root saga where the cancel was triggered from.

For reporting, I resorted to a simple console.error which will print the above infos

saga-cancel-error

As for the Saga monitor, manual cancellations are displayed as normal errors,

saga-cancel-monitor


 automatic cancellation errors are not displayed on loser effects

race-auto-cancel

yelouafi:
FYI. released a new version 0.4.0 with included cancellation support ([and other features](https://github.com/yelouafi/redux-saga/releases/tag/v0.4.0)). 

[Here are the docs for task cancellation](https://github.com/yelouafi/redux-saga#task-cancellation). Do you have any remarks ?

aikoven:
Looks great! Glad I could help 😀

PR #33, opened 2016-01-09
Add support to umd builds

cef62:
I've added support to generate valid umd builds using webpack. 
In production I use webpack/browserify too but often I need to create snippets or examples easy to share. Without an umd build this results nearly impossible.
Working on a blog post (WIP) on your library I've had workaround the problem, you can se a [jsbin](https://jsbin.com/vacasi/edit?html,js,console) using the new umd build.

I've already added instruction on the readme, let me know if you'd like something different.

Anyway thanks for the library, you're doing a great work on it!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ node_modules
 
 # compiled js files
 lib
+
+# compiled umd js files
+dist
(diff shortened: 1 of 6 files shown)


yelouafi:
I was just thinking about adding this! I like the jsbin ides, one can play with the library without heavy installs. Thanks! 

Looking forward for you blog post.

cef62:
Thanks @yelouafi, happy to help! 
After you'll push a new version on npm services like npmcdn.com will be able to serve the library with url like [https://npmcdn.com/redux-saga/lib/redux-saga.js](https://npmcdn.com/redux-saga/lib/redux-saga.js).

I'll let you know when the post is ready :)

yelouafi:
Already pushed. BTW the link is https://npmcdn.com/redux-saga@0.4.0/dist/redux-saga.js

Thanks again :)

PR #42, opened 2016-01-20
add `budo` to devDependencies

mz026:
add `bodo` into dev dependencies, which is required by `npm run counter` and other examples


yelouafi:
thanks

PR #71, opened 2016-02-02
yield [] should return immediately an empty array instead of blocking…

slorber:
… forever, like Promise.all([]) did

@yelouafi I encountered a bug while trying to replace:

```javascript
function* cancelAll(tasks) {
  try {
    yield tasks.map(cancel)
  }
  catch(err) { void(0); }
}

let subtasks = [];
yield fork(cancelAll,subtasks);
````

by:

```javascript
let subtasks = [];
yield subtasks.map(cancel);
```

It seems your new promiseless implementation does not handle empty arrays and this PR seems to solve the issue

### DIFF
--- a/src/proc.js
+++ b/src/proc.js
@@ -308,6 +308,13 @@ export default function proc(
     let completed
     const results = Array(effects.length)
 
+    function checkEffectEnd() {
+      if(completedCount === results.length) {
+        completed = true
+        cb(null, results)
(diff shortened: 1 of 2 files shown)


yelouafi:
thanks! I missed this one :)

yelouafi:
FYI released a patch version 0.6.1 with this fix.

PR #87, opened 2016-02-07
Warn user when using the monitor without process.env.NODE_ENV set to …

### LINKED ISSUE #85: Monitor: Cannot read property 'effect' of undefined
I set up the monitor pretty early in the development of my app, and ended up not using it too much because it's quite verbose when mapping over an array and yielding effects.

Now that I try using it again, an error shows up right after dispatching `LOG_EFFECT`:

screenshot from 2016-02-07 21 02 29

Adding the following in [`case LOG_EFFECT:`](https://github.com/yelouafi/redux-saga/blob/master/examples/sagaMonitor/index.js#L41) :

```javascript
      effectsById[action.effectId] = {...action,
        status: PENDING,
        start: time()
      }
```

makes the error go away but then I just (obviously) get:

screenshot from 2016-02-07 21 08 50

romseguy:
…'development'. Closes #85.

### DIFF
--- a/examples/sagaMonitor/index.js
+++ b/examples/sagaMonitor/index.js
@@ -21,6 +21,7 @@ const ERROR_STYLE = 'color: red'
 const AUTO_CANCEL_STYLE = 'color: lightgray'
 
 const time = () => performance.now()
+const env = process.env.NODE_ENV
 
 let effectsById = {}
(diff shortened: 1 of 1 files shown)


yelouafi:
thanks

PR #88, opened 2016-02-08
Add troubleshooting entry on sagas freezing the app

gaearon:
Just had this happen to me :smile:


yelouafi:
Yeah. Experienced that too :)

PR #90, opened 2016-02-09
Make real-world example work again

prashaantt:
I tried the real-world example and couldn't get it to compile or run without making these changes.


prashaantt:
My bad, I just noticed that `examples/real-world` has its own `package.json`. But I was misled by the `real-world` script in the root `package.json` that somehow made me believe I should be able to run it from root itself.  Perhaps you could remove it from there to prevent this confusion.

yelouafi:
>Perhaps you could remove it from there to prevent this confusion.

Yes. I think you're right

prashaantt:
I just found a better way to handle this:

```json
// package.json in root

"scripts": {
  "real-world": "npm --prefix examples/real-world install examples/real-world"
}
```

I'll submit a new PR if you like it.

yelouafi:
>I'll submit a new PR if you like it.

Of course. Thanks!

PR #109, opened 2016-02-12
please review - not certain if this is right

paul42:
when I cloned the repo, did npm install and then npm test, a single test failed, the above - there was one extra portion in the 'actual' section - was this extra logging left in, or should it be added to the expected part of the test? Or is it just my machine?

node: 4.2.3
npm: 2.14.7

I'm still trying to dive into the ecosystem (react/redux/redux-saga) so I just was thinking that if another novice like myself followed the instructions on the main page and got 1 test erroring out, it might just be my machine?

Thanks,
Paul

### DIFF
--- a/test/recipes/authorizationFlow.js
+++ b/test/recipes/authorizationFlow.js
@@ -162,6 +162,7 @@ test('Recipes: authorization flow', assert => {
     { type: 'LOGIN_SUCCESS', token: createToken(1) },             // t: 50
     { type: 'LOGIN_SUCCESS', token: createToken(2) },             // t: 100
     { type: 'LOGIN_SUCCESS', token: createToken(3) },             // t: 150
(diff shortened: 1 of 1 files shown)


yelouafi:
That's really curious. I made a change in my yesterday commit because the test failed. Initially i thought it was a manipulation error so I dropped the last result and the test passed. 

Turns out I was running the tests with node v0.10.40. Now That I switched to 5.0.0 I get the same results as yours. Not sure why the 2 versions give 2 different results, I'm using a magnitude of 50millis so I don't think it's related to timeout precision.

That's why I never trust tests with setTimeout.

yelouafi:
Ok here are some results; seems Node v0.10.40 accumulates timeout diffs more than I expected

image

paul42:
interesting... so the root cause was different behavior because of node version, I'll have to go back and read the actual logic being tested to verify to understand why I got the {token: ... } stuff in there - what is the suggested resolution for this?

yelouafi:
To be able to predict test results we have to drop timeouts and work only with promises that resolve in a predefined order

PR #112, opened 2016-02-15
fix exported function name: counter -> congratulate

designeng:
Reducer exported name should be equal reducer's file name.


yelouafi:
thanks!

PR #115, opened 2016-02-15
fix code typos

RainerAtSpirit:
btw code example in example repo is slightly different than what's used in the BeginnerTutorial. Not sure if it would be better to update the example or the tutorial.


yelouafi:
Thanks.

I'll have to update the repo example to match the tutorial

PR #131, opened 2016-02-17
Fix typos

Gpx:


yelouafi:
Thanks for all the effort!

Gpx:
No problem, I'm learning redux-saga and fixing one type here and there just takes few seconds.

PR #143, opened 2016-02-22
Fix realworld example

pavelkornev:
- update redux-sage semver
- removed unused redux-thunk


pavelkornev:
Redux-saga is crucial in this PR. It does not work at this moment with 0.3.3 version.

yelouafi:
Thanks. this code in webpack.config

https://github.com/yelouafi/redux-saga/blob/master/examples/real-world/webpack.config.js#L31-L52

pulled the dep. from the repo source. So didn't notice it in my local tests

PR #158, opened 2016-02-28
Fix error in the API reference CPS section

mehdichamouma:
In the note of the cps section it may be "fn call cb(error)" instead of "fn call fn(error)"


yelouafi:
thanks

PR #165, opened 2016-03-01
Attempt to fix https://github.com/yelouafi/redux-saga/issues/152

slorber:


yelouafi:
thanks; I'll have also to fix `cps`, an the most subtle one which is `fork`

yelouafi:
FYI pushed the fix. In the case of fork, I fail the forked task 

https://github.com/yelouafi/redux-saga/blob/master/src/internal/proc.js#L318-L322

https://github.com/yelouafi/redux-saga/blob/master/src/internal/proc.js#L125-L127

slorber:
great :)

yelouafi:
FYI https://github.com/yelouafi/redux-saga/releases/tag/v0.9.3

slorber:
thanks will update

PR #191, opened 2016-03-14
Add note regarding PUT effects on the store

tarjei:
I'm not completely happy with the sentence, but I thought I'd at least suggest something.


yelouafi:
thanks

PR #194, opened 2016-03-16
Better error printing in non-browser environments (node.js)

fasterthanlime:
I'm using redux-saga in an Electron app and most of my sagas run node-side.

In node.js, error objects passed to `console.warn` or `console.error` just print the error's prototype name and message, not the trace:

stack-trace

I changed it to manually print the trace, if one is available:

sample-error

It could be improved even more by tracking the whole saga chain so we know how we ended up there, but that's already a big step up :)

### DIFF
--- a/src/internal/proc.js
+++ b/src/internal/proc.js
@@ -73,8 +73,8 @@ export default function proc(
   )
 
   /**
-    this maybe called by a parent generator to trigger/propagate cancellation
-    W'll simply cancel the current effect, which will reject that effect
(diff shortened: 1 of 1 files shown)


yelouafi:
Thanks this would be very useful also in my unit tests

yelouafi:
>It could be improved even more by tracking the whole saga chain so we know how we ended up there

yeah, I think it could be achieved by passing the call hierarchy as param to `proc`

PR #204, opened 2016-03-19
Added "jsnext:main" entry to package.json for ES2015 module import.

ynohtna:
What do you think of rollup's proposal for supporting ES2015 module imports: [https://github.com/rollup/rollup/wiki/jsnext:main](url) ?

I've only been using rollup (instead of webpack) on a new project for a couple of days and haven't formed an opinion yet, but this small addition of a "jsnext:main" entry to package.json supports directly importing redux-saga's ES2015 source with rollup's "tree shaking" code elimination, and without needing an additional rollup commonjs plugin.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.9.4",
   "description": "Saga middleware for Redux to handle Side Effects",
   "main": "lib/index.js",
+  "jsnext:main": "src/index.js",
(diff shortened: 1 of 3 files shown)


yelouafi:
thanks. Haven't worked with rollup before, if this small addition would make it work then great :)

PR #208, opened 2016-03-22
Small docs fixed

KELiON:
While reading docs I found few super small typos and wrong return values and fixed them in this PR.


yelouafi:
thank

PR #215, opened 2016-03-28
Fix spelling errors

superRaytin:
Guess `alerady ` should be `already` :)


yelouafi:
Yes, it should :) thanks!

PR #239, opened 2016-04-01
Update docs

### LINKED ISSUE #238: Non blocking calls
What is `Api.storeItem` and `Api.clearItem` in [Non blocking calls](http://yelouafi.github.io/redux-saga/docs/advanced/NonBlockingCalls.html) section of docs?

It seems to me that it should be:
```javascript
yield call(Api.storeItem, {token})
// and
yield call(Api.clearItem, 'token')
```
Instead of:
```javascript
yield call(Api.storeItem({token}))
// and
yield call(Api.clearItem('token'))
```
Am I missing something here?

Mistereo:
Fix Api.clearItem/Api.storeItem usages.
Close #238


yelouafi:
thanks

PR #240, opened 2016-04-01
Export monitorActions.

danscan:
Export actions so users can easily ignore them in other middleware.  E.g.,
```javascript
import { monitorActions } from 'redux-saga';

const ACTION_TYPES_BLACKLIST = [
  ...monitorActions
];

export default store => next => action => {
  // If action type is not in blacklist, log it
  if (ACTION_TYPE_BLACKLIST.indexOf(action.type) < 0) {
    console.log('dispatching action:', action);
  }

  return next(action);
}
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -11,9 +11,11 @@ export { runSaga, storeIO } from './internal/runSaga'
 export { takeEvery, takeLatest } from './internal/sagaHelpers'
 
 import * as effects from './effects'
+import * as monitorActions from './internal/monitorActions';
 import * as utils from './utils'
 
 export {
   effects,
+  monitorActions,
   utils
 }


yelouafi:
Actually they are already exported from `redux-saga/utils` (see [this example](https://github.com/yelouafi/redux-saga/blob/master/examples/sagaMonitor/index.js#L5-L9)).

Alse note the actions are only fired on dev. mode not in production.

danscan:
Ok– I missed this.  This should work for my use case.  My issue is that I
want to filter out the internal redux-saga actions in my logger, so it only
affects me in dev.

On Mon, Apr 4, 2016 at 9:51 AM Yassine Elouafi <notifications@github.com>
wrote:

> Actually they are already exported from redux-saga/utils (see this example
> <https://github.com/yelouafi/redux-saga/blob/master/examples/sagaMonitor/index.js#L5-L9>
> ).
>
> Alse note the actions are only fired on dev. mode not in production.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/yelouafi/redux-saga/pull/240#issuecomment-205303603>
>
-- 
-- Dan Scanlon
Senior Software Engineer @ Tabtor
267.373.8883

yelouafi:
You can also use the sagaMonitor from the examples; it'll intercept the actions for you, plus you can have some nice logs about the saga activity. see the repo examples for usage

PR #258, opened 2016-04-14
Changing case in import to match path of Counter

saiyeek:
Was throwing :

> ERROR in ./examples/counter/src/main.js
> Module not found: Error: Cannot resolve 'file' or 'directory' ./Components/Counter in /home/nitesh/projects/learn/redux-saga/examples/counter/src
>  @ ./examples/counter/src/main.js 23:15-46


yelouafi:
thanks

PR #263, opened 2016-04-19
Add ES Modules build

dsblv:
Hello!

First of all, thank you for the awasome module!

This PR is intended to make `redux-saga` more [rollup](https://github.com/rollup/rollup)-friendly by transpiling `jsnext:main` files with babel.

Background: https://github.com/reactjs/redux/pull/1369

To keep things simple I used [es2015-weppack](https://github.com/gajus/babel-preset-es2015-webpack) babel preset, which is the same as just `es2015` but without commonjs module conversion. For commonjs builds, the `transform-es2015-modules-commonjs` plugin is added via `BABEL_ENV=commonjs` variable.

There is one problem though: people who use `jsnext:main` should import effect creators differently from what the docs suggest:

```js
import { take, put } from 'redux-saga/es/effects'

// or

import { effects } from 'redux-saga'

const { take, put } = effects
```

I don't konw how to work around it right now, any ideas?

Personally I like the second way better as it doesn't use per-file import so you're free to change your file structure anytime you want without breaking people's code.


yelouafi:
Sorry I did some changes on the package.json file before I saw this. could you make another PR?

Also I think script commandes for examples need to be prefixed with `cross-env BABEL_ENV=commonjs`

dsblv:
I could just rebase this one ontop of the `master`, does it sound good?

yelouafi:
Ok, thanks

yelouafi:
@dsblv Is this still relevant?

PR #272, opened 2016-04-24
Match buffer names in documentation.

secobarbital:
https://github.com/yelouafi/redux-saga/tree/master/docs/api#buffers


yelouafi:
yeap I missed that, thanks!

PR #273, opened 2016-04-24
Only check eventChannel matcher if defined

secobarbital:
The `eventChannel` function was enforcing the matcher function when passed 2 arguments, even though the function supports no matcher in any case.

### DIFF
--- a/src/internal/channel.js
+++ b/src/internal/channel.js
@@ -113,7 +113,7 @@ export function channel(buffer) {
 }
 
 export function eventChannel(subscribe, buffer = buffers.none(), matcher) {
-  if(arguments.length > 1)
+  if(typeof matcher !== 'undefined')
     check(matcher, is.func, 'Invalid match function passed to eventChannel')
 
   const chan = channel(buffer)


src/internal/channel.js
```diff
@@ -113,7 +113,7 @@ export function channel(buffer) {
 }
 
 export function eventChannel(subscribe, buffer = buffers.none(), matcher) {
-  if(arguments.length > 1)
+  if(typeof matcher !== 'undefined')
```
yelouafi: IMO it's be better to write `if(arguments.length > 2)`. because this would allows catching common typo Errors: i.e. the user provided a matcher argument but provided arg is undefined due to some typo error.

secobarbital: That will fix the issue I was having, which was passing 2 arguments without a `matcher`. However, I believe it is idiomatic to test arguments for `undefined` instead of testing number of arguments.

With the introduction of default arguments, there is now a reason to intentionally pass `undefined` arguments in order to use default values and still pass an argument after the defaulted one. This particular function is a great example: it has a default argument in the middle position, which means I have to intentionally pass in `undefined` as `buffer` if I wanted to take advantage of the default and still pass in a `matcher`.

BTW the `matcher` argument is [not documented](https://github.com/yelouafi/redux-saga/tree/master/docs/api#eventchannelsubscribebuffer) so we should probably fix the doc as well, but I don't know enough about it to fix the doc for you.

yelouafi: I understand, but in the above function there is no default value for the `matcher` argument. If I do the check only if `matcher` is not undefined, then it implies that this call form is allowed `eventChannel(subscribe, buffer, undefined)`. I dont think there is a case where a developer will want to pass undefined intentionally (except for setting the arg to the default value, which  is not the case above since match has no default value). If we got an undefined value it's likely  because of some typo error.

secobarbital: Got it. I will make the change so that we fix the documented behavior (2 arguments), but IMO it is the function caller/callee contract that `undefined` is the same as not passing an argument in that position.

I think one of the uses of `undefined` is to indicate no argument, which is why it is used to determine whether to use the default value or not. In fact for this particular function I would make the identity function the default matcher, and not perform any checks at all: `eventChannel(subscribe, buffer = buffers.none(), matcher = x => x)`

The mistake of passing in an undefined variable could be made in any argument position, so if I did that for the second argument, I would unintentionally get your default buffer type and that could also lead to very subtle bugs in my application. Without a default for the third argument we would be handling the two arguments inconsistently.

yelouafi: It's a valid POV but I personally think using `undefined` is not a good idea for indicating no args as it can overlaps with the so common undefined Errors (maybe also my typed lang. Java/FP background). i agree with your point on the `buffer` arg;  I considered making it mandatory or requiring a `null` instead of `undefined` but didn't like it much from the API viewpoint. So I just settled with the no-buffer solution. Trade-offs as usual :).

yelouafi: As a side note initially the buffer argument was checked using arguments.length like match, but since the arg is used internally by the lib (to resolve the `yield channel` effects) I adopted the default value approach because it makes the internal client code less complicated

yelouafi:
@secobarbital  I'm merging this as is and will do the change locally (sorry but I'm working on a patch today. Thanks for the contribution!)

secobarbital:
Thanks! Sorry for the delay, can only work on this on the weekends ;)

PR #279, opened 2016-04-25
Fix mistake in Error Handling part of docs issue: #278

### LINKED ISSUE #278: Mistake in Error Handling section of docs?
In `fetchProductsApi` in the last example I think the intent was to return an object literal - which doesn't work as expected with a single-line arrow function.

Had me scratching my head for longer than it should have 😄 

I'll submit a PR in a sec

See: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Returning_object_literals

matthetherington:
Fixes #278


yelouafi:
thanks

Issue #302, opened 2016-05-04
Ability to put directly to the store (putRedux / putStore)


tgriesser:
Would it make sense to add an effect similar to `put`, but which skips the redux-saga middleware and goes straight to the reducer? I currently have a `putRedux` effect which handles this, but the code to implement it is a bit ugly and far reaching - I wanted to see if there were any tips on how this could be easily implemented (with channels or otherwise), or if not if this would be a good addition to the core.

An example use case would be if you're receiving an action sent via a socket channel, and you'd want to put this external action to the redux store, but might not want to run the normal saga effect as would be called if the action were generated locally. 

Another would be if you had an xhr request to create a new resource, and the response provides a bit more info about this object and you want to re-use the same action name, sending the value/action directly to the reducer.

yelouafi:
Actually the puts are executed simply calling dispatch. The action gets back to the saga middleware via the normal middleware pipeline which  then emits the action to all sagas

So if I understand you want the putted action to be ignored by the saga middleware and not emitted to sagas?

tgriesser:
Yep, that's exactly what I'm looking for.

yelouafi:
Yeah that would be possible by setting some special property (e.g. SKIP_SAGAS) on the action which the middleware can check. But not sure if the Api should be to expose a new Effect or simply expose the special property. Mainly because the first solution require mutating the dispatched action by the middleware

tgriesser:
I'm going to close this, since the solution I'm after makes more sense if it's more generalized (skipping all middleware and not limited to just the sagas).

PR #303, opened 2016-05-05
Add putNext effect

tgriesser:
This is what I had in mind for #302 - adds a `putNext` effect which calls the lifted `next` function in the middleware chain with an `action`.

I suppose mutating the `action` with some sort of special property would be another option. This way seemed a bit more straightforward, interested to see what anyone else thinks.


yelouafi:
I don't think using the `next` param to forward actions *from the middle* is a good thing. 

Besides it's not the documented behavior for middleware (should use the the dispatch argument). This would also bypass other middlewares in the chain before the saga middleware. It also introduces too much complications on the implementation side (how many files need to be modified, change signature of the core proc).

OTOH if the intention here is to skip *all* middlewares and dispatch directly to Reducers, I think a more clean solution should be a dedicated store enhancer which will invoke the reducers directly

tgriesser:
Good points, I think i'll go with the dedicated store enhancer, thanks for taking a look.

PR #305, opened 2016-05-05
log error for check

floydwch:
The type check would be suppressed if it failed in a try-catch block:

```
  try {
    const json = yield call(fetch(target).then(resp => resp.json())) // wrong, it should be a function instead of a promise
  } catch (e) {
    yield put({type: 'FETCH_FAILED', message: e.message})
  }
```

There is no difference between the error from internal check and the exception from rejection. It's a little confusing for newcomers.


yelouafi:
makes sense. Thanks

PR #310, opened 2016-05-06
Added missing line for Api token storage in docs. Fixes #300.

### LINKED ISSUE #300: Missing code in docs
Hi there,

Amazing stuff in here, and also the docs are wonderful!

I was reading the *Advanced - Non blocking calls* docs section, and noticed that when it is said that *we need to move the token storage operation into the authorize task*, the `authorize` function code is missing a line for the actual token storage operation.

```js
function* authorize(user, password) {
  try {
    const token = yield call(Api.authorize, user, password)
    yield put({type: 'LOGIN_SUCCESS', token})
    yield call(Api.storeItem, {token}) // This is the missing line
  } catch(error) {
    yield put({type: 'LOGIN_ERROR', error})
  }
}
```

Just wondering if the line is actually missing or whether I am missing something here.

Thanks in advance!

nick13jaremek:


yelouafi:
Thanks

PR #314, opened 2016-05-08
Make `delay` cancellable

aikoven:
Add cancellation support for `delay()` function.

Also replace extra function `() => resolve(val)` with third argument for `setTimeout`.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,23 +5,9 @@ export { runSaga } from './internal/runSaga'
 export { END, eventChannel, channel } from './internal/channel'
 export { buffers } from './internal/buffers'
 export { takeEvery, takeLatest } from './internal/sagaHelpers'
-
-import { CANCEL } from './internal/proc'
-
-export { CANCEL }
+export { delay, CANCEL } from './internal/utils'
 
 import * as effects from './effects'
 import * as utils from './utils'
-
-export function delay(ms, val=true) {
-  let timeoutId
-  const promise = new Promise(resolve => {
-    timeoutId = setTimeout(() => resolve(val), ms)
-  })
-
-  promise[CANCEL] = () => clearTimeout(timeoutId)
-
-  return promise
-}
 
 export { effects, utils }
--- a/src/index.js
+++ b/src/index.js
@@ -3,16 +3,27 @@ export default middleware
 
 
 export { runSaga } from './internal/runSaga'
-export { CANCEL } from './internal/proc'
 export { END, eventChannel, channel } from './internal/channel'
 export { buffers } from './internal/buffers'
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -3,16 +3,27 @@ export default middleware
...
-export { delay } from './internal/utils'
 
+import { CANCEL } from './internal/proc'
 import * as effects from './effects'
 import * as utils from './utils'
 
+export function delay(ms, val=true) {
+  let timeoutId;
+  const promise = new Promise((resolve) => {
+    timeoutId = setTimeout(resolve, ms, val);
```
sompylasar: The third argument support isn't cross-browser ( http://stackoverflow.com/a/12657665 ), and would require a polyfill (though core-js that comes with babel seems to polyfill it https://github.com/zloirock/core-js/blob/e482646353b489e200a5ecccca6af5c01f0b4ef2/modules/web.timers.js ).

aikoven: That's correct. Although I'm not sure about current browser compatibility requirements for `redux-saga`, so maybe I should revert it.

yelouafi: @aikoven TBH didn't thank of browser compat. But in this simple case I think we could call the 'cross-browser' version of setTimeout (without args...)

aikoven:
I have updated the PR to use 2-arguments version of `setTimeout`. 

Also probably there's a better place for `delay` function than root module, maybe `src/utils.js`?

yelouafi:
>Also probably there's a better place for delay function than root module, maybe src/utils.js?

I was hesitant on this

- It makes more sense to put the function on the utils module as you said

- OTOH thought this would involve more imports for the user

```js
import { takeEvery } from 'redux-saga'
import { takr, put,  ... } from 'redux-saga/effects'
import { delay } from 'redux-saga/utils'

```

What do you think?

aikoven:
> OTOH thought this would involve more imports for the user

Yep, that's not very convenient. Besides, `src/utils` module acts as a facade, only exporting values from other modules.

Last options I see are to put it in `internal/sagaHelpers` or to introduce new module.

yelouafi:
>Last options I see are to put it in internal/sagaHelpers or to introduce new module.

I dont think it's natural place. this module implements helpers Generators (manually) which is not the cas of `delay`

We can let the implementation in `src/internal/utils` and export the function from the root module `src/index`. We can move the `CANCEL` constant itself from `src/internal/proc` to `src/internal/util` to avoid cross reference issues

aikoven:
Updated PR.
Not sure if notifications are sent with new commits.

yelouafi:
thanks! I'll make a 0.10.4 asap

PR #317, opened 2016-05-09
Enhanced sagaMonitor to work in Node environment.

sompylasar:
* In Node environment, filtered out the console CSS styles `%c` and
object output `%O` that aren't supported by the Node console.
* Added poor man's `console.group` and `console.groupEnd` that are
missing from the Node console without a polyfill.
* Added fallback for `performance.now()` that is missing from the Node
environment.
* Do not export to `window` unless it's available.
* Added verbose output that can show low-level calls (disabled by
default, some setter should be added to change the value in runtime).
* Added export of the snapshot-logging function for use from external
code.


yelouafi:
Great! thanks for this one!

PR #320, opened 2016-05-09
Clean up code style

billyjanitsch:
This PR is a general clean up of the code style. I was mainly aiming for consistency and tried to converge on the most commonly-used patterns across the project.

Let me know if you'd prefer that I move any of the conventions in the other direction (e.g. for single-line conditionals, blocks were used around 50/50, so I wasn't sure which you preferred.)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,6 @@
 import middleware from './internal/middleware'
 export default middleware
 
-
 export { runSaga } from './internal/runSaga'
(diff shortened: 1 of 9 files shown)


yelouafi:
thanks for the cleanup

> for single-line conditionals, blocks were used around 50/50, so I wasn't sure which you preferred

I thinks blocks are fine. 

Maybe we should put some rules in the eslint config

PR #326, opened 2016-05-12
fix anchor link to delay(ms, [val])

pjvds:
This pull request fixes the anchor link to the `delay(ms, [val])` function description.


yelouafi:
thanks

PR #328, opened 2016-05-13
Throw errors outside Promise chain.

laat:
In my apps I use window.onerror to log uncaught exceptions to the
back-end. However uncaught exceptions does not get processed by
`window.onerror` but is instead logged with `console.error`.

The setTimeout trick runs the function in the event loop outside the
Promise. Since the error is uncaught it gets processed by
`window.onerror`, and produces a clickable stack-trace in the console.


laat:
After reading a bit more of the code, I have come to the conclusion that this is a bit silly. We should perhaps instead let the log method be as it is, and use the setTimeout trick to rethrow unrecoverable exceptions whenever we encounter them. That should work on both node and in the browser.

PR #329, opened 2016-05-13
[RFC] Allow for chaining effects

totallymike:
I'm working on a fairly large project which currently executes asynchronous side effects in Promises via a middleware.

It's been pretty effective thus far, but we've outgrown it, so we're moving over to redux-saga.

:heart: :heart: :heart:

However we do lose sanity points in some circumstances:

```javascript
function checkStatus(response) {
  if (response.ok) { return response }
  throw new Error('NOT OKAY')
}

function extractJSON(response) {
  // Response.prototype.json returns a promise containing the JSON contents
  // of the body
  return response.json()
}

fetch(endpoint, options)
  .then(checkStatus)
  .then(extractJSON)
  .then(data => store.dispatch({ type: 'REQUEST_SUCCESS', payload: data }))
```

In order to achieve the same flow using sagas we need to do this:

```javascript
const response = yield call(fetch, endpoint, options)
yield call(checkStatus, response)
const data = yield call(extractJSON)
yield put({ type: 'REQUEST_SUCCESS', payload: data })
```

# Actual proposal:

I propose a new effect called `chain` which acts similarly to `compose` in redux, or any number of other pipeline/compose-style mechanisms you can find out there:

```javascript
/**
 chain([...effects], ...args)
**/
const data = yield chain([fetch, checkStatus, extractJSON], endpoint, options)
```

In the above example, effects are processed left to right.  The first effect receives any args you pass, and subsequent effects receive the return value of the previous.  I'm not married to this order though.  If we wish to make it look more like `compose`, we could execute right-to-left.

# Behavioral bits

So if we have a `chain` effect, we need to decide how it fits into the ecosystem.  Does it block?  Can you pass real effect objects to it?

## Going off the deep-end (this is where dragons live)

The way I see it, `chain` blocks until everything you pass it resolves.  This could be made more explicit by, instead of just passing functions, we pass `effects`:

```javascript
const data = yield chain([
  call(fetch, endpoint, options),
  call(checkStatus), /* what to do here, though, for arguments? */
  call(parseJSON),
])
```

For the remainder of my ramblings, assume this is a rough sketch, and we'd have a more declarative way of showing arguments passed to subsequent effects.

This does afford some extra fun additional behavior and crazy ambiguities as well, like forking instead of calling.  This could be treated like `tap` in lodash or Ruby, where the return values are ignored, and they just sit in the chain doing things:

```javascript
function* logger(thing) {
  console.log(thing)
}

yield chain([
  call(fetch, endpoint, options),
  fork(logger),
  call(checkStatus),
  call(parseJSON),
])
```

However, what if a fork throws an error?  Would it abort the whole chain?  What if the chain was finished by the time the forked effect threw its error?

### Okay this stuff is just madness now

If we somehow figure out how the above is supposed to work, what happens if we do this?:

```javascript
yield chain([
  take(FETCH_REQUEST),
  call(fetch, ...args),
  /* and so on */
])


totallymike:
For the record, I advocate for treating `chain` like `call`, at least for the time being, the way it looks in the first example.  This would be simplest to implement at the moment.

This RFC comes in the form of a pull request because I figure I'd throw up a spike for that behavior as we discussed it.

jbuck79:
I am very interested in this because I think it's a cool idea. But can I play devil's advocate for a second? Why don't you just create a function that calls the api and returns the json or an error object if it fails? That's what I did in my small app and it seems to be working well. The only drawback I can see is if you had some functions that needed json and some others that needed another type. But there aren't that many different types. You could easily further abstract it out into multiple fetch functions.

jbuck79:
The Nesting Sagas example here is pretty much what I did.

http://wecodetheweb.com/2016/01/23/handling-async-in-redux-with-sagas/

yelouafi:
Sorry for the delay. I was already thinking of something like this long ago. IMO it's better to keep things simple so we can get precise semantics: The way I think of it is exactly like function composition but with Generators (or functions returning promises).

But I think this can be implemented directly on user land: so `chain` would be like a higher order saga which takes a list of input sagas and then return a Saga with the above sequencing behavior. You can then either call the result Saga in a blocking way or fork it.

yelouafi:
closing this. Feel free to open another PR when you have something.

PR #335, opened 2016-05-17
Update docs to reference included delay effect

frederikb:
The current docs either don't tell where the delay effect is coming from or import it from 'somewhere'. Since delay is included by default in redux-saga we should update the docs to clear this up.

Some docs (recipes) which include an inline example of a delay effect haven't been amended.


yelouafi:
thanks

PR #341, opened 2016-05-22
fix { delay } import in tutorial

nmalancea:
It was previously giving this error:
`
redux-saga error: uncaught at check
call: argument fn is undefined
`


yelouafi:
thanks

Issue #286, opened 2016-04-28
Where is `getState` passed into a saga?


bricejlin:
I'm having issues accessing `getState`. Here's my code

```js
function* foo(getState) {
  console.log(getState())
}

export default function* rootSaga(getState) {
  yield [
    bar(),
    foo(getState),
  ];
}
```

entry point file:
```js
const sagaMiddleware = createSagaMiddleware();

const middleware = process.env.NODE_ENV === 'production' ?
  [ thunk ] :
  [ sagaMiddleware, thunk, logger() ];

const store = createStore(rootReducer, applyMiddleware(...middleware));

sagaMiddleware.run(rootSaga);
```
When I run this in the browser, I get this is the error:
```
utils.js?1f12:176 uncaught at rootSaga 
 at rootSaga 
 getState is not a function
```

Any ideas what I'm doing wrong?

fhelwanger:
It was changed on 0.10. Check: https://github.com/yelouafi/redux-saga/releases/tag/v0.10.0

> Removed deprecated `getState` argument passed to Sagas. You can use the `select` Effect to get the store state

To see how to use `select`: http://yelouafi.github.io/redux-saga/docs/api/index.html#selectselector-args

bricejlin:
@fhelwanger ah I see. I was following the docs, which I guess is outdated. Thanks for the pointer!

PR #355, opened 2016-05-29
Update docs replacing getState with select

jiayihu:
I noticed that the docs are still using `getState` argument of Sagas, deprecated in 0.10. Since I'm new to `redux-saga` I'd like to receive a feedback before continuing the update. Am I doing it right?
Thanks.

Related also to #286


yelouafi:
thanks

jiayihu:
Sorry it was a Work In Progress actually 😅 There are also some other places in the docs where it's still used `getState`, so should I make a new commit instead of `--amend` right?

yelouafi:
I think so. You can make new commits then just submit a new PR

PR #359, opened 2016-05-31
Fix takeEvery/takeLatest handling of arrays of symbols

jscinoz:
Fixes https://github.com/yelouafi/redux-saga/issues/246


yelouafi:
thanks

jscinoz:
No problem. I noticed another issue when providing a single symbol to takeEvery. I'll open a separate PR for that

PR #363, opened 2016-06-02
Support single symbol argument to takeEvery

jscinoz:
Follows from https://github.com/yelouafi/redux-saga/pull/359


yelouafi:
thanks

PR #373, opened 2016-06-06
use reduce instead of forEach.

nicholaswyoung:
To my mind, `reduce()` is greatly preferred when transforming data, rather than just looping over it. What do you think?

### DIFF
--- a/examples/real-world/actions/index.js
+++ b/examples/real-world/actions/index.js
@@ -4,9 +4,10 @@ const SUCCESS = 'SUCCESS'
 const FAILURE = 'FAILURE'
 
 function createRequestTypes(base) {
-  const res = {};
-  [REQUEST, SUCCESS, FAILURE].forEach(type => res[type] = `${base}_${type}`)
-  return res;
+  return [REQUEST, SUCCESS, FAILURE].reduce((acc, type) => {
+		acc[type] = `${base}_${type}`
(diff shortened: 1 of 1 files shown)


examples/real-world/actions/index.js
```diff
@@ -4,9 +4,10 @@ const SUCCESS = 'SUCCESS'
 const FAILURE = 'FAILURE'
 
 function createRequestTypes(base) {
-  const res = {};
-  [REQUEST, SUCCESS, FAILURE].forEach(type => res[type] = `${base}_${type}`)
-  return res;
+  return [REQUEST, SUCCESS, FAILURE].reduce((acc, type) => {
+		acc[type] = `${base}_${type}`
+		return acc
+	}, {})
```
Wenqer:
```
return [REQUEST, SUCCESS, FAILURE].reduce((acc, type) => 
  Object.assign(acc, {[type]: `${base}_${type}`}), {})
```

smashercosmo: That's less performant

nicholaswyoung: @wenqer `Object.assign` is useful, but mostly for merging objects. Not so much here.

yelouafi:
I think it's not a big deal here to mutate the object. Thanks

PR #397, opened 2016-06-20
Addons section to External Resources doc

stipsan:
Added links to the following addons:
* redux-saga-sc
* redux-form-saga
* redux-electron-enhancer

Originally I was just gonna do a PR for a link to my own package `redux-saga-sc` as I just finished working on its features for v1 and released it with documentation.
But I did a search to see if there were other libraries out there that lets you do cool stuff in your sagas and redux.

I thought it would be nice to have them in the section as well, as it looks like a nice ecosystem is starting to grow around redux-saga 😃


yelouafi:
thanks. 

>I thought it would be nice to have them in the section as well, as it looks like a nice ecosystem is starting to grow around redux-saga

yeah. That's nice to see. I hope there will be more add-ons like this

stipsan:
cool! I hope so too 😄

PR #398, opened 2016-06-20
fix es build to properly use jsnext:main field

mgmcdermott:
This commit takes the techniques done in redux here:
https://github.com/reactjs/redux/blob/3c900de44f1410fd2508401536b6e634e013e048/package.json#L25
and react-router here:
https://github.com/reactjs/react-router/blob/a3578f012aa85ba7e137e2bd08619e84212d5893/package.json#L20
to make proper use of the jsnext:main field.

Currently, anyone using redux-saga package through the jsnext:main field in their own projects is forced to not only use babel-loader to transpile redux-saga but also has to make sure they use the stage 2 preset. This adds some complexity to the babel setup of this project (splitting on environments) and introduces the use of BABEL_ENV, however it is necessary to compile to both umd/cjs and es modules.


yelouafi:
thanks for working on this.

Issue #396, opened 2016-06-20
Error when emitting null in eventChannel


Moeriki:
**Note:** `onFirebaseAuthStateChange` is an event emitter from Firebase. Emits a user object when a user signed on. Emits `null` when user signed out.

This logs an error.
```javascript
const firebaseAuthStateChangeChannel = () =>
  eventChannel((emit) => {
    onFirebaseAuthStateChange((user) => {
      emit(user)
    })
  })

export function* authWatch() {
  const userChannel = yield call(firebaseAuthStateChangeChannel)
  while (true) {
    const user = yield take(userChannel)
    if (user == null) {
      yield put(userLost(user))
    } else {
      yield put(userFound(user))
    }
  }
}
```

```
utils.js:188 uncaught at check Saga was provided with an undefined action
Hints:
    - check that your Action Creator returns a non-undefined value
    - if the Saga was started using runSaga, check that your subscribe source provides the action to its listeners
```

This works fine.
```javascript
const firebaseAuthStateChangeChannel = () =>
  eventChannel((emit) => {
    onFirebaseAuthStateChange((user) => {
      emit(user || 'null')
    })
  })

export function* authWatch() {
  const userChannel = yield call(firebaseAuthStateChangeChannel)
  while (true) {
    const user = yield take(userChannel)
    if (user === 'null') {
      yield put(userLost(user))
    } else {
      yield put(userFound(user))
    }
  }
}
```

I see how the error could be related. But I'm just wondering if this is intended design, and is there another way around this?

Thanks.

Moeriki:
On second thought. The channels should feed back a result much like an action creator does.

Turning the code into `emit({ user })` makes perfect sense then.

kuy:
@Moeriki Yes, it's a good workaround. I also use the same approach in [my exmaple](https://github.com/kuy/redux-saga-examples/blob/master/microblog/firebase.js#L10).

yelouafi:
@Moeriki yes. As you said it's by design to keep things consistent with Redux actions.

PR #402, opened 2016-06-21
explain to sanitize event source

Moeriki:
Does an explanation like this make sense?

Related: #396.


yelouafi:
Sounds good. Thanks

PR #406, opened 2016-06-23
Add debouncing example with takeLatest helper usage

R1ZZU:
This `while (true)` pattern looks a bit scary at first.. IMO version with `takeLatest` is much more friendly 🐼


yelouafi:
thanks. 

Maybe we should replace the entire example and keep only the version with `takeLatest`

PR #410, opened 2016-06-27
Introduce effectsSuspended hook

### LINKED ISSUE #399: Suspend execution of Effects
According to the documentation, the `yield` statements return descriptor objects, that are then processed by the middleware to perform the actual invocation of the side effects. Is there a way to suspend this processing? For example, when doing time-travel debugging?
Ideally, I'd want to store a boolean in the store indicating whether side effects should be performed or not. Then, I'd use that value to guard the processing of the descriptor objects. How can this be accomplished? Will there be a hook in `proc` for that?

rybon:
Fixes #399.

This allows for an `effectsSuspended` hook to be attached to the SagaMonitor object. This hook will be called in `runEffect` and returns a boolean. If `true` then the processing of the side effect will be cancelled. This helps with time travel debugging or replaying a Redux log during a test suite, where triggering side effects is not desirable.

Example:

```
const sagaMiddleware = createSagaMiddleware({
  effectsSuspended() {
    // imaginary selector
    if (noSideEffectsAllowed(store.getState()) {
      return true;
    }
    return false;
  }
});
```

### DIFF
--- a/src/internal/proc.js
+++ b/src/internal/proc.js
@@ -252,6 +252,9 @@ export default function proc(
   }
 
   function runEffect(effect, parentEffectId, label = '', cb) {
+    if (monitor && monitor.effectsSuspended()) {
+      return;
+    }
     const effectId = nextEffectId()
     monitor && monitor.effectTriggered({effectId, parentEffectId, label, effect})
 


yelouafi:
This could be an interesting feature, and if it can solve the time travel problem, it'd be great, but the PR only implements a part of it. For example, how do we resume once sagas are suspended? and more importantly we need an example of how this pause/resume is used to implement time travel debugging.

There is also an issue with 'ignoring' the effect above. Say for example we have a Saga like this

```js
function *saga() {
  yield take(ACTION)
  const resp = yield call(api)
  yield put(SUCCESS)
}
```

Suppose we suspend the saga while trying to execute the call effect. If we simply ignore the effect, then resuming it later will execute the put effect with an undefined action.

rybon:
You're right, that would lead to strange situations. I'm not sure how to solve this really. I guess a full page reload would be required to reset all sagas to their initial state. I'm ok with that for now, until we have a way to restore the internal state of the sagas using the Redux log. Replaying the Redux log does that job for the store, but not for the sagas just yet.

Do you have any ideas for solving this problem? This PR is also meant as a starting point for a discussion.

ruandao:
const resp = yield call(api)

can we save the `api` result on fly
when saga resume, and refill with it

like:
```
    const g = generator();
    g.next()
    .then((result) => {
            if(onPause) {
                addToResumeArr(g, result)
            }
     })

     ...
     if (resume) {
          const [g, result] = popFromResumeArr()
         g.next(result)
     }
```

PR #412, opened 2016-06-27
Proposal: takeSequence helper

R1ZZU:
This is a feature proposal PR, so any comments/suggestions are highly appreciated 🐱 

How this helper works:
* handles actions matching passed pattern
* starts new task
* if there is some task already running for this action, it will ignore this action

Basically, purpose of this helper is to provide implementation of `throttling` pattern out of the box

I'm still not sure about the name of this helper... Any thoughts? 💭

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -6,6 +6,7 @@
 * [`Saga Helpers`](#saga-helpers)
   * [`takeEvery(pattern, saga, ...args)`](#takeeverypattern-saga-args)
   * [`takeLatest(pattern, saga, ..args)`](#takelatestpattern-saga-args)
+  * [`takeSequence(pattern, saga, ..args)`](#takesequencepattern-saga-args)
 * [`Effect creators`](#effect-creators)
   * [`take(pattern)`](#takepattern)
   * [`takem(pattern)`](#takempattern)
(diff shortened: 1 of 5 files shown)


yelouafi:
>I'm still not sure about the name of this helper

Yeah. the name suggests that the saga doesn't ignore actions and queue them one after another.

Andarist:
im curious, how would you use that to achieve `throttle`? I mean, as far as I can see it it is not a standard `throttle` with `delay` parameter, just more like:

```
while (true) {
	const action = yield take(pattern)
	// do smth
}
```

am I right? its just waiting for the previous run to finish before accepting new action, so its kinda `takeFirstInSequence` ?

R1ZZU:
No, it's not like `takeFirstInSequence`, but `take` -> `spawn task` -> `ignore all incoming actions while task in progress` -> `take` -> ...

> it is not a standard throttle with delay parameter
yeah, but you are able to use `yield call(delay, duration);` inside spawned task and this will `throttle` your task.

Andarist:
well, my thought up name `takeFirstInSequance` was not ideal ofc, but your explanation is exactly what I was thinking about

well, in my opinion that would not be throttling, just delaying, truly throttled tasks would be independent of each other, in your solution they are completely dependant on the completion of previous one

some time ago I've implemented a helper in my project for throttle and I think its more accurate

```
export function* throttle(ms, pattern, task, ...args) {
  const throttleChannel = yield actionChannel(pattern, buffers.sliding(1))

  for (;;) {
    const action = yield take(throttleChannel)
    yield call(task, ...args, action)
    yield call(delay, ms)
  }
}
```

thinking now if it is possible to implement it as general helper with `fsmIterator`

R1ZZU:
well, yeah. I'm not saying my helper is direct implementation of throttling, but it could be used for it

```js
function* rootSaga() {
    yield* takeSequence(USER_INPUT, logUserInput);
}

function* logUserInput(action) {
    yield call(delay, 500);
    yield call(log, action.payload);
}
```

The main purpose of this helper is to provided an ability to ignore incoming actions, while previously spawned task not completed yet. Keep in mind that spawned task could delegate execution to more complex logic like races etc., so this helper is applicable not only for throttling, but a set of problems.

yelouafi:
Sorry for the long delay.

If I'm not mistakn the corresponding generator code is

```js
function* takeSequence(pattern, worker, ...args) {
  while(true) {
    const action = yield take(pattern)
    const task = yield fork(worker, ...args, action)
    yield join(task)
  }
}
```

But then `fork` followed by a `join` is equivalent to a simple `call`

```js
function* takeSequence(pattern, worker, ...args) {
  while(true) {
    const action = yield take(pattern)
    yield call(worker, ...args, action)
  }
}
```

I'm willing to add this helper as well, but I dont like much the `takeSequence` name. the helper name should suggest that the helper skip actions while blocked on the current task. or at the very least should not suggest that he is taking all actions in sequence

PR #424, opened 2016-07-05
Add recipe for retrying XHRs

nuc:
Hey there,

Based on that discussion: https://github.com/yelouafi/redux-saga/issues/349 I added an example for retrying XHRs.

I also added a second example for unlimited retries, for which I'd like some feedback if you think that it won't work as expected.

Thanks!

### DIFF
--- a/docs/recipes/README.md
+++ b/docs/recipes/README.md
@@ -105,3 +105,82 @@ function* watchInput() {
   yield* takeLatest('INPUT_CHANGED', handleInput);
 }
 ```
+
+## Retrying XHR calls
+
+To retry a XHR call for a specific amount of times, use a for loop with a delay:
+
+```javascript
+
+const delay = (ms) => new Promise(resolve => setTimeout(resolve, ms))
+
+function* updateApi(data) {
+  for(let i = 0; i < 5; i++) {
+    try {
+      const apiResponse = yield call(apiRequest, { data });
+      return apiResponse;
+    } catch(err) {
+      if(i < 5) {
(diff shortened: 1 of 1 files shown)


apfelfabrik:
Any feedback on this? If this is the recommended way to do retries, it should probably be in the recipes, if this is not the recommended way some suggestions for improvement would be great.

yelouafi:
Thanks and sorry for the delay.

juanibiapina:
One further extension to this recipe could be a way to extract the retry logic for reusability. It's a little cumbersome to duplicate this logic the way it is now.

Andarist:
this should work
```javascript
function* retry(maxTries, saga, ...args) {
	let tries = 0

	while (true) {
		tries++

		try {
			const result = yield call(saga, ...args)
			return result
		} catch (err) {
			if (tries === maxTries) {
				throw err
			}
		}
	}
}
```

nuc:
Would it make sense to add it to `internal/utils`, similarly as `delay`?

Andarist:
Yeah, it would certainly make sense. However I think its more appropriate as a helper (like `takeEvery` etc) than as util.

nuc:
Cool! I can prepare something. Cheers

aghreed:
Anything come of this yet? I'm also looking to have a reusable "retry", and it would be pretty neat if it were baked into the library as a helper. Happy to help contribute if you'd like as well.

Andarist:
Contribution would be greatly appreciated!

Issue #436, opened 2016-07-12
Issues with introduction text


jzaefferer:
Someone recommended this project at a conference, but after reading the first few paragraphs of https://github.com/yelouafi/redux-saga#redux-saga I can't tell at all if this is really what I'm looking for.

I think that's mainly caused by the vocabulary used there. Both "thunks" and "saga" get used without any introduction, apparently assuming that the reader already knows what these are. I have no idea. Same for "Side Effects middleware" and "Asynchronous Actions". Accordingly, the rest doesn't make more sense either.

Looking for other introductions, I found the beginner tutorial, but that isn't a good replacement to understand what this is even about. The other Introduction article (Background on the saga concept) links to a video and a scientific paper (of course a PDF). That doesn't work for me either.

I'm sure there's lots of interesting ideas here, they just need to be a bit more accessible...

slorber:
Hi @jzaefferer 

You may find my answer here easier to understand:
http://stackoverflow.com/questions/34570758/why-do-we-need-middleware-for-async-flow-in-redux/34623840#34623840

jzaefferer:
Not really, since it also uses a lot of vocabulary that I'm not familiar with. While I've read plenty of FP articles trying to explain monads, it never "clicked", so using that to explain redux-saga isn't helping. Same for "thunk", except that I only heard this term for the first time a few days ago.

I'm coming from an application that uses React, react-router and Ampersand models and collections. I have some problems in managing data in that architecture, and redux-saga was recommended to me to solve some of those problems. I still can't tell if it does.

slorber:
@jzaefferer if you don't involve time to understand underlying concepts it's hard to explain you redux-saga in the first place. It is not the typical library that beginners should start with. Plus the redux-saga project is not mature yet so the documentation can probably be made better over time.


This is a thunk, in the context of Redux:

```javascript
const userId = "123"

// Create a thunk (task usually done inside a Flux actionCreator)
const thunk = dispatch => fetch(`http://data.com/${userId}`)
    .then(res => res.json())
    .then(
      data => dispatch({ type: 'USER_PROFILE_LOADED', data }),
      err => dispatch({ type: 'USER_PROFILE_LOAD_FAILED', err })
    );

// You can dispatch a thunk
reduxStore.dispatch(thunk);
```


[...]

PR #437, opened 2016-07-13
Update README introduction

mxstbr:
Based on the discussion in #436, a first draft of an update to the introduction text

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 The mental model is that a saga is like a separate thread in your application that's solely responsible for side effects. This thread can be started, paused and cancelled from the main application with normal redux actions, it has access to the full redux application state and it can dispatch redux actions as well.
 
-It uses a soon-to-be-built-in JavaScript feature called Generators to make those asynchronous flows easy to read, write and test. *(if you're not familiar with [here are some introductory links](https://yelouafi.github.io/redux-saga/docs/ExternalResources.html))* By doing so, these asynchronous flows look like your standard synchronous JavaScript code. (kind of like `async`/`await`, but generators have a few awesome features we need)
+It uses an ES6 feature called Generators to make those asynchronous flows easy to read, write and test. *(if you're not familiar with them [here are some introductory links](https://yelouafi.github.io/redux-saga/docs/ExternalResources.html))* By doing so, these asynchronous flows look like your standard synchronous JavaScript code. (kind of like `async`/`await`, but generators have a few more awesome features we need)
 
 You might've used `redux-thunk` before to handle your data fetching. Contrary to redux thunk, you don't end up in callback hell, you can test your asynchronous flows easily and your actions stay pure.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,24 +2,13 @@
 
 [![Join the chat at https://gitter.im/yelouafi/redux-saga](https://badges.gitter.im/yelouafi/redux-saga.svg)](https://gitter.im/yelouafi/redux-saga?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge) [![npm version](https://img.shields.io/npm/v/redux-saga.svg?style=flat-square)](https://www.npmjs.com/package/redux-saga)
 
-An alternative Side Effects middleware (aka Asynchronous Actions) for Redux applications. Instead of dispatching Thunks which get handled by the `redux-thunk` middleware, you create *Sagas* to gather all your Side Effects logic in a central place.
+`redux-saga` is a library that aims to make asynchronous flows (e.g. data fetching) in React.js applications easier and better.
```
yelouafi: Do you think we can borrow something from this http://survivejs.com/blog/redux-saga-interview/#redux-middlewares. Mainly, define what 'Side Effects' are in the context of Redux (async and impure calls).

README.md
```diff
@@ -2,24 +2,13 @@
...
 
-This means application logic lives in 2 places:
+The mental model is that a saga is like a separate thread in your application that's solely responsible for asynchronousity. This thread can be started, paused and cancelled from the main application with normal redux actions, it has access to the full redux application state and it can dispatch redux actions as well.
```
yelouafi: While this is correct. There are also things that are synchronous but not pure (like saving to a browser storage, clearing a timeout ID, ...). If we define Side Effects in the first paragraph, we can simply put here that a Saga is solely responsible for Side Effects, or to reflect the opinions of the library: redux-saga proposes to 'isolate' all sorts of Side Effects (including async calls) inside Sagas.

README.md
```diff
@@ -2,24 +2,13 @@
...
 
-- Reducers are responsible for handling state transitions between actions.
-- Sagas are responsible for orchestrating complex/asynchronous operations.
+It uses a soon-to-be-built-in JavaScript feature called Generators to make those asynchronous flows easy to read, write and test. *(if you're not familiar with [here are some introductory links](https://yelouafi.github.io/redux-saga/docs/ExternalResources.html))* By doing so, these asynchronous flows look like your standard synchronous JavaScript code. (kind of like `async`/`await`, but generators have a few awesome features we need)
```
yelouafi: Perhaps worth mentioning that Generators are part of the ES6 standard

yelouafi:
sorry for the delay. I added some notes.

mxstbr:
Updated with your feedback! Let me know what you think, I definitely think this is a lot clearer.

yelouafi:
Looks great! thanks.

Sorry but one last remark, actually there is no mention that this a Redux middleware. Perhaps we should mention that

- The library is a redux middleware (although it can be used outside of Redux)  
-  [Optionally define what is Redux] a library which makes managing application state in consistent and predictable
- redux without middleware can not perform side effects but only transform state inside reducers

mxstbr:
> The library is a redux middleware (although it can be used outside of Redux)

:+1: Done

> [Optionally define what is Redux] a library which makes managing application state consistent and predictable
> redux without middleware can not perform side effects but only transform state inside reducers

I don't think we should add this to the introductory text, I think assuming that people know redux is a fair assumption for a library called `redux-saga` that mentions it's a redux middleware

yelouafi:
>I don't think we should add this to the introductory text, I think assuming that people know redux is a fair assumption for a library called redux-saga that mentions it's a redux middleware

Right.

I'll update the docs repo right away. Thanks again!

mxstbr:
Awesome! 🎉

PR #438, opened 2016-07-14
simple ring buffer implementation

Baisang:
Saw your TODO in `buffers.js` about using a ring buffer and decided to try writing it.

I'm not sure what to put as the default `limit` (it can't be `Infinity`, obviously) as I am not very familiar with what this buffer is used for. Feel free to change it.

It passes your provided tests, but I'm not sure how much performance increase this will provide. Up to you whether to merge it in or not, but would like your feedback regardless.

The `new Array(n)` constructor is used to make sure we're always working with a fixed size array. Theoretically this should result in performance boosts.

The length variable within the `ringBuffer` simplifies checking if the buffer is empty or not (no edge case where `pushIndex == popIndex` but the buffer could be full or empty).


Baisang:
Err, I tried committing the changes after I did `npm install` to add the output, but it seems to have changed things a lot more differently than I expected. Sorry about that; please just consider my changes to `buffer.js` by themselves.

yelouafi:
Thanks for working on this. There is one concern though: when taking out elements from the buffer (using `take` or shifting in the Sliding Buffer case) the code just updates the indexes which can cause memory leaks since the underlying array is still holding a reference to the removed data. So I think we need to free the reference to the data in the array cells (by setting it to `null` or `undefined`) 

>I'm not sure what to put as the default limit (it can't be Infinity, obviously) as I am not very familiar with what this buffer is used for. Feel free to change it.

I think we can just put a low limit as default (eg `10` like the max num of listeners in Nodejs). The lib user can still put a higher number if required (Another option could be: when there is no limit argument=default to Inifnity, allocate a bigger array and copy old data here. But I don't like much this approach since it can hide memory leaks from the lib user)

>It passes your provided tests, but I'm not sure how much performance increase this will provide. Up to you whether to merge it in or not, but would like your feedback regardless.

I'll try to run some perfs tests locally. But I expect it to be more performant b/c we're working with a fixed array in memory and most VMs has optimizations for this cases (continuous blob on OS memory). In the actual impl. `arr.shift` (maybe even `arr.push`) calls will likely be causing deoptimizations on VMs

>The length variable within the ringBuffer simplifies checking if the buffer is empty or not (no edge case where pushIndex == popIndex but the buffer could be full or empty).

the length could be calculated from `Math.abs(pushIndex - popIndex) % limit` but it's not obvious to me which approach is more performant: calculating on demand or updating the length var each time (though I don't think this has significant impact)

Baisang:
Thanks for looking at my PR, I implemented the changes you mentioned.

Regarding the potential memory leaks in the shifting case, `arr[pushIndex] = it` replaces the reference to the removed data, unless I am missing something else. I modified the `take` call to set the cell to null.

`%` can be expensive at times so I think it is best to use the length variable, but of course in practice this could be different.

yelouafi:
>Regarding the potential memory leaks in the shifting case, arr[pushIndex] = it replaces the reference to the removed data

Yes. You're right.

PR #442, opened 2016-07-18
added clarification in apply docs

Baisang:
Was using apply and was slightly confused on how the function should be called, this should clarify things for others.


yelouafi:
thanks

PR #444, opened 2016-07-19
Eventchannel example

Baisang:
This is something that took me a little time to work through, so I figured I'd put up an example of how to make sagas work well with provided event handler definitions in a library.

If it's irrelevant feel free not to merge it.


yelouafi:
I think we need a more concrete example. Like subscribing to a websocket, browser history event.

Andarist:
I've forked @Baisang branch and changed this example into socket.io one, hope its good - https://github.com/yelouafi/redux-saga/pull/464

PR #449, opened 2016-07-21
action is undefined and I think incorrect

michaelAndersonCampingWorld:
I believe that in this example that you're actually trying to set the lastAction to be winner.action when it's truthy.


yelouafi:
Yes you're right. Thanks

PR #455, opened 2016-07-25
Allow helpers return task effects into sagas in order to continue sag…

Andarist:
Allow helpers return task effects into sagas in order to continue saga execution, additionally it makes those helpers cancelable too.

I'm not sure if there is any con of that, except that any custom generator will be treated accordingly. Just got really confused that `takeEvery` is not like `fork` - non-blocking and that there is even no possibility to resume saga (to my knowledge) when `takeEvery` is used so it may be only the last statement of the saga (if not forked).

Anyway - I was just diving into the code to understand how `takeEvery` is handled under the hood, thought of this little fix that is getting the job done, however I might not see a big picture and anyhow it would be nice to understand some more of the saga internal architecture if im wrong.


src/internal/proc.js
```diff
@@ -356,7 +356,8 @@ export default function proc(
   }
 
   function resolveIterator(iterator, effectId, name, cb) {
-    proc(iterator, subscribe, dispatch, getState, options, effectId, name, cb)
+    const task = proc(iterator, subscribe, dispatch, getState, options, effectId, name, cb)
```
yelouafi: Note that here we pass `cb` as last param to `proc`. It will be called once the iterator is done

yelouafi:
the `yield iterator` is a blocking call. It means we'll have to wait for the iterator to terminate before resuming the caller. In your PR we'll be calling the continuation `cb` twice: 1st with `task` and later with the return value of the iterator

Andarist:
ok, I understand this, good to know

still dont you think that such helper as `takeEvery` should return its task to the caller? it has no return value, so its stoping caller indefinitely

Ofc this should be done in some other way than i've proposed here

Andarist:
Probably here, those special helpers just dont need to specify its `cb` so it can by ignored in `proc` when iterator returns. 

As I can see from a quick look into the code - those helpers dont even specify their callbacks. So such case with no callback could probably return task to the caller? And if callback is specified then it would be handled by `proc`?

This would require just an if `resolveIterator` and both scenarios would be handled just fine I think - no double calling `cb`.

Or I have thought about wrapping `takeEvery` such it would not be a blocking call and therefore could yield task to the caller. Just would have to wrap it so it use forking mechanism which would for a iterator-based worker in the background. It would preserve blocking nature of yielding iterator but in the same time make helpers (at least those 2 present ones) non-blocking.

Aint sure how I would implement it yet, but might give it a try if you green light it + maybe give some tips about desired implementation details.

yelouafi:
Actually you can have the desired behavior using `fork`

```js
yield fork(takeEvery, action, saga, ...args)
```

>still dont you think that such helper as takeEvery should return its task to the caller? it has no return value, so its stoping caller indefinitely

So you're proposing to make yielding to helpers non blocking and return a task descriptor *by default* to the caller. That sounds reasonable to me. The only concern is that could break the existing behavior (blocking); not sure how this will affect existing code (OTOH since blocking on saga helpers will block the parent saga forever (b/c saga helpers never return) I don't see how changing the behavior could negatively affect an existing flow)

>Aint sure how I would implement it yet, but might give it a try if you green light it + maybe give some tips about desired implementation details.

we can test if the iterator was created by  a Saga helper; 

https://github.com/yelouafi/redux-saga/blob/master/src/internal/proc.js#L328

So we'll need to put some flag on the iterators returned by the following method (which can be tested in the above code)

https://github.com/yelouafi/redux-saga/blob/master/src/internal/utils.js#L92-L98

If the iterator was created by a saga helper we call `runForkEffect` (instead of `resolveIterator`)

https://github.com/yelouafi/redux-saga/blob/master/src/internal/proc.js#L430

But for this we need to make  `runForkEffect` support forking iterators directly (instead of functions returning iterators).

Andarist:
@yelouafi 

Ok, I've managed to implement this, wasn't that hard after all ;) Would love to see your code review on that, also I didnt write tests (yet). Please take a loot at this PR, if everything is ok I'll write tests and rebase to master, so it can be merged if you decide to proceed with this.

Pre-existing tests are passing, so things should be fine.

PS. I had problem with `cross-env BABEL_ENV=es babel src --out-dir es` script, but didnt change anything regarding that. Probably some wrong node/npm version here and things got crazy a little bit.

Andarist:
Tweaked existing tests to accomodate new behaviour. Aint sure if any additional tests are needed, personally dont think so.

Andarist:
@yelouafi 
I've rebased the branch now, so it can be merged without conflicts. When you have some spare time I would appreciate you looking at that. Cheers!

yelouafi:
Seems ok to me; thanks for all the work!

Could you update the [API docs for the helpers](https://github.com/yelouafi/redux-saga/tree/master/docs/api#saga-helpers)? (If you don't have time it's OK, I can do it)

Andarist:
Yeah, sure, gonna update the commit some time soon.

Andarist:
Updated API docs and some other parts of the docs as well, aint sure if I've spotted every relevant part though.

PR #461, opened 2016-07-29
Improving the documentation regarding select()

bobkocisko:
Improving the documentation regarding select() so that no false guarantees are made about state and saga run timing


yelouafi:
Yeah; that's a subtle case, thanks

PR #464, opened 2016-08-01
Eventchannel doc

Andarist:


docs/advanced/Channels.md
```diff
@@ -161,25 +161,38 @@ export function* saga() {
...
+import socket from './socketConnection' // already established and exported connection
 
-export function* saga() {
+function* pong() {
+  yield call(delay, 5000)
+  socket.emit('pong')
+}
+
+export function* watchOnPings() {
+  const socketChannel = eventChannel(emit => {
```
yelouafi: It's not possible to test this form. Can we extract this into a separate a function like `createWebSocketChannel` and then call it using a declarative form `yield call(createWebSocketChannel)`?

Andarist: Ye, gonna amend it in few minutes

Andarist: @yelouafi changed

yelouafi:
thanks

PR #466, opened 2016-08-03
Fix link in API Reference

y0za:
The destination tag of `put(channel, action)` is `<h3 id="putchannel-action">`. I fixed the link to match its id.


yelouafi:
Thanks

PR #483, opened 2016-08-11
Added throttle helper

Andarist:
- [x] approval
- [x] tests
- [x] docs


yelouafi:
Seems good to me. TBH didn't think of using action channels with sliding buffers for throttling. I'll update the throttling recipe on the docs with your implementation.

And thanks for all your contributions!

Andarist:
So should I close this PR? I mean do you plan to keep such helpers out of the core (as you have mentioned adding it just to the recipes)? Did you maybe consider this mono-repo thing so maybe more helpers could be included and imported by ppl just on demand?

yelouafi:
Oh sorry I was not quite explicit! I meant we could pass to the next steps you mentioned (tests and docs). I dont mind adding helpers for common cases to the core.

yelouafi:
For the recipes I just meant your solution is cleaner than the actual one & that I should update it.

Andarist:
Can make it as part of this PR - updating recipe. But when I'm done with docs and tests - is it even needed in the recipes? I mean underlaying implementation? It can just be explained alongside with the helper I think.

Also I have similar helper written as generator for `trailingThrottle`. Dont know though if it deserves geting a standalone helper (which I can implement using `fsmIterator`) or do you think its good only for a recipe?

```
function* delayFirstCall(ms, buffer, task, ...args) {
	yield call(delay, ms)

	if (!buffer.isEmpty()) {
		return
	}

	yield call(task, ...args)
}

export function* trailingThrottle(ms, pattern, task, ...args) {
	const buffer = buffers.sliding(1)
	const throttleChannel = yield actionChannel(pattern, buffer)

	for (;;) {
		const firstActionInSequence = yield take(throttleChannel)
		yield fork(delayFirstCall, ms, buffer, task, ...args, firstActionInSequence)

		for (;;) {
			yield call(delay, ms)

			if (buffer.isEmpty()) {
				break
			}

			const action = yield take(throttleChannel)
			yield call(task, ...args, action)
		}
	}
}
```

Andarist:
I've updated the PR with test and docs 🎉, please take a look. :)

docs/api/README.md
```diff
@@ -216,6 +217,57 @@ function* takeLatest(pattern, saga, ...args) {
 }
 ```
 
+### `throttle(ms, pattern, saga, ...args)`
+
+Spawns a `saga` on an action dispatched to the Store that matches `pattern`. After spawning a task it's still accepting incoming actions into the underlaying `buffer`, but keeping at most 1 (the most recent one), but in the same time holding up with spawning new task for `ms` miliseconds (hence it's name - `throttle`). Purpose of this is to ignore incoming actions for a given period of time while processing a task.
+
+Each time an action is dispatched to the store. And if this action matches `pattern`, `takeLatest`
+starts a new `saga` task in the background. If a `saga` task was started previously (on the last action dispatched
+before the actual action), and if this task is still running, the task will be cancelled.
```
yelouafi: The whole paragraph belongs to `takeLatest` and should be removed

docs/api/README.md
```diff
@@ -216,6 +217,57 @@ function* takeLatest(pattern, saga, ...args) {
...
+#### Example
+
+In the following example, we create a simple task `fetchAutocomplete`. We use `throttle` to
+start a new `fetchAutocomplete` task on dispatched `FETCH_AUTOCOMPLETE` action. However since `throttle` ignores consecutive `FETCH_AUTOCOMPLETE` for some time, we ensure that user won't flood our server with requests.
+
+```javascript
+import { throttle } from `redux-saga`
+
+function* fetchAutocomplete(action) {
+  const autocompleteProposals = yield call(Api.fetchAutocomplete(action.text))
```
yelouafi:
```js
const autocompleteProposals = yield call(Api.fetchAutocomplete, action.text)
```

Andarist:
@yelouafi Fixed mentioned issues

yelouafi:
>Can make it as part of this PR - updating recipe. But when I'm done with docs and tests - is it even needed in the recipes? I mean underlaying implementation? It can just be explained alongside with the helper I think.

Hmm sounds right

>Also I have similar helper written as generator for trailingThrottle. Dont know though if it deserves geting a standalone helper

>Also I have similar helper written as generator for trailingThrottle. Dont know though if it deserves geting a standalone helper 

I think so; you can open a PR for it


the code seems ok; I've left some notes on docs.

I'm having some trouble to understand what the test does.  Maybe I missed something but each worker waits for a def in the array. but since most of defs are not resolved then the worker will never put results into the expected array. My point is this won't permit us to observe throttling since even if the workers are not throttled they will not make it to the `actual.push(...)` instructions. (and sorry If I missed something, was a quick review)

Andarist:
>> Also I have similar helper written as generator for trailingThrottle. Dont know though if it deserves geting a standalone helper

>I think so; you can open a PR for it

aint sure with which one should I proceed - implementing `trailingThrottle` as built-in helper or make a recipe out of it?

> I'm having some trouble to understand what the test does. Maybe I missed something but each worker waits for a def in the array. but since most of defs are not resolved then the worker will never put results into the expected actual array. My point is this won't permit us to observe throttling since even if the workers are not throttled they will not make it to the actual.push(...) instructions. (and sorry If I missed something, was a quick review)

In fact there are only 5 `defs` created (all geting resovled eventually) and 35 `actions` dispatched. So not every `action` is geting their corresponding `watcher` started (cause of throttling)

docs/recipes/README.md
```diff
@@ -2,63 +2,19 @@
...
 }
 
 function* watchInput() {
-  while (true) {
-    const { input } = yield take('INPUT_CHANGED')
-    yield fork(handleInput, input)
-    // throttle by 500ms
-    yield call(delay, 500)
-  }
+  yield throttle(500, 'INPUT_CHANGED', handleInput)
```
axelson: As a user it would be nice to have `throttle` imported in this example so that I know where to find it from.

Andarist: added, thanks 👍

yelouafi:
>In fact there are only 5 defs created (all geting resovled eventually) and 35 actions dispatched. So not every action is geting their corresponding watcher started (cause of throttling)

You're right,I didn't notice the `Math.floor(payload / 10)`.

PR #485, opened 2016-08-12
Undo recipe

neurosnap:
I created this recipe because the section `TaskCancellation` got me most of the way to creating simple undo functionality but it left a few things out.  Granted this recipe is very similar to the task cancellation section so I understand if it isn't different enough.  I figured `undo` was a common enough problem that having another example could be useful.

Excellent library btw, I was shocked at how easy it was to implement the ability to undo a task.

### DIFF
--- a/docs/recipes/README.md
+++ b/docs/recipes/README.md
@@ -184,3 +184,54 @@ export function* watchUpdateResource() {
 }
 
 ```
+
+## Undo
+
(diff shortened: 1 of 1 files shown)


yelouafi:
Sorry for the delay. Looks like a special case of debouncing but it's a good demo for optimistic updates (using a delay).

>Excellent library btw, I was shocked at how easy it was to implement the ability to undo a task.

Thanks!

PR #490, opened 2016-08-15
Typings for put.

yeerkkiller1:
Makes T extend Action, so code like:
`yield put([])`
Fails at compile time, instead of at runtime.

### DIFF
--- a/effects.d.ts
+++ b/effects.d.ts
@@ -1,3 +1,4 @@
+import {Action} from "redux";
 import {Channel, Task, Buffer, Predicate} from "./types";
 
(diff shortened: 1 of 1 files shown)


yelouafi:
thanks

aikoven:
What if you want to put to a channel of type other than Action?

yeerkkiller1:
That's true, sorry about that. Fixed with https://github.com/yelouafi/redux-saga/pull/512

yelouafi:
Missed that; thanks @aikoven for the catch

PR #494, opened 2016-08-17
Fix deprecated dependency

Andarist:
ES build stopped working because of `babel` change (more here - https://github.com/yelouafi/redux-saga/issues/489#issuecomment-240388728 ). I've made a fix in `babel-preset-es2015-webpack` but it didnt get merged in as it got deprecated in favor of new `babel-preset-es2015`, more here - https://github.com/gajus/babel-preset-es2015-webpack/pull/16#issuecomment-240405169 . 

With updated package build runs smoothly.


yelouafi:
thanks!

PR #500, opened 2016-08-18
Fixed missing yields in Sequancing Sagas example (fixes #499)

### LINKED ISSUE #499: Sequencing Sagas example needs to yield
The [Sequencing Sagas](http://yelouafi.github.io/redux-saga/docs/advanced/SequencingSagas.html) example reads as such:

```js
function* playLevelOne() { ... }

function* playLevelTwo() { ... }

function* playLevelThree() { ... }

function* game() {
  const score1 = yield* playLevelOne()
  put(showScore(score1))

  const score2 = yield* playLevelTwo()
  put(showScore(score2))

  const score3 = yield* playLevelThree()
  put(showScore(score3))
}
```

If I understand Sagas correctly, then those `put` lines should be changed to `yield put`:

```js
function* game() {
  const score1 = yield* playLevelOne()
  yield put(showScore(score1))

  const score2 = yield* playLevelTwo()
  yield put(showScore(score2))

  const score3 = yield* playLevelThree()
  yield put(showScore(score3))
}
```

Is that correct?

Andarist:


yelouafi:
thanks

PR #501, opened 2016-08-19
Added new effect - flush. It can be used to clear channel from buffer…

### LINKED ISSUE #481: Buffer flush, takeAll
Im just encountering a scenario where I need to:
1. buffer some type of actions
2. until another action happens
3. take all buffered actions and do smth with them
4. proceed with taking actions from the channel normally, 1 by 1 with take

I thought that 3rd step is easily done with a for loop, but in order to make it right I need to pass to the `actionChannel` my custom `buffer`, so I can reference it later while checking its `buffer.isEmpty` in order to pull all buffered msgs at once, cause simple `take` wont work as its a blocking effect. I thought it would be nice to have some kind of a helper/effect for such a situation.

Also I imagine that one could need to just flush all contents of the channel and it would be nice to have such effect too.

In fact it could be just a 1 single effect called i.e. `flush` which would return to the saga all buffered content as an array and empty the underlaying buffer at the same time. So it would cover both cases.

```
const actions = yield flush(actionChannel) // first case when I need buffered content
yield flush(actionChannel) // second case when I just need to flush it away
```

If that sounds reasonable I can implement it

Andarist:
…ed items and geting them back in a batch in the saga. (closes #481)


sergej-s:
+1 flushing everything from channel is useful feature.

src/internal/proc.js
```diff
@@ -600,6 +601,21 @@ export default function proc(
     cb(!!mainTask.isCancelled)
   }
 
+  function runFlushEffect(channel, cb) {
+    const bufferedItems = []
+    for (;;) {
+      let exposedItem
+      channel.take(item => exposedItem = item)
```
yelouafi: if the buffer is empty, this callback wont be called until the next item is put into the channel. The saga will remain blocked until then

yelouafi:
thanks for the PR. a flush Effect can effectively a  very useful future which allows more operations on channels (debouncing, buffering with time...)

But I'm having some trouble to understand the proposed implementation (left some notes). Also there is a need for more complete semantics

- is this a blocking Effect: I guess no, we want to get only the buffered messages if any, otherwise we should resume with ... an empty array (right?)

- if we try to flush a closed channel, the result should be the same as taking (resume with END which will end the Saga) ?

For the implementation I think it should be done at the channel level (having a `channel.flush` method perhaps backed by a similar method in the buffer)

yelouafi:
for the semantics question, I have a decision table which describes exhaustively the semantics of each action of the channel (sort of transition function for a state machine). If we add the `flush` action to the table we can define the complete semantics for it (what the action should do on each channel's state)

channels-trans-table

src/internal/proc.js
Andarist: Ye, you are right ofc. ;) I just remember now that's the reason I had to create buffer on my own and pass it to the channel, so I could reference it later and take directly from it instead of the channel, to achieve this flushing in my app.

Andarist:
1. It should be non-blocking.

2. Yeah, I was not sure what should be done if closed channel is tried to be flushed. Didnt yet understand how cancelation works and what behaviours must be accounted into such implementation for it to work. We can ofc resume with `END` although when Im looking at `take` implementation - its resuming with end on:

```
if(closed && buffer.isEmpty()) {
      cb(END)
```

Isnt it wrong? I mean shouldnt close channel always resume `END` then and not only if it is closed and empty? Especially that closing the `channel` aint clearing `buffer`?

yelouafi:
> Isnt it wrong? I mean shouldnt close channel always resume END then and not only if it is closed and empty? Especially that closing the channel aint clearing buffer?

No because the channel has still some items to deliver. Future takers will get buffered items until the channel is empty. Only then we can resolve with END since no more items can be added to the channel. For the flush effect of course we will deliver all buffered items not just one.

For cancelation it's not a problem. Since the effect is synchronous (non blocking) it can't be cancelled

Andarist:
```
state							| 	action
--------------------------------------|--------------
closed	| 	takers	|	buffer	|	flush
--------------------------------------|---------------
T 		| 	empty	|	empty	|	it(END)
T 		| 	empty	|	part		|	it(buf.flush())
T 		| 	empty	|	full		|	it(buf.flush())
--------------------------------------|---------------
T 		| 	filled	|	empty	|	not possible, cause takers should be resolved?
T 		| 	filled	|	part		|	
T 		| 	filled	|	full		|	
--------------------------------------|---------------
F 		| 	empty	|	empty	|	it([])
F 		| 	empty	|	part		|	it(buf.flush())
F 		| 	empty	|	full		|	it(buf.flush())
--------------------------------------|---------------
F 		| 	filled	|	empty	|	not possible, cause takers should be resolved?
F 		| 	filled	|	part		|	
F 		| 	filled	|	full		|	
--------------------------------------|---------------
```

Iint sure about the cases when takers are filled, cause I dont know how exactly scheduling things work here yet. But the rest of it is pretty straight forward I guess

Andarist:
@yelouafi 

I've updated my PR with corrected version of the `flush` effect. Gonna attach tests later.

yelouafi:
This seems ok to me.

>Also thought that flush could take matcher argument, but aint sure how to proceed with that, cause most likely it will need to implemented in the buffer itself, but for now it stays agnostic of what's inside it (which is nice). In the same time I've found a bug (#507) which to be solved would require similar implementation I think - matcher argument in buffer.take

Actually I'm not sure if we should consider it a bug, I'll put some comment on the issue.

yelouafi:
And of course thanks for the great work!

Issue #515, opened 2016-08-31
Does call effect creator really only works with generator and promise returning functions?


chodorowicz:
In the docs it's stated that first argument of `call` function must be
>  A Generator function, or normal function which returns a Promise as result
https://yelouafi.github.io/redux-saga/docs/api/index.html#callfn-args

I'm wondering if it's really true. I've found examples in the issues by @slorber which use functions which are not generators and don't return promises (e.g. https://github.com/yelouafi/redux-saga/issues/79#issuecomment-215665308). Also I'm trying out that in my code and it seems to work. But I also don't wont write to code which doesn't follow the docs and function signatures. It would be quite useful to call arbitrary functions using this side effect creator.

Greets! :)

slorber:
As far as I know, call can be used with normal / sync functions too. The advantage of using call fr a sync function being probably that the call can easily be tested like all other effects, while invoking the function directly without yielding the call would make this harder.

Not sure my comment is a good example of this as I'm using arrow functions `call(() => doSomething())` so it does actually not really help for testability compared to having simply called `doSomething()` directly.


I guess call will work fine and maybe handle sync functions returning promises, normal results, array of promises etc... maybe a function returning a generator too?

chodorowicz:
Thanks! 
Yeah, of course, the example I've linked doesn't help with testing, just shows a non promise function being argument of call. And exactly, as you've stated - using call with sync function  simplifies tests - when using call we don't to mock so much, it's enough to check yielded effects' descriptors.

If it's not bad of forbidden to use `call` with standard functions then maybe we could amend the docs @yelouafi?

Andarist:
Well, @yelouafi himself has proposed wrapping simple function into `call` effect here - https://github.com/yelouafi/redux-saga/pull/464#discussion_r73314259 for testability purposes and its part of the examples in the docs, so I would say its recommended all the way.

And also `runCallEffect` supports it directly - https://github.com/yelouafi/redux-saga/blob/e18dfa2592b493d91b82ef68d026e311827a34ba/src/internal/proc.js#L417 .

chodorowicz:
@Andarist thanks a lot for confirmation and pointing to the place in code! 🙇

PR #517, opened 2016-08-31
Improving API docs for call effect

rauliyohmc:
Stating that normal functions can be used as well (as per [#515](https://github.com/yelouafi/redux-saga/issues/515))


docs/api/README.md
```diff
@@ -290,6 +290,9 @@ If the result is a Promise, the middleware will suspend the Generator until the
 resolved, in which case the Generator is resumed with the resolved value. or until the Promise
 is rejected, in which case an error is thrown inside the Generator.
 
+If the result is not an Iterator object nor a Promise, the middleware will immediately return that value back to the saga, 
+so that it can resume its execution.
```
Andarist: maybe add here `synchronously` on the end of the sentence?

yelouafi:
thanks both!

PR #519, opened 2016-09-01
fix multiple default exports in shopping-cart example

mheiber:
Fixes this problem with `npm run shop`

` SyntaxError: Only one default export allowed per module.`

screen shot 2016-09-01 at 12 48 22 pm


mheiber:
whoops, duplicate of #516

PR #529, opened 2016-09-05
[bugfix] log() compatible with IE9

cyrilluce:
In Internet Explorer 9
`typeof console.info.call === 'undefined'`
Any cancel effect will call log('info', ...) finally cause exception and break tasks.


yelouafi:
thanks for the fix!

PR #530, opened 2016-09-05
Implemented new type of buffer - infinity, which will expand itself a…

Andarist:
…utomatically in case of overflow

Would give people new type of self-expanding buffer as alternative (for example here - https://github.com/yelouafi/redux-saga/issues/511), more close to old (non ring) implementations, which got recently introduced as breaking change.


src/internal/buffers.js
```diff
@@ -29,6 +36,11 @@ function ringBuffer(limit = 10, overflowAction) {
             pushIndex = (pushIndex + 1) % limit
             popIndex = pushIndex
             break
+          case ON_OVERFLOW_ACCOMODATE:
+            doubledLimit = 2 * limit
+            arr.length = length = limit = doubledLimit
```
yelouafi: length shouldn't be modified because it denotes the number of items in the array not he length of the array

yelouafi:
this `Infinity` name looks scary :). I think a name like `dynamic` or `expanding` is more appropriate

But I think there is an issue with the implementation, expanding the limit of the ring buffer will create a *hole* on the ring between actual limit and the new limit; while what you want is to create more room between the pop index and the push index.

I think a more simple implementation is to create a new array, and copy all the items (between the pop and push indexes) into it

Andarist:
Changed the scary name to `expanding` ;)

Also changed implementation along your suggestion, so new array is created upon expanding.

src/internal/buffers.js
```diff
@@ -29,36 +54,32 @@ function ringBuffer(limit = 10, overflowAction) {
             pushIndex = (pushIndex + 1) % limit
             popIndex = pushIndex
             break
+          case ON_OVERFLOW_EXPAND:
+            doubledLimit = 2 * limit
+
+            flushedItems = flush()
+            arr = [].concat(flushedItems)
```
yelouafi: Do we need to create a new Array here? `flush` already returns a new array

Andarist: you are right, gonna remove that line in a minute

Andarist:
Adjusted things as you have suggested

yelouafi:
thanks

Issue #441, opened 2016-07-18
Weird (incorrect?) error handling behaviors


dieppe:
Hello,

We've been enjoying using `redux-saga` for a month now, and we're quite happy with it (so, thank you for that 😄 ). 

There is something that we cannot figure out though: error bubbling and how to properly handle it in the root saga. (our goal is to have our root saga handle all the uncaught errors and restart the app when that's the case)

As a disclaimer: I don't know if all the following is expected behavior or not (I guess not). But even if it is, it is really confusing and we could not find anything in the docs related to that.

Anyway, here it goes:

I've noticed the following error handling (/bubbling) patterns:

```js
try {
  yield fork(...);
}
catch (error) { ... }
```

```js
try {
  const task = yield fork(...);
  yield task.done;
}
catch (error) { ... }
```

and 

```js
const task = yield fork(...);
try {
  yield task.done;
}
finally {
  const error = task.error();
}
```

I'll post different scenarii, all quite similar, but with different error bubbling behaviors (except scenario 2, for which I found none) .

(all these have been run against `redux-saga@0.11.0` -- transformed with `babel@6.10.1` and react native packager)

---

**Scenario 1**: Immediately throwing in a forked function
```js
// this is not a generator but fork expect a generator or a "normal" function that returns a Promise.
function immediatelyFailingFork () {
  throw new Error('immediatelyFailingFork');
}

export default function* root () {
  const error = yield fork(immediatelyFailingFork);
  console.log(error); // logs the error properly
}
```

Note: we don't fork a generator, but a function.
Expected behavior: a task is returned by `fork`.
Actual behavior: the error is returned by `fork`.

---

**Scenario 2**: Throwing in a function forked from a forked generator
```js
function immediatelyFailingFork () {
  throw new Error('immediatelyFailingFork');
}

function* launcher () {
  yield fork(immediatelyFailingFork);
}

export default function* root () {
  yield fork(launcher);
}
```

Note: same as Scenario 1, but we put the `fork` of the throwing function inside another `fork`.
Excepted behavior: Error bubbling to the `root` saga.
Actual behavior: The error is swallowed completely. No "uncaught at root" error is logged by `redux-saga` (Seems only logical since no task is returned in scenario 1, hence no error propagation can take place).

For this scenario, I tried the following error handling strategies:

**2.1**
```js
export default function* root () {
  try {
    yield fork(launcher);
  }
  catch (error) {
    console.log(error); // never called
  }  
}
```

**2.2**
[...]

Issue #391, opened 2016-06-16
Usage with global exception handler


jonaskello:
I'm using [zonejs](https://github.com/angular/zone.js/) to install a global exception handler that I want to catch all uncaught exceptions in my application (including async exceptions). It works fine for all but exceptions that happens in my sagas. I've been trying to troubleshoot this and It seems that redux-saga catches exceptions, logs them and then does nothing more. If my understanding of what happens in redux-saga is correct it means the exceptions are swallowed by redux-saga and never reaches the global exception handler. 

If the above is actually what happens, I am wondering if there is a rationale behind swallowing the exceptions in the sagas or if it can be changes to they are re-thrown so we can catch them at a higher level?

yelouafi:
You can catch uncaught exceptions from your root Saga

```js
const rootTask = sagaMiddleware.run(rootSaga)
rootTask.done.catch(...)
```

yelouafi:
But note that the method above works more as a crash reporting mechanism. i.e. when the error reaches that level, all sagas in the way are already cancelled.

jonaskello:
Ok, I'll try using that hook. Will it catch all errors including async errors?

yelouafi:
yes. all errors that are raised inside sagas are trapped

jonaskello:
Ok, thanks!

tadas-subonis:
@yelouafi this should be in the docs :-) :
```
const rootTask = sagaMiddleware.run(rootSaga);
rootTask.done.catch(function (err) {
    console.log("Error in Sagas", err);
});
```

slorber:
@yelouafi also we often see people using a single root saga starting other sagas with fork, call array or something similar

I'm not sure it's a good idea to let people use redux-saga this way because a single uncatched error anywhere will make all the sagas cancelled, leading sometimes to the app becoming unusable. We all know it's quite annoying when this happens (see error boundaries in React) so it may be worth advocating for a clever isolation of sagas inside of a single top-level saga

madebyherzblut:
@slorber is there already a gist/example where I can see this isolation in action?

slorber:
@madebyherzblut use spawn instead of fork/call, or run multiple times `runSaga` method

madebyherzblut:
Ah, thank you! I didn't know that I can run `runSaga` more than once.

Issue #343, opened 2016-05-23
How to catch global saga errors


hoschi:
I know how to handle expected errors within a saga, this is not the problem.

I have effects which can produce errors, which should "never" happen. E.g fetch user profile information. This ajax call only fails if the api server isn't running.

These errors are handled in
https://github.com/yelouafi/redux-saga/blob/ae39cf06dee57ddd3db935a0023db5f584eb1748/src/internal/proc.js#L243
which logs the error to the console. Is this overwritable to e.g. dispatch an action to show the error to the user or send it to a error logging backend API?

pavelkornev:
Have you tried like so:
```
store.runSaga(rootSaga).done.catch((e) => {
  console.log(e.message)
})
```

yelouafi:
I think what @hoschi  is asking is a 'generic' way to handle errors inside tasks. Catching the error on the root saga will abort all the currently executing tasks on the execution tree. The same observation holds for the above snippet in proc.js, when you reach that point it's too late because all the tasks in the way have already been canceled.

redux-saga follows the same synchronous style for handling errors, so in order to handle some errors in a generic way you'll have to implement a generic saga for handling those errors and then call it inside your catch blocks. A better way IMO, is to not use try/catch to handle application errors and use explicit return values instead. JavaScript hasn't good support for try/catch as it doesn't provide Typed catch blocks. so your catch block can swallow also application bugs (e.g. undefined var)

hoschi:
It is OK when the hell breaks loose and nothing works at this moment in the browser. I *just* want to send as much information as I can above this error to a logging service to:
* know I have a bug
* get information about the bug

When the solution of @pavelkornev catches the error before the logger in proc.js, this is the "good enough" solution for my use case.

Initially I thoght it would be more complicated to catch these kind of errors and I wanted to reuse the existing code of proc.js, but replace `console.log` with `sendToBackendLogger`. Perhaps this was confusing.

hoschi:
This works in the way mentioned, but good enough for me, thanks.

Issue #250, opened 2016-04-07
Handling uncaught exceptions


axelson:
So I am working on a project where there are many sagas and sometimes they result in errors. In development the errors get logged to the console (which is quite nice), but in production I'd like to be able to capture the exception and send it to an error monitoring service (such as honeybadger or getsentry), is there a way to do this?

The only thing I can think of is instead of passing many sagas to `createSagaMiddleware` to only pass a single root saga and to manually capture the error in the root saga after forking every other saga. Is there another approach that I could use?

ronanwize:
@axelson We are using an approach where our sagas will dispatch error actions to the redux store. We then have the option to use middleware to log errors, or to use other sagas to catch that and send the errors to an error monitoring service (we use sentry).

A dead simple example is like:

```
function* doSomething({payload}) {
  const output = yield call(myLongRunningFunction, payload);

  if(output.error) {
    yield put({type: 'DO_SOMETHING_ERROR', payload: output.error });
  }
}

function* doSomethingErrorWatcher() {
  while(true) {
    const errAction = yield take('DO_SOMETHING_ERROR');
    yield call(myFuncThatPostsToSentry, errAction);
  }
}
```

Naturally our code is alot more complex than that but that's the basic idea. We actually have two layers that do this, one that runs within our sagas and another that is redux middleware but that is because we have specific business requirements to log these in different places.

axelson:
@ronanwize that's a very interesting approach! Thanks for sharing. But wouldn't this still not catch uncaught exceptions in your `doSomething()` method? For example if the `myLongRunningFunction()` did not exist, than no error would be reported.

yelouafi:
@axelson For `fork`ed sagas you can use the `done` property to catch errors

```javascript
function* wrap(saga, ...args) {
  const task = yield fork(saga, ...args)
  task.done.catch(reportTheError) // handle the error
  return task
} 

// wrap the fork effect
const wrapFork = (saga, ...args) => call(wrap, saga, ...args)

// then use it instead of fork
function* mySaga() {
  while(true) {
    const action = take(ACTION)
    yield wrapFork(doSomething, action)
  }
}
```

you can do the same thing with `call`ed sagas by putting the call in a try/catch then rethrow the error.

ronanwize:
[...]

PR #533, opened 2016-09-08
Added option for handling uncaught errors

secobarbital:
Adds support for an `onerror` option to `createSagaMiddleware` which is very similar to the `logger` option except it receives the actual `Error` object instead of the massaged strings that the former receives.

This is especially useful for sending uncaught exceptions to error tracking services, as requested in #250, #343, #391, #441, etc.


src/internal/proc.js
```diff
@@ -244,6 +244,9 @@ export default function proc(
       iterator._deferredEnd && iterator._deferredEnd.resolve(result)
     } else {
       if(result instanceof Error) {
+        if(onerror) {
```
yelouafi: I think we should put this inside the below `if(!task.cont) {` block. Because here it will notify all errors not just uncaught ones; for example

```js
function* child() {
  const {err, res} = yield call(fetch, 'url')
  if(err)
   throw new Error(err)
}

function* parent() {
  try {
    yield call(child)
  } catch(err) {
     // you don't want to report this one
  }
}
```

secobarbital: Ok I will move it down. Do we need to check for `instanceof Error` in there as well then? What other kinds of `result` can we expect here?

yelouafi:
looks fine to me. Esp. tests are exhaustives. Thanks!

PR #534, opened 2016-09-09
Added API docs about the flush effect

Andarist:


docs/api/README.md
```diff
@@ -539,6 +539,32 @@ function* takeOneAtMost() {
 }
 ```
 
+### `flush(channel)`
+
+Creates an effect that instructs the middleware to flush all buffered items from the channel. Flushed items are returned back to the saga, so can they be utilized.
```
smashercosmo: Probably, "so they can be"

Andarist: What do you mean? Just indicating that they can be utilized by the user if he wants to and thats it not mandatory?

smashercosmo: I mean the word order. I could be wrong, cause I'm not a native speaker) but I thought that verb should go after subject.

Andarist: ah, ofc, you are right, didnt catch it while second reading, thanks!

yelouafi:
thanks

PR #539, opened 2016-09-10
Extracted stdChannel as thin wrapper of eventChannel, so only it has …

### LINKED ISSUE #507: [BUG] take defined with matcher will allways take from the buffer
Snippet to reproduce
```
	const chan = channel()
	yield put(chan, { type: 'OTHER_ACTION' })
	yield fork(function* () {
		while (true) {
			const action = yield take(chan, 'TEST')
			console.info('action', action)
		}
	})

	yield put(chan, { type: 'TEST' })
	yield put(chan, { type: 'JUST_ANOTHER_ACTION' })
```

If `buffer` is not-empty during a `take` effect channel will always `take` from it, completely ignoring the `matcher` argument. 

Possible solution? implementing `matcher` arugment in `buffer.take`?

Andarist:
…the logic of taking based on pattern. (fixes #507)


src/internal/io.js
```diff
@@ -16,29 +16,14 @@ const FLUSH  = 'FLUSH'
...
-        channel = null
-      } else {
-        throw new Error(`take(patternOrChannel): argument ${String(channel)} is not valid channel or a valid pattern`)
-      }
-    } else {
-      pattern = '*'
-    }
-  } else {
-    pattern = '*'
+export function take(patternOrChannel = '*') {
```
yelouafi: I dont like much the default parameter here. Because it may swallow typo errors. Adding a `check(patternOrChannel, is.notUndef, ...)` could provide better error detection.

Andarist: ok, have added a check for that
```
  if (arguments.length) {
    check(patternOrChannel, is.notUndef, 'take(patternOrChannel): patternOrChannel is undefined')
  }
```

yelouafi: but w've still left the `patternOrChannel = '*'`. which will assign a '*' to the argument if it's undefined.

The actual code will transpile to 

```js
function take() {
  var patternOrChannel = arguments.length <= 0 || arguments[0] === undefined ? '*' : arguments[0];

  if (arguments.length) {
    check(patternOrChannel, is.notUndef, 'take(patternOrChannel): patternOrChannel is undefined');
  }
}
```

The default value will be assigned before our check has a chance to run

Andarist: Oh, missed that as I've focused on checking if default argument has an impact on `arguments.length`. Could change `check(patternOrChannel, ...` to `check(arguments[0], ...`, but dont know how much you like such solution.

In previous version of `take` was defaulting too to `*`, so I just wanted to keep that behaviour cleanly in the code and it seems better to make it a default parameter than to override argument later. 

Just checked againt behaviour for some scenarios:
```
// function test(a) {}
test() // arguments.length === 0
test(undefined) // arguments.length === 1

// function test2(a = '*') {}
test() // arguments.length === 0, a === '*'
test(undefined) // arguments.length === 1, a === '*'
```

yelouafi: Imagine this scenario
```js
import module from '...'

yield take(module.matchActionn) // typo should be matchAction
```
in the above example, `module.matchActionn` will resolve to `undefined` and instead of reporting the error will assign the argument a value of `*`. that's because default parameter will be assigned (since arguments[0] is undefined) before the check for undefined is reached.

yelouafi: oops! sorry forget my last comment. I think checking arguments[0] should be ok

Andarist: 🎉  its already pushed out

yelouafi:
thanks for all!

Issue #540, opened 2016-09-12
[Question] Serial async 


km-tr:
```js
function* serial() {
  yield ['hoge', 'huga', ...].map( (item) => call(xxxfunc
, item) );
}
```
↑this parallel.

I want to do serial(series？) from array.
What should i do?

Andarist:
I would recommend similar thing like I've proposed in other question thread - https://github.com/yelouafi/redux-saga/issues/532#issuecomment-245337343

```
function* serial() {
	// map your data to effects (which are just plain objects)
	const effects = ['hoge', 'huga', ...].map( (item) => call(xxxfunc, item) );
	// yield those created effect-objects to the saga, so it can run them all in parallel
	yield effects
}
```

km-tr:
Thank you!

The point is, does this mean I can not to be serial?

Andarist:
Oh sorry. Ive misunderstood your intention. I guess then just regular loop
will do fine. Just iterate through your array

PR #542, opened 2016-09-14
Added new type of effect - waterfall. Can be used to run consecutive …

Andarist:
…effects (implements #540)

Aint sure how to attach `.cancel` on this effect callbacks. Would appreciate guiding through that.


yelouafi:
Do we need a dediczted effect here? I think this could be simply implemented in userland

Andarist:
Thought about it only as a helper as its ofc trivial to implement it. Got kinda requested here - #540. Feel free to close if you think it is an overkill

yelouafi:
I prefer to keep the list of effects as minimal as possible on the core. Perhaps we should have this in a separate repo.

PR #564, opened 2016-10-02
Throwing error, when using anonymous functions in 'takeEvery', 'takeLatest'

invibot:
[+] Added: error is thrown when using anonymous functions in 'takeEvery', 'takeLatest'

### DIFF
--- a/src/internal/sagaHelpers.js
+++ b/src/internal/sagaHelpers.js
@@ -43,6 +43,9 @@ function safeName(pattern) {
 }
 
 export function takeEvery(pattern, worker, ...args) {
+  if(!worker.name) {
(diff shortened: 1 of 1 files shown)


Andarist:
but why should such thing be constrained at this level? more appropriate place for such a constraint would be an `eslint` rule

Andarist:
You would have to write your own plugin, possibly this might be a good starting resource for learning about how to do such a thing - https://www.kenneth-truyers.net/2016/05/27/writing-custom-eslint-rules/

yelouafi:
As @Andarist noted, this seems out of the scope of the library. Could you motivate why this shouldn't be allowed?

Issue #550, opened 2016-09-19
Do not bubble Errors from Put Effects


michaelgilley:
I came across an issue where a reducer was expecting a property on action data that wasn't there and so it threw an error but whenever this occurs already inside the catch statement of a saga it causes that saga, and its parents, and subsequently their children to all abort(!). This led me to question why Put effects are bubbling up errors to begin with since (it seems to me) the effect should act like an event emitter on the Redux store bus. If something throws downstream shouldn't it be captured and logged but not bubble back into the saga?

**Example:**
```js
function * watcherSaga () {
  yield * takeLatest(FOO, workerSaga)
}

function * workerSaga (action) {
  try {
    const res = yield call(xhr, {foo: action.data.foo})
    yield put(actions.fooSuccessful(res.body)
  } catch (err) {
    // uncaught thrown error here will cause the entire app to crash!
    yield put(actions.fooFailure(err)
  }
}

export default reducer ({type, data}) {
  switch (type) {
    case FOO_FAILURE:
      return state.merge({
        // this will throw and flow back into the workerSaga
        foo: data.oops.not.here
      })
    default:
      state
  }
}
```

I propose that the [runPutEffect](https://github.com/yelouafi/redux-saga/blob/master/src/internal/proc.js#L443) function handle errors much like the [Cancel Effect](https://github.com/yelouafi/redux-saga/blob/master/src/internal/proc.js#L443) by logging out any errors but returning the cb without the error. While the reducer *should* handle these error cases on its own it would still be a good defensive check to keep the entire app from cancelling.

Thoughts?

michaelgilley:
@yelouafi Any thoughts on this?

yelouafi:
Yeah it makes sense if the effect is non blocking.

OTOH, what could be the alternative path for an app (other than crashing)?

michaelgilley:
If other parts of the app do throw that will indeed create problems. However, when using a more complex collection of combined reducers only the sub-reducer is failing so there is some grace there. So of course the issue that is causing an error needs addressing but the referenced PR makes the difference between a portion of the app failing and the entire saga tree falling down.

slorber:
Agree that error in action listeners of store (middleware/reducer/other...) should not crash sagas

Inspired from backend distributed systems, when a message is published to the log/bus, it can't be removed. Other systems consume the message from the log/bus, and if they fail, it should have no impact on the system that emitted the message in the first place (in this case the saga that did the put effect)

[...]

PR #567, opened 2016-10-03
Don't cancel saga when put dispatch throws

michaelgilley:
Ref #550 

When putting onto the store via dispatch the operation should be considered an emit on the Redux store that does not bubble up errors from reducers/helpers/etc. subsequently cancelling the saga.

This patch simply logs such errors and renders the put as a noop.

### DIFF
--- a/src/internal/proc.js
+++ b/src/internal/proc.js
@@ -442,7 +442,7 @@ export default function proc(
       } catch(error) {
         // If we have a channel or `put.sync` was used then bubble up the error.
         if (channel || sync) return cb(error, true)
-        log('error', `uncaught at ${name}`, error.stack)
+        log('error', `uncaught at ${name}`, error.stack || error.message || error)
       }
 
       if(sync && is.promise(result)) {
--- a/src/internal/proc.js
+++ b/src/internal/proc.js
@@ -440,7 +440,8 @@ export default function proc(
       try {
         result = (channel ? channel.put : dispatch)(action)
       } catch(error) {
-        return cb(error, true)
+        if (channel) return cb(error, true)
+        log('error', `uncaught at ${name}`, error.message)
       }
 
       if(sync && is.promise(result)) {
--- a/src/internal/proc.js
+++ b/src/internal/proc.js
@@ -441,7 +441,7 @@ export default function proc(
         result = (channel ? channel.put : dispatch)(action)
       } catch(error) {
         if (channel) return cb(error, true)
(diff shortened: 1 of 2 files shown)


yelouafi:
It makes sense as the put effect is non blocking. A couple of remarks.

We could make the handling more consistent. Right now the put effect can be either blocking (using `put.sync(...)`) or non blocking (the default, using `put(...)`). So what I propose is

- If the put is non-blocking (`sync == false`) then we do not bubble the error, just log it as in the PR.

- If the put is blocking (`sync == true`) then keep the actual behavior.

For logging, I noticed that logging `error.stack` gives better error logging in console.chrome. Relying on console.error (invoked by `log('error', ...)`) doesn't show where the error exactly occured (in this example inside the `checkoutStatus` reducer of the shopping cart example)

err-1

logging error.stack explicitly shows the orignial stack of the error

err-2

michaelgilley:
I agree on both accounts. I have updated the PR.

A few notes however:

1. This behavior should be documented
1. `put.sync` should be added to the documentation
1. Logging is [inconsistent](https://github.com/yelouafi/redux-saga/blob/master/src/internal/proc.js#L353) and at times unhelpful (as you pointed out here). I vote we make error reporting more robust.

src/internal/proc.js
```diff
@@ -441,7 +441,7 @@ export default function proc(
         result = (channel ? channel.put : dispatch)(action)
       } catch(error) {
         if (channel) return cb(error, true)
-        log('error', `uncaught at ${name}`, error.message)
+        log('error', `uncaught at ${name}`, error.stack)
```
yelouafi: I think we should check if `error instanceof Error` because dev. could throw other things (like `throw "some error message"`

yelouafi:
>This behavior should be documented

Would you like to add docs as part of this PR ?

>put.sync should be added to the documentation

Yeah! forgot it 😅

>Logging is inconsistent and at times unhelpful (as you pointed out here). I vote we make error reporting more robust.

You're right. Do you have any other suggestions to improve error reporting ? (you can open a separate issue for this)

src/internal/proc.js
michaelgilley: Yeah, this is the sort of thing that should be handled by the logger. For now I'll add a check though.

michaelgilley:
I've add error handling for put in the API docs and I'll raise the log reporting in another issue.

yelouafi:
Looks fine. Thanks!

PR #571, opened 2016-10-05
Add example for writing takeEvery/takeLatest tests

michaelgilley:
It took a little while to find how to test `takeEvery` or `takeLatest` for better coverage so I added an example here.

### DIFF
--- a/examples/counter/test/sagas.js
+++ b/examples/counter/test/sagas.js
@@ -1,8 +1,9 @@
 import test from 'tape';
 
+import { takeEvery } from '../../../src'
 import { put, call } from '../../../src/effects'
 import { delay } from '../../../src'
-import { incrementAsync } from '../src/sagas'
+import rootSaga, { incrementAsync } from '../src/sagas'
 
 test('incrementAsync Saga test', (t) => {
   const generator = incrementAsync()
@@ -23,6 +24,21 @@ test('incrementAsync Saga test', (t) => {
     generator.next(),
     { done: true, value: undefined },
     'counter Saga must be done'
+  )
+
+  const watcher = rootSaga()
+  const worker = takeEvery('INCREMENT_ASYNC', incrementAsync);
+
+  t.deepEqual(
+    watcher.next().value,
+    worker.next().value,
+    'rootSaga takes from every INCREMENT_ASYNC action'
+  )
+
+  t.equal(
+    watcher.return().done,
+    true,
+    'rootSaga finishes if canceled (returned)'
   )
 
   t.end()


examples/counter/test/sagas.js
```diff
@@ -25,5 +26,20 @@ test('incrementAsync Saga test', (t) => {
...
+  const worker = takeEvery('INCREMENT_ASYNC', incrementAsync);
+
+  t.deepEqual(
+    watcher.next().value,
+    worker.next().value,
+    'rootSaga takes from every INCREMENT_ASYNC action'
+  )
+
+  t.equal(
+    watcher.return().done,
```
Andarist: isnt this one testing effectively `true === true`? 

this `watcher.return().done` will always evaluate to `true`

michaelgilley: In this specific case it does but in other sagas (with try/catch/finally) this would not be true. So in this case the test asserts that takeEvery is **not** wrapped in a finally block.

On another note, this is especially done to increase coverage. If you want to get even better coverage than this one would need to run a `saga.throw(err)` assertion as well. I left this off here though as a bit too verbose.

michaelgilley: Am I missing something? I went in to update this and the tests are still passing. It looks like takeEvery is still returning an iterator. Do I have something wrong??

Andarist: Hm, you are right, but there is a caveat. Look at what is returned by `takeEvery`:
```
{ name: 'takeEvery(INCREMENT_ASYNC, incrementAsync)',
      next: [Function: next],
      throw: [Function],
      return: [Function: kReturn],
      '@@redux-saga/HELPER': true }
```

And now because of this - ` '@@redux-saga/HELPER': true` its gonna be interpreted by the `redux-saga` in the special way:
`: is.helper(effect)                                    ? runForkEffect(wrapHelper(effect), effectId, currCb)`

So effectively in real saga the task descriptor is going to be injected into the saga, and not an iterator.

yelouafi:
>And now because of this - '@@redux-saga/HELPER': true its gonna be interpreted by the redux-saga in the special way:
: is.helper(effect) ? runForkEffect(wrapHelper(effect), effectId, currCb)
So effectively in real saga the task descriptor is going to be injected into the saga, and not an iterator.

I'm thinking if we may use a declarative form for the helpers as well. This way we can test them the same way we test other effects.

could be something like

```js
// helper
function* takeEveryHelper (pattern, worker) { ... }

//effect creator: exposed to the user
const takeEvery = (pattern, worker) => fork(takeEveryHelper, pattern, worker)
```

michaelgilley:
@yelouafi You're proposing that the `takeEvery` and `takeLatest` helpers be made available as effects only?

yelouafi:
@michaelgilley yes. I dont see any reason to keep the iterator approach.

michaelgilley:
Closing in favor of #596

PR #583, opened 2016-10-13
Fixed issue when more than 2 sync effects were geting resolved during…

### LINKED ISSUE #578: Race effect when used with concurrent effects does not cancel alternate tasks
I'm seeing tasks not cancelled when using `race` with tasks that (sometimes) returns immediately. In my case I have the following:

```js
yield race({
  route: take(myChannel),
  move: take(MOVE_ACTION)
});
```

If `myChannel` has items in it's buffer, then the listener for `MOVE_ACTION` is never cancelled, which breaks things further down the line when I call the race again (I have to trigger `MOVE_ACTION` twice, first one to clear away the old listener).

Andarist:
… the race (fix #578)


yelouafi:
Thanks for the fix. Seems ok to me.

[Zalgo](http://blog.ometer.com/2011/07/24/callbacks-synchronous-and-asynchronous/) keeps resurging :)

PR #584, opened 2016-10-13
Call cancel on CPS function cancellation

Alxandr:
According to comments in the source, cps functions should be able to add their
cancellation logic. However, that does not work at the moment. This pull-request
fixes that.


Andarist:
Thanks for the PR, I'll hopefully have some time in the weekend to look into it.

PR #585, opened 2016-10-13
Fix code block on recipes page.

knpwrs:
This renders fine on GitHub but not on GitBook.


Andarist:
Thanks for noticing and fixing right away!

PR #588, opened 2016-10-14
Fix broken links and minor grammar error

dooart:
- There were two broken links to the "buffers" section of API
- "putted" -> "put"


Andarist:
Thanks for fixes! Highly appreciated. 🎉

PR #590, opened 2016-10-17
[Typescript] SagaIterator: Iterable -> IterableIterator

jupl:
Currently when I have a saga like this:

```ts
function* saga(): SagaIterator {
  for(;;) {
    const {end} = yield race({
      timeout: call(delay, 4000),
      end: take([`${previousColor}`, `${nextColor}`]),
    })
    if(end) { break }
    yield put(autoNextColor())
  }
}
```

The following results in `next` being undefined when testing:

```ts
const generator = saga()
generator.next() // Property 'next' does not exist on type 'Iterable<TakeEffect<any> | ...>'.
```

Using `IterableIterator` instead of `Iterable` works as the type of a regular generator function is an `IterableIterator`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,7 +4,7 @@ import {Task, Buffer, Channel, Predicate} from "./types";
 
 export {Effect, Pattern, Task, Buffer, Channel, Predicate};
 
-export type SagaIterator = Iterable<Effect|Effect[]>;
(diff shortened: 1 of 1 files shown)


Andarist:
@aikoven could you take a look on it?

aikoven:
👍

PR #593, opened 2016-10-19
Adding $$LogSagas to global in react-native

benhughes:
Quick little check so the $$LogSagas can be added while debugging react-native

### DIFF
--- a/examples/sagaMonitor/index.js
+++ b/examples/sagaMonitor/index.js
@@ -15,6 +15,12 @@ const CANCEL_STYLE = 'color: #ccc'
 
 const IS_BROWSER = (typeof window !== 'undefined' && window.document)
 
+const globalScope = (
+  typeof window.document === 'undefined' && navigator.product === 'ReactNative' ? global
+  : IS_BROWSER ? window
+  : null
+)
+
 // `VERBOSE` can be made a setting configured from the outside.
 const VERBOSE = false
 
@@ -388,8 +394,8 @@ const logSaga = () => {
 }
 
 // Export the snapshot-logging function to run from the browser console or extensions.
-if(IS_BROWSER) {
-  window.$$LogSagas = logSaga
+if(globalScope) {
+  globalScope.$$LogSagas = logSaga
 }
 
 // Export the snapshot-logging function for arbitrary use by external code.


examples/sagaMonitor/index.js
```diff
@@ -14,6 +14,7 @@ const ERROR_STYLE = 'color: red'
 const CANCEL_STYLE = 'color: #ccc'
 
 const IS_BROWSER = (typeof window !== 'undefined' && window.document)
+const IS_REACT_NATIVE typeof window.document === 'undefined' && navigator.product === 'ReactNative'
```
Andarist: you are missing an assignment here

examples/sagaMonitor/index.js
```diff
@@ -392,6 +393,10 @@ if(IS_BROWSER) {
   window.$$LogSagas = logSaga
 }
 
+if (IS_REACT_NATIVE) {
```
Andarist: could you refactor it, so both `IS_BROWSER` and `IS_REACT_NATIVE ` cases are handled at the same time? some kind of obtaining global object at the top and checking here if it got obtained

benhughes:
How about this?

examples/sagaMonitor/index.js
```diff
@@ -13,8 +13,13 @@ const EFFECT_TYPE_STYLE = 'color: blue'
 const ERROR_STYLE = 'color: red'
 const CANCEL_STYLE = 'color: #ccc'
 
+let globalScope = null;
+if (typeof global !== 'undefined')
```
Andarist: could be made `const` with ternary operator like so:
```
const globalScope = (
  typeof window.document === 'undefined' && navigator.product === 'ReactNative' ? global
  : (typeof window !== 'undefined' && window.document) ? window
  : null
)
```

benhughes:
Done, not a massive fan of nesting ternary expressions,, but seems to work OK.

Andarist:
@benhughes 

Cool, could be merged, but would prefer if you squash those 4 commits into one, to keep the history cleaner. Additionally you can rebase too. Thanks in advance!

benhughes:
Done, if any issues are raised from this please feel free to assign the issues to me. Thanks :)

Andarist:
Thanks!

PR #596, opened 2016-10-20
Added declarative effect creators for helpers

Andarist:
cc @yelouafi 

todo:
- [x] tests
- [x] docs


yelouafi:
Besides making testing helpers more trivial. This seems to make the code simpler (getting rid of helper related functions). I think this also would reduces confusion around helpers vs effects, yield* vs yield...

But as this will be a breaking change (yield* will no longer work) I think we should let the old exports from the helpers module with a deprecation warning for at least one minor version before removing it.

Andarist:
Yeah, that's probably the good idea with depraction working, gonna figure it out how to handle it.

granmoe:
I agree with @yelouafi 's comments. I've been teaching this out at work and one of the biggest points of confusion is that the helpers *look* like the effects but are not effects and are imported from a different location. Also agree with the point about testing.

Andarist:
@yelouafi 
Updated the code, please take a look, if everything is fine Im gonna fix the tests and prepare docs and merge to the master and possibly draft a new release.

yelouafi:
looks fine to me

src/internal/sagaHelpers.js
```diff
@@ -94,3 +95,11 @@ export function throttle(delayLength, pattern, worker, ...args) {
     q4() { return ['q2', yDelay] }
   }, 'q1', `throttle(${safeName(pattern)}, ${worker.name})`)
 }
+
+const deprecationWarning = (helperName) =>
+`import ${helperName} from 'redux-saga' got deprecated in favor of import ${helperName} from 'redux-saga/effects'.
```
ianmetcalf: `got deprecated` -> `is deprecated`

ianmetcalf: I had just noticed this was inconsistent with the `put.sync` deprecation warning above
```js
put.sync = deprecate(put.resolve, 'put.sync is deprecated in favor of put.resolve, please update your code')
```

PR #599, opened 2016-10-20
Clarify the argument to expanding buffers

mjrussell:
Currently docs suggest it expands dynamically to the limit provided, instead of its behavior which is to be initialized with a given size.


Andarist:
It might be a good idea to also reflect that in `src/internal/buffers.js`. Would you like to change it there as well?

mjrussell:
Sure thing

mjrussell:
Updated and rebased

Andarist:
Thanks!

Issue #598, opened 2016-10-20
Missed actions with take when using put


mjrussell:
So I've been struggling with trying to create an example for this and I haven't been able to do so, but I think the result of the saga monitor should be compelling enough to hopefully explain a bit about what I'm seeing.

I've been very careful in my app to avoid using blocking actions (like `call`) in the same saga as those that use `take`. And use helpers where I can. My understanding is that `put` is safe because it is non-blocking. However, I'm seeing an issue where a `put` appears to be taking so long to complete that it misses a subsequent `take` completely. To be fair, I could refactor my code to use `actionChannels` (this works quite well in this case). **However** it forces me to question my understanding of `put` and other places in my code where I use `put` and `take` together.

Here is the sagaMonitor which shows how I am missing a `take` because of the length of time it takes for the `put` to complete. Both sagas should get a `SOCKET_PUSH`, followed by a `SOCKET_SEND_PARSE_RESPONSE`, and then a final `SOCKET_PUSH`, but you can see the second one missed the `SOCKET_SEND_PARSE_RESPONSE` because the `put` took too long (185ms).

![image](https://cloud.githubusercontent.com/assets/1641895/19570013/e79764b4-96ac-11e6-822e-706866d252a6.png)

A little background info on my saga. The `appSocketForwardParser` has two types of messages it cares about `SOCKET_PUSH` and `SOCKET_SEND_PARSE_RESPONSE`. The `SOCKET_PUSH` come from another saga listening to a websocket `onMessage` and the `SOCKET_SEND_PARSE_RESPONSE` are dispatched via redux actions. Internally in the saga, we are building a map from to parse socket responses to redux actions, that can be added to dynamically with `SOCKET_SEND_PARSE_RESPONSE` messages which are then dispatched with `put`.

The only difference between `appSocketForwardParser` and `appSocketForwardParserCopy` is the following line which is the source of the `put` in the sagaMonitor:

``` js
yield put({ ...parsedAction, type: 'FOO' }); // original yield put({ ...parsedAction });
```

Essentially Im just clobbering the type of the parsedAction, so this redux action gets dispatched and nothing is "looking for it".

My guess right now is that `put` takes longer to return because this sets up a lot of initial data in my store and forces rendering of many components. There isn't another saga listening for this type, and I don't have any middleware that perform an action for this type either.

Thanks for the great, library and any help you can provide on this

Andarist:
[...]

Issue #413, opened 2016-06-27
Forked tasks can't always take actions from its siblings


axelson:
Original Reference: https://github.com/yelouafi/redux-saga/issues/277#issuecomment-228443152

I would expect this code to work as-is.

``` js
export default function * rootSaga () {
  yield fork(watchPing)
  yield put({type: 'PING', val: 0})
}

function * watchPing () {
  yield * takeEvery('PING', ackWorker)
  // yield * fixedTakeEvery('PING', ackWorker)
}

let first = true

function * ackWorker (action) {
  // yield Promise.resolve()
  if (first) {
    first = false
    yield put({type: 'PING', val: action.val + 1})
    yield take(`ACK-${action.val + 1}`)
  }
  yield put({type: `ACK-${action.val}`})
}

export function * fixedTakeEvery (actionType, saga) {
  const worker = function * (...args) {
    yield Promise.resolve()
    yield fork(saga, ...args)
  }
  yield * takeEvery(actionType, worker)
}
```

The above code results in PING, PING, ACK-1. But if you uncomment the yield Promise.resolve() or switch from takeEvery() to fixedTakeEvery() you will get the full sequence of PING, PING, ACK-1, ACK-0.

I tried to put this into an actual test case, but I'm getting different behavior (although the test is still failing).

``` js
test('inter-saga fork/take back from forked child 3', assert => {
  assert.plan(1);

  const actual = []
  const chan = channel()

  const middleware = sagaMiddleware()
  const store = createStore(() => {}, applyMiddleware(middleware))

  let first = true

  function* root() {
    yield [fork(watchPing)]
  }

  function* watchPing () {
    yield * takeEvery('PING', ackWorker)
    // yield * fixedTakeEvery('PING', ackWorker)
  }

  function* ackWorker (action) {
    // yield Promise.resolve()
    if (first) {
      first = false
      yield put({type: 'PING', val: action.val + 1})
      yield take(`ACK-${action.val + 1}`)
    }
    yield put({type: `ACK-${action.val}`})
    actual.push(1)
  }

  function* fixedTakeEvery (actionType, saga) {
    const worker = function * (...args) {
      yield Promise.resolve()
      yield fork(saga, ...args)
    }
    yield * takeEvery(actionType, worker)
  }


  middleware.run(root).done.then(() => {
    assert.deepEqual(actual, [1,1],
      "Sagas must take actions from each forked childs doing Sync puts"
    );
    assert.end();
  })

[...]

Issue #235, opened 2016-03-30
Solving Sync/Async dispatch tension with a global re-entrant lock


yelouafi:
## Problem

To handle the `put` effect (i.e. when a saga does `yield put(action)`) we actually have 2 possibilities

#### 1. Execute the dispatch synchronously (immediately)

```javascript
runPutEffect(action, next) {
  next( dispatch(action) )
}
```

#### 2. Execute the dispatch asynchronously (scheduling in a microtask using `Promise.then`

```javascript
runPutEffect(action, next) {
  Promise.resolve().then(() => next( dispatch(action) ))
}
```

Both approaches solves issues of each other. But not their proper issues

#### 1. The first one is *startup saga actions*. Actions which are fired by a Saga immediately after it is started.

Let's see a simple example

```javascript
function* saga() {
  yield put(action)
}

const store = createStore(
  reducer, 
  applyMiddleware(
    createSagaMiddleware(saga)
  )
)
```

The issue is that `createSagaMiddleware(saga)` will starts the Saga while still in the `applyMiddleware` phase, i.e. before the `applyMiddleware` returns. This will result in the middleware taking and executing the `put(action)` Effect. But we can't actually dispatches actions during middleware setup (see this issue https://github.com/reactjs/redux/issues/1240).

The immediate solution is to delay the startup of sagas until the store creation completes, but Redux do not actually have a lifecycle API for middleware (like with react) so the only way left is to start the Saga asynchronously using microtasks

```javascript
sagaMiddleware = (saga) => ({getState, dispatch}) => {
  Promise.resolve().then(() => startSaga(saga))
  return next => action ...
}
```

Now the saga startup will be delayed, and when the promise executes its then callback, the store creation will be already completed. so the dispatch can be executed without problem. But imagine now that the first instruction of `saga` was a `take`

```javascript
function* saga() {
  yield take(actionType)
}
```

If we delay the saga startup, then any actions dispatched synchronously after the store creation (e.g. by components is `componentDidMount`) will be missed by the saga, because it hasn't yet started and thus hasn't yet yielded the `take` effect. So we now have an issue with *startup store actions*.

We can go further and try to save all actions dispatched before saga startup in a temporary queue then delivers them once the saga starts on the next microtask but this will create other issues; plus there is another issue with synchronous dispatches

#### 2. The second issue is related to nested dispatches

[...]

PR #602, opened 2016-10-27
Provided failing test for issue #598

Andarist:
In the provided test I've used a simple store subscriber to cause a dispatch, but it can ofc be (more commonly) a React component's lifecycle method. 

I cant yet find a solution for this 'bug'. At first I thought that simple https://github.com/yelouafi/redux-saga/commit/8b126bb7f01ab285ed35e0f6821190bc1d43ea96 would help, but it broke other test - https://github.com/yelouafi/redux-saga/blob/8a76a1c7eeb97de58072b2e3ab721a59675d0365/test/proc/takeSync.js#L206 . 

So for now I have no idea how to satisfy both scenarios, maybe @yelouafi would think of smth

cc @mjrussell

### DIFF
--- a/test/proc/takeSync.js
+++ b/test/proc/takeSync.js
@@ -525,3 +525,46 @@ test('inter-saga fork/take back from forked child', assert => {
   store.dispatch({type: 'TEST'})
   store.dispatch(END)
 });
+
+test('put causing sync dispatch response in store subscriber', assert => {
+  assert.plan(1);
+
+  const actual = []
+
+  const reducer = (state, action) => action.type
+  const middleware = sagaMiddleware()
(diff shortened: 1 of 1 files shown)


yelouafi:
Didn't have time to look at this but the scheduler (asap) needs to be fixed (cf #413). Actually i use the asap helper to delay running put effects until the saga executing puts become 'stable' (ie ready to take, also cf #235 for some bg). But I think the issue is that the flush method actually doesn't handle nested queues correctly.

Andarist:
related I think - https://github.com/yelouafi/redux-saga/issues/604

Andarist:
Closing ss it got included in https://github.com/yelouafi/redux-saga/pull/622

PR #609, opened 2016-10-31
Improve monitor api

yelouafi:
I recently started working on a GUI version of the monitor. So far the change I see are required are

- notifies the monitor when root sagas are started by `sagaMiddleware.run` or `runSaga`.   

- added `actionDispatched` trigger to the monitor contract. We need this in order to figure out saga/take effects that reacted to a given action. Also added a silent flag `SAGA_ACTION` (non enum. prop) added to distinguish actions dispatched by sagas from others.

I'ld like to get some feedback before merging. I'll publish the UI version as soon as I get something usable

- [x] add changes to monitor API
- [x] update tests
- [x] update docs


src/internal/middleware.js
```diff
@@ -37,20 +37,24 @@ export default function sagaMiddlewareFactory(options = {}) {
   function sagaMiddleware({getState, dispatch}) {
     runSagaDynamically = runSaga
     const sagaEmitter = emitter()
+    const sagaDispatch = !sagaMonitor ? dispatch : wrapSagaDispatch(dispatch)
 
-    function runSaga(saga, ...args) {
+    function runSaga(saga, args, sagaId) {
```
yelouafi: No that's an internal function, the public one is declared on the bottom `sagaMiddleware.run = (saga, ...args) = ...`

examples/sagaMonitor/index.js
```diff
@@ -181,7 +180,14 @@ function logSimpleEffect(effect) {
...
   let data, log
 
-  if(data = asEffect.take(effect.effect)) {
+  if(effect.root) {
+    data = effect.effect
+    log = getLogPrefix('run', effect)
+    log.formatter.addCall(data.saga.name, data.args)
+    logResult(effect, log.formatter)
+  }
+
```
Andarist: kinda weird this `else if` being put after a whiteline, no place in the code follows that pattern (i think)

yelouafi: right. I did a global style fix on the file

src/internal/middleware.js
```diff
@@ -37,20 +37,24 @@ export default function sagaMiddlewareFactory(options = {}) {
   function sagaMiddleware({getState, dispatch}) {
     runSagaDynamically = runSaga
     const sagaEmitter = emitter()
+    const sagaDispatch = !sagaMonitor ? dispatch : wrapSagaDispatch(dispatch)
```
Andarist: wondering if distinguishing permanently regular `dispatch`es and `put`s could be leveraged to handle `asap` queue better in order to fix cases like - https://github.com/yelouafi/redux-saga/issues/598

yelouafi: hmm... dont think so. for #598 seems the external dispatch needs to be queued as well (which is not the case actually) but I think it's more related to the asap#isSuspended state. More fundamentally IMO what we need primarily is to come up with some unambiguous semantics for scheduling before tempting any impl.

yelouafi: After some thoughtsI think you're right. I might have some solution to asap which could fix both #598 (ie #602 failing test) as well as #4.
I'm going to merge this and try with the new scheduling idea

PR #610, opened 2016-11-02
Use spawn instead of fork in root saga and add fake error handling so…

granmoe:
… that a failed call doesn't make app unresponsive

### DIFF
--- a/examples/async/src/actions/index.js
+++ b/examples/async/src/actions/index.js
@@ -3,6 +3,8 @@ export const REQUEST_POSTS = 'REQUEST_POSTS'
 export const RECEIVE_POSTS = 'RECEIVE_POSTS'
 export const SELECT_REDDIT = 'SELECT_REDDIT'
 export const INVALIDATE_REDDIT = 'INVALIDATE_REDDIT'
+export const SET_ERROR = 'SET_ERROR'
+export const CLEAR_ERROR = 'CLEAR_ERROR'
 
 export function selectReddit(reddit) {
   return {
@@ -31,5 +33,18 @@ export function receivePosts(reddit, posts) {
     reddit,
     posts,
     receivedAt: Date.now()
+  }
+}
+
+export function setError(errorMessage) {
+  return {
+    type: SET_ERROR,
+    errorMessage
+  }
+}
+
+export function clearError() {
+  return {
+    type: CLEAR_ERROR
   }
 }
--- a/examples/async/src/containers/App.js
+++ b/examples/async/src/containers/App.js
@@ -22,12 +22,21 @@ class App extends Component {
   }
 
   render() {
-    const { selectedReddit, posts, isFetching, lastUpdated } = this.props
+    const { selectedReddit, posts, isFetching, lastUpdated, errorMessage } = this.props
     return (
       <div>
+        { errorMessage &&
+          <p>
+            Something bad happened:
+            <br/>
+            {errorMessage}
+            <br/>
+            Please try again
+          </p>
+        }
         <Picker value={selectedReddit}
                 onChange={this.handleChange}
-                options={[ 'reactjs', 'frontend' ]} />
+                options={[ 'reactjs', 'frontend', 'simulate network failure' ]} />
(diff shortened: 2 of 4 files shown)


granmoe:
To kill the app and see this issue, put a debugger right before the call to reddit and replace the URL with something that will fail.

Andarist:
One thing I dont understand - why did you change `fork`s to `spawn`s?

slorber:
@Andarist to avoid killing all the sagas on any single http call error

Related to https://github.com/yelouafi/redux-saga/issues/570

Andarist:
@slorber 
but that is already handled by `try/catch` over `yield call(fetchPosts, reddit)`

slorber:
That's right @Andarist :) still I believe sagas should be isolated from each others, it's easy to forget catching errors and make your whole app unusable

granmoe:
@Andarist @slorber Yes, I am handling all errors in the try/catch because the catch just swallows them and doesn't re-throw. I think to make the example more clear, I could handle certain types and then re-throw the rest. My intent here is to show some more nuanced error handling and clarify the fact that if you just fork all sagas, any unhandled exception will kill them all (this was a stumbling block for me in learning saga). I'm very very open to suggestions, but this example caused confusion for me when I was a newcomer to saga (which, granted, was only like two months ago).

granmoe:
Or maybe we should be using the `task.done.catch` feature in the root saga? Mainly, I just want this example to highlight a good, idiomatic pattern for the problem of top level exception handling / not killing all your sagas on an unhandled exception.

granmoe:
Would something like this work for centralized error handling in all sagas in an app?

```
function* rootSaga () {
  yield sagas.map(saga =>
    spawn(function* () {
      while (true) {
        try {
          yield call(saga)
        } catch (e) {
          console.log(e)
        }
      }
    })
  )
}
```

Or would it possibly cause complex bugs due to scheduling issues internal to redux-saga? I'm wondering if something could get messed up going from the catch back to the try. Assuming this all happens synchronously, there should be no problem, right? Or wrong?

slorber:
@granmoe uncatched errors are automatically logged as far as I know so it's not really worthy to catch just for logging unless you customize error message

There are multiple ways that you can do.

Personnally in my app I'm simply doing:

```
    sagas.forEach(saga =>
      runSaga(saga,reduxSagaIO)
    );
```

granmoe:
Sorry, my console.log is just an example (a confusing one)! We will be sending uncaughts through an error monitoring service. What is 'runSaga' and 'reduxSagaIO' in your example?

granmoe:
Nevermind, I found runSaga in the docs

granmoe:
I would still be curious to see your reduxSagaIO. And does this approach kill a saga that throws an uncaught exception? Or do you handle that in reduxSagaIO?

slorber:
@granmoe I have a special setting, because I actually make Redux-saga work on top of a legacy framework I have that is not redux (we'll migrate someday to redux)

granmoe:
OK, I pushed a new commit with my top level error handling solution. My only concern is if a situation happens, somehow, where calling the saga itself triggers an exception synchronously, how should that be handled? Maybe something like this to prevent infinite synchronous exceptions from a saga?

```
  yield sagas.map(saga =>
    spawn(function* () {
      let hasError = false
      while (!hasError) {
        try {
          setTimeout(() => { hasError = false }, 0)
          yield call(saga)
        } catch (e) {
          hasError = true
          handleError(e)
        }
      }
      throw new Error('Unrecoverable exception thrown while spawning saga', saga)
    })
  )
```

granmoe:
I'm kind of questioning this whole idea now. Maybe it would be better to rip out the `while (true)` that I have at the top here and instead just send the error to whatever logging/monitoring service you want to use and then re-throw. Handle network errors at the source and don't ever throw anything in a saga that you don't explicitly handle.

granmoe:
I'm going to close this for now and possibly come back to it. It would be great to get a consensus from the community regarding the right approach here.

PR #611, opened 2016-11-02
White spacing, use createAsEffectType helper

abroz:


src/internal/io.js
```diff
@@ -145,17 +145,19 @@ export function flush(channel) {
   return effect(FLUSH, channel)
 }
 
+const getAsEffectFn = type => effect => effect && effect[IO] && effect[type]
```
Andarist: Its not really a `get` but more of a `create` function here. I think `createAsEffectType` is a sufficient name here.

Andarist:
Commented on the code. Aditionally when you change it, it would be nice if you rebase to master + squash those commits.

abroz:
Done.

Andarist:
Thanks

PR #622, opened 2016-11-10
Fix scheduler issues of sync puts/forks/takes

### LINKED ISSUE #413: Forked tasks can't always take actions from its siblings

yelouafi:
Replace the previous lock based scheduler with a new one based on [semaphores](https://en.wikipedia.org/wiki/Semaphore_(programming)) (to handle nested schedules). 

Fixes #413 and #598 (via test case provided in #602)


src/internal/asap.js
```diff
@@ -1,21 +1,31 @@
...
+}
+
 asap.flush = () => {
-  let nextTask
-  while((nextTask = queue.shift())) {
+  //console.log('flush start')
+  semaphore--
+  while(!semaphore && queue.length) {
+    const nextTask = queue.shift()
+    semaphore++
```
Andarist: wondering if `asap.suspend()` shouldnt be used here for consistency, but then `semaphore--` maybe should get its own function, with more descriptive name, smth like - `asap.lift()`

Andarist:
Wow this is more than cool! 🎉 Superb important for the lib and handled gracefully! Just dont forget to erase those commented out `// console.log`s ;) Should I close PR#602?

cc @mjrussell

mjrussell:
👏 nicely done!

src/internal/asap.js
yelouafi: did some refactorings

yelouafi:
>Should I close PR #602 ?

Yes the test is already included in this PR (thanks to it the #598 case was greatly simplified)

src/internal/proc.js
```diff
@@ -487,22 +486,30 @@ export default function proc(
   function runForkEffect({context, fn, args, detached}, effectId, cb) {
     const taskIterator = createTaskIterator({context, fn, args})
 
-    asap.suspend()
-    const task = proc(taskIterator, subscribe, dispatch, getState, options, effectId, fn.name, (detached ? null : noop))
+    try {
+      /**
+        put the scheduler in the `suspended` state
+        this will prevent nested puts from perfoming until this Saga is ready to take them
```
Andarist: perfo*r*ming

src/internal/proc.js
```diff
@@ -487,22 +486,30 @@ export default function proc(
   function runForkEffect({context, fn, args, detached}, effectId, cb) {
     const taskIterator = createTaskIterator({context, fn, args})
 
-    asap.suspend()
-    const task = proc(taskIterator, subscribe, dispatch, getState, options, effectId, fn.name, (detached ? null : noop))
+    try {
```
Andarist: I see there is a new `try` here, but without corresponding `catch`, shouldnt something be done with the `error`? Also - what is it guarding from? `fn` throwing? 

this simple `try` can swallow errors easily and make debugging a nightmare

src/internal/scheduler.js
```diff
@@ -0,0 +1,53 @@
...
+let semaphore = 0
+
+/**
+  Executes a task 'atomically'. Tasks scheduled during this execution will be queued
+  and flushed after this task has finished (assuming the scheduler endup in a released
+  state).
+**/
+function exec(task) {
+  try {
+    suspend()
```
Andarist: same comment goes here for `try` without `catch` block

src/internal/proc.js
yelouafi: try/finally wont swallow errors since there is no swallowing catch. It simply ensures that the scheduler is always in a consistent state; every suspend must have a corresponding flush.

src/internal/proc.js
```diff
@@ -428,12 +428,11 @@ export default function proc(
   }
 
   function runPutEffect({channel, action, sync}, cb) {
-    /*
-      Use a reentrant lock `asap` to flatten all nested dispatches
-      If this put cause another Saga to take this action an then immediately
-      put an action that will be taken by this Saga. Then the outer Saga will miss
-      the action from the inner Saga b/c this put has not yet returned.
-    */
+    /**
```
yelouafi: yes I think it's redundant

src/internal/asap.js
Andarist: really nice job with that refactor, `scheduler` got much clearer now

Andarist:
No much to do here more if all tests are passing, merge at asap! (pun intended ;) )

PR #623, opened 2016-11-10
Add missing yield in fork model example

eiriklv:
From what I can see there should be a `yield` here, as `call` doesn't actually do anything without it - and it should be the same as the example above right?


Andarist:
Yep, thanks for noticing and sending PR! `call` alone is creating a wrapper around what should be called, its just a plain object (which we call effect) which can be treated specially by a `redux-saga`

PR #628, opened 2016-11-13
Fix inconsistent display of code snippet in docs

eiriklv:
Replace single quotes with backticks for correct display


eiriklv:
Made a mistake when creating the PR. Disregard.

PR #634, opened 2016-11-15
Permits saga monitors to only hook the functions they want.

skellock:
Since monitors are passive and observe-only, this PR makes it ok for monitors to listen to a subset of what's available.

Makes the interface a bit more backwards compatible as new things show up.

Cheers.

### DIFF
--- a/src/internal/middleware.js
+++ b/src/internal/middleware.js
@@ -1,4 +1,4 @@
-import { is, check, uid as nextSagaId, wrapSagaDispatch, SAGA_ACTION } from './utils'
+import { noop, is, check, uid as nextSagaId, wrapSagaDispatch, SAGA_ACTION } from './utils'
 import proc from './proc'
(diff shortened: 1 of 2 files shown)


yelouafi:
thanks

yelouafi:
BTW you may be interested to look at

https://github.com/yelouafi/redux-saga-devtool

it's still a WIP, but you may find soemething useful in the source for the reactotron saga plugin.

skellock:
I'll check that out shortly, thanks!  I saw in your PR that you were creating something like this.  That's awesome.

Reactotron is sort of general purpose, so I'll most likely not get as in-depth as what you're planning.

Same with Redux Dev Tools.  No way I'm aiming to replace that.  They're hard core.  I'm more interested in exploring additional logging with React Native.  :)

You're doing wonderful work here dude.  Keep it up.

PR #636, opened 2016-11-16
Missing a babel-register

minooo:
After run `npm install`, when I run `npm run real-world`，it will go wrong said that can not find `babel-register`,
when I add this package, it works well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
   "homepage": "https://github.com/yelouafi/redux-saga#readme",
   "dependencies": {},
   "devDependencies": {
+    "babel-register": "^6.18.0",
     "babel-cli": "^6.1.18",
     "babel-core": "^6.14.0",
     "babel-eslint": "^6.0.3",


Andarist:
Why would that be desirable?

minooo:
@Andarist when I run `npm run real-world`，it will go wrong said that can not find `babel-register`,
when I add this package, it workds well.

yelouafi:
I think it's needed because of this line in package.json which requires `babel-register` from the npm command line

```js
"scripts": {
   ...
   "real-world": "npm --prefix examples/real-world install examples/real-world && cross-env BABEL_ENV=cjs node --require babel-register examples/real-world/server.js",
   ...
  },
```

It permits running the real-world example directly from the main repo (auto install deps && not having to `cd examples/real-world && npm start`). 

Initially I thought it was fine to provide users with scripts in the main package.json to directly start examples from the main repo folder. But I think now this is not a so good idea, we pollute the scripts section of the main package with too many demo related scripts (npm run counter, npm test counter).

We can accept this PR as a fix for the current solution. But I think we should migrate all examples to have their own package/build config using eg. create-react-app like it's done actually in the redux repo.

Andarist:
Sounds good to me

PR #639, opened 2016-11-17
Add throttle type definition

yenshih:
Base on [Issue 624](https://github.com/yelouafi/redux-saga/issues/624)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -76,46 +76,52 @@ export const buffers: {
   expanding<T>(limit?: number): Buffer<T>;
 };
 
+type HelperFunc0<A> = (action?: A) => any;
+type HelperFunc1<A, T1> = (arg1: T1, action?: A) => any;
+type HelperFunc2<A, T1, T2> = (arg1: T1, arg2: T2, action?: A) => any;
+type HelperFunc3<A, T1, T2, T3> = (arg1: T1, arg2: T2, arg3: T3,
+                                   action?: A) => any;
+type HelperFunc4<A, T1, T2, T3, T4> = (arg1: T1, arg2: T2, arg3: T3, arg4: T4,
+                                       action?: A) => any;
+type HelperFuncRest = (...args: any[]) => any;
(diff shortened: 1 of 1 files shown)


aikoven:
Please see https://github.com/yelouafi/redux-saga/issues/619, let's make it a little more strict

yenshih:
I've seen that issue and create a new commit.
Hope it will be better.

index.d.ts
```diff
@@ -76,28 +76,57 @@ export const buffers: {
...
+                  pattern: Pattern<A>,
+                  worker: HelperFunc3<A, T1, T2, T3>,
+                  arg1: T1, arg2: T2, arg3: T3): SagaIterator;
+  <A, T1, T2, T3, T4>(ms: number,
+                      pattern: Pattern<A>,
+                      worker: HelperFunc4<A, T1, T2, T3, T4>,
+                      arg1: T1, arg2: T2, arg3: T3, arg4: T4): SagaIterator;
+  <A>(ms: number,
+      pattern: Pattern<A>,
+      worker: HelperFuncRest, ...args: any[]): SagaIterator;
```
aikoven: That's not what I meant. With this version you can do this:
```ts
throttle(100, 'some_type',  (arg: number) => {}, 'not-a-number')
```
So even if there are less than four arguments, it still won't type-check correctly because it matches the last overload.
The correct way would be to add `...rest` overload with four five parameters:
```ts
  <A, T1, T2, T3, T4, T5>(ms: number,
                   pattern: Pattern<A>,
                   worker: (arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5, ..rest: any[]) => any,
                   arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5, ...rest: any[]): SagaIterator;
```

aikoven:
Looks good now 👍

yenshih:
Thanks for your help and patience!

PR #644, opened 2016-11-20
Saga Error Handling Example

granmoe:
Use spawn to start sagas so that an uncaught exception doesn't terminate all of them. Restart a saga on an asynchronous exception. Terminate it on a sync exception.

If you simply `yield call` your sagas within your root saga, then any uncaught exception in any of them will kill all of them, making your app unresponsive. To isolate each saga from the others, I use `spawn`. This way, an error in one doesn't affect any of the others, and they remain responsive.

However, we still don't want the saga that had an uncaught exception to die because maybe some crazy async error was thrown that we somehow didn't add specific handling for. To deal with this, we can add a `while (true) { try { yield call(saga) } catch (e) { } }` that will restart the saga after handling its error. But if the error is simply something that is thrown when the saga starts (we make a typo during development, for example, and try to access a property from undefined), we don't want to start this saga, causing it to immediately throw the error, catch the error, and repeat forever. To check if the error is sync, I added the `isSyncError` flag that only gets set to `false` in a deferred manner, after the current stack is cleared, via `setTimeout`.

I've wrestled with this problem quite a bit, and I can't come up with any better solution. I would be happy to see some alternatives, but this is a significant pain point that just isn't addressed in any of the documentation or examples. Therefore, I think we need to offer an idiomatic way of solving it in this example.

This is a resurrection of #610, which I had closed in order to get a different pull request in in the meantime (can't have two PRs at once).

### DIFF
--- a/examples/async/webpack.config.js
+++ b/examples/async/webpack.config.js
@@ -2,7 +2,7 @@ var path = require('path')
 var webpack = require('webpack')
 
 module.exports = {
-  devtool: 'cheap-module-eval-source-map',
+  devtool: 'inline-source-map',
   entry: [
     'webpack-hot-middleware/client?reload=true',
     path.join(__dirname, 'src', 'main')
--- a/examples/async/src/actions/index.js
+++ b/examples/async/src/actions/index.js
@@ -1,8 +1,9 @@
-
 export const REQUEST_POSTS = 'REQUEST_POSTS'
 export const RECEIVE_POSTS = 'RECEIVE_POSTS'
 export const SELECT_REDDIT = 'SELECT_REDDIT'
 export const INVALIDATE_REDDIT = 'INVALIDATE_REDDIT'
+export const SET_ERROR = 'SET_ERROR'
+export const CLEAR_ERROR = 'CLEAR_ERROR'
 
 export function selectReddit(reddit) {
   return {
@@ -31,5 +32,18 @@ export function receivePosts(reddit, posts) {
     reddit,
     posts,
     receivedAt: Date.now()
+  }
+}
+
+export function setError(errorMessage) {
+  return {
+    type: SET_ERROR,
+    errorMessage
+  }
+}
+
+export function clearError() {
+  return {
+    type: CLEAR_ERROR
   }
 }
--- a/examples/async/src/containers/App.js
+++ b/examples/async/src/containers/App.js
@@ -22,12 +22,21 @@ class App extends Component {
   }
 
   render() {
-    const { selectedReddit, posts, isFetching, lastUpdated } = this.props
+    const { selectedReddit, posts, isFetching, lastUpdated, errorMessage } = this.props
     return (
       <div>
+        { errorMessage &&
+          <p>
+            Something bad happened:
+            <br/>
+            {errorMessage}
+            <br/>
+            Please try again
+          </p>
+        }
         <Picker value={selectedReddit}
                 onChange={this.handleChange}
-                options={[ 'reactjs', 'frontend' ]} />
+                options={[ 'reactjs', 'frontend', 'simulate network failure' ]} />
         <p>
           {lastUpdated &&
             <span>
@@ -67,7 +76,7 @@ App.propTypes = {
 }
 
 function mapStateToProps(state) {
-  const { selectedReddit, postsByReddit } = state
+  const { selectedReddit, postsByReddit, errorMessage } = state
   const {
     isFetching,
     lastUpdated,
@@ -81,7 +90,8 @@ function mapStateToProps(state) {
     selectedReddit,
     posts,
     isFetching,
-    lastUpdated
+    lastUpdated,
+    errorMessage
   }
 }
 
--- a/examples/async/src/reducers/index.js
+++ b/examples/async/src/reducers/index.js
@@ -2,7 +2,9 @@ import { combineReducers } from 'redux'
 import {
   SELECT_REDDIT,
   REQUEST_POSTS,
-  RECEIVE_POSTS
+  RECEIVE_POSTS,
+  SET_ERROR,
+  CLEAR_ERROR
 } from '../actions'
 
 function selectedReddit(state = 'reactjs', action) {
@@ -35,7 +37,6 @@ function posts(state = {
 
 function postsByReddit(state = { }, action) {
   switch (action.type) {
-    case REQUEST_POSTS:
     case RECEIVE_POSTS:
       return { ...state,
         [action.reddit]: posts(state[action.reddit], action)
@@ -45,9 +46,21 @@ function postsByReddit(state = { }, action) {
   }
 }
 
+function errorMessage(state = null, action) {
+  switch (action.type) {
+    case SET_ERROR:
+      return action.errorMessage
+    case CLEAR_ERROR:
+      return null
+    default:
+      return state
+  }
+}
+
 const rootReducer = combineReducers({
   postsByReddit,
-  selectedReddit
+  selectedReddit,
+  errorMessage
 })
 
 export default rootReducer
--- a/examples/async/src/sagas/index.js
+++ b/examples/async/src/sagas/index.js
@@ -1,15 +1,17 @@
 /* eslint-disable no-constant-condition */
-
-
-import { take, put, call, fork, select } from '../../../../src/effects'
+import { take, put, call, fork, spawn, select } from '../../../../src/effects'
 import fetch from 'isomorphic-fetch'
 import * as actions from '../actions'
 import { selectedRedditSelector, postsByRedditSelector } from '../reducers/selectors'
 
 export function fetchPostsApi(reddit) {
+  if (reddit === 'simulate network failure') {
+    throw new Error('Failed to fetch')
+  } else {
     return fetch(`http://www.reddit.com/r/${reddit}.json` )
-            .then(response => response.json() )
-            .then(json => json.data.children.map(child => child.data) )
+      .then(response => response.json() )
+      .then(json => json.data.children.map(child => child.data) )
+  }
 }
 
 export function* fetchPosts(reddit) {
@@ -21,6 +23,7 @@ export function* fetchPosts(reddit) {
 export function* invalidateReddit() {
   while (true) {
     const {reddit} = yield take(actions.INVALIDATE_REDDIT)
+    yield put(actions.clearError())
     yield call( fetchPosts, reddit )
   }
 }
@@ -29,11 +32,13 @@ export function* nextRedditChange() {
   while(true) {
     const prevReddit = yield select(selectedRedditSelector)
     yield take(actions.SELECT_REDDIT)
+    yield put(actions.clearError())
 
     const newReddit = yield select(selectedRedditSelector)
     const postsByReddit = yield select(postsByRedditSelector)
-    if(prevReddit !== newReddit && !postsByReddit[newReddit])
-      yield fork(fetchPosts, newReddit)
+    if(prevReddit !== newReddit && !postsByReddit[newReddit]) {
+      yield call(fetchPosts, newReddit)
+    }
   }
 }
 
@@ -44,6 +49,23 @@ export function* startup() {
 
 export default function* root() {
   yield fork(startup)
-  yield fork(nextRedditChange)
-  yield fork(invalidateReddit)
+  const sagas = [nextRedditChange, invalidateReddit]
+
+  yield sagas.map(saga =>
+    spawn(function* () {
+      let isSyncError = false
+      while (!isSyncError) {
+        isSyncError = true
+        try {
+          setTimeout(() => isSyncError = false)
+          yield call(saga)
+        } catch (e) {
+          if (isSyncError) {
+            throw new Error(saga.name + ' was terminated because it threw an exception on startup.')
+          }
+          yield put(actions.setError(e.message))
+        }
+      }
+    })
+  )
 }


granmoe:
Here's a wild and probably bad idea: should the middleware itself provide some mechanism for terminating a saga that immediately, synchronously throws an exception? This is probably too heavy-handed and would make the library too opinionated. Just spitballing here.

granmoe:
Doesn't anyone have any opinions or advice on this?

examples/async/webpack.config.js
```diff
@@ -2,7 +2,7 @@ var path = require('path')
 var webpack = require('webpack')
 
 module.exports = {
-  devtool: 'cheap-module-eval-source-map',
+  devtool: 'inline-source-map',
```
granmoe: I've had better luck with this source map setting

granmoe:
I'm really struggling to get any response from the community for this problem. I've tried reddit, gitter, here, reactiflux. Can anyone help?

aikoven:
Nice observation with sync errors on startup!

I typically use a higher-order saga, like this:
```js
function autoRestart(generator) {
  return function* autoRestarting(...args) {
    while (true) {
      try {
        yield call(generator, ...args);
      } catch (e) {
        console.log(`Unhandled error in '${generator.name}'`, e);
      }
    }
  }
}

const mySaga = autoRestart(function* mySaga () {
  // ...
});
```

Another option is to make a custom effect creator:
```js
function forkAutoRestarting(fn, ...args) {
  return fork(function* () {
    while (true) {
      try {
        yield call(fn, ...args);
      } catch (e) {
        console.log(`Unhandled error in '${generator.name}'`, e);
      }
    }
  });
} 

function* rootSaga() {
  yield forkAutoRestarting(childSaga);
}
```
However, I like the first one more, because it's probably saga's own concern whether it supports restarting, not its parent's.

Andarist:
I really like @aikoven approach. It makes sense to have steer things like this.

Unhandled exceptions is not a problem of saga itself, but more of a general coding problem.

@granmoe If your solution works for you, it should be fine for you to use it I think. or do you have some issues with it?

Also see a commentary here - https://github.com/yelouafi/redux-saga/issues/632#issuecomment-262352406 about `onError`. Maybe that feature could be utilised for more general use cases as some kind of global-saga try/catch? WDYT? How would you see it working in various scenarious? Differences between fork/spawn?

granmoe:
Thanks for the feedback!

The reason I want to wrap the root saga itself is because 99% of the time, I don't want any of my sagas to die permanently if an unhandled exception occurs. This is the one important difference between mine and @aikoven 's approach.

@Andarist No, I don't have any problems with my current solution. Yes, a global try/catch (with the option to restart sagas after non-sync errors) would be perfect. 

I know it's easy to accomplish with about 20 lines of code (like @aikoven and I have above, for example), but many co-workers have voiced concern about this chunk of code because it feels like too low-level of an abstraction, like something redux-saga should handle for you. Sure, this amounts to just hiding that little chunk of code in the library, but at least then we can have one standard way of doing it that everyone (and apparently a lot of people have need for this) can use.

Andarist:
> The reason I want to wrap the root saga itself is because 99% of the time, I don't want any of my sagas to die permanently if an unhandled exception occurs. This is the one important difference between mine and @aikoven 's approach.

Well, with @aikoven 's solution the main outcome is pretty much the same - those sagas wont die permanently. Or didnt I grasp something?

> @Andarist No, I don't have any problems with my current solution. Yes, a global try/catch (with the option to restart sagas after non-sync errors) would be perfect.

Could you post example API with explanation what providing onError and what its absence would cause?

> I know it's easy to accomplish with about 20 lines of code (like @aikoven and I have above, for example), but many co-workers have voiced concern about this chunk of code because it feels like too low-level of an abstraction, like something redux-saga should handle for you. Sure, this amounts to just hiding that little chunk of code in the library, but at least then we can have one standard way of doing it that everyone (and apparently a lot of people have need for this) can use.

Well, I feel that although many people would like it to see in `redux-saga` it would break other's assumptions (even if they are minority) and we should then provide a way of reaching present behaviour. So there is no win in this.

However various strategies for error handling could be included in the docs as examples with explanations for everybody to see and chose between them. With the most basic one being using `spawn` in the rootSaga and `fork` in every `spawn`ed one.

@aikoven 
Would you agree on using your code in the docs or maybe even making one of them a built-in helper? I could provide implementation of this using `fsmIterator` internally (in the way other helpers like `takeEvery` etc are implemented)

granmoe:
@Andarist Sorry, the actual difference with @aikoven 's solution is that I take into account synchronous errors and I DO let the saga die in that case. Disregard my previous comment.

> Well, I feel that although many people would like it to see in redux-saga it would break other's assumptions (even if they are minority) and we should then provide a way of reaching present behaviour. So there is no win in this.

> However various strategies for error handling could be included in the docs as examples with explanations for everybody to see and chose between them. With the most basic one being using spawn in the rootSaga and fork in every spawned one.

Good points. I think providing examples in the docs is a good compromise. Although, you mention creating a built-in helper using @aikoven 's solution, and I would prefer that. I would argue that the helper should use my check for synchronous errors, though, because otherwise it could continually restart a saga that throws on startup every time.

> Could you post example API with explanation what providing onError and what its absence would cause?

Example API would be a helper that abstracts this code:

```javascript
export default function* root() {
  yield fork(startup)
  const sagas = [nextRedditChange, invalidateReddit]

  yield sagas.map
      let isSyncError = false
      while (!isSyncError) {
        isSyncError = true
        try {
          setTimeout(() => isSyncError = false)
          yield call(saga)
        } catch (e) {
          if (isSyncError) {
            throw new Error(saga.name + ' was terminated because it threw an exception on startup.')
          }
          yield put(actions.setError(e.message))
        }
      }
})
```

So then in my original example in this bug, I would do:

```javascript
export default function* root() {
  yield fork(startup)
  const sagas = [nextRedditChange, invalidateReddit]

  yield sagas.map(restartOnAsyncError(customErrorHandler))
```

Or if you want to restart the saga(s) regardless of whether the error is synchronous or not, there could be another helper available:


```javascript
export default function* root() {
  yield fork(startup)
  const sagas = [nextRedditChange, invalidateReddit]

  yield sagas.map(restartOnError(customErrorHandler))
```

(customErrorHandler would be defined by the user, of course)

This would get my vote :-) And we could update the examples and docs accordingly. What do you think?

granmoe:
I'd also be willing to help out in getting the examples and docs up to date, and anything else I can help with.

aikoven:
@Andarist 
> Would you agree on using your code in the docs or maybe even making one of them a built-in helper?

Absolutely! @granmoe's sync error detection should be included as well. I think we should fail loudly in that case.

@granmoe 
> Sorry, the actual difference with @aikoven 's solution is that I take into account synchronous errors and I DO let the saga die in that case.

The purpose of these snippets was to show generic ways to handle restarts, without needing to keep that logic inside real saga code, by using higher order saga or custom effect creator. Surely they're now incomplete without your sync error handling trick :)

Andarist:
> @Andarist 
> > Would you agree on using your code in the docs or maybe even making one of them a built-in helper?

> Absolutely! @granmoe's sync error detection should be included as well. I think we should fail loudly in that case.

There are only 2 things I worry with this trick.
1. its a trick :P could be non-obvious in the source after a while, however I guess its the best way of detecting this
2. it can happen not only on a startup, and I feel like it should be applied on each sync step of the generator, which is rather a complex thing to do (not overly complex, but still)

Also the whole error handling thing got me to thinking yesterday. I was searching for an interesting topic for my company's blog to write and post on and finally it seemed that it might be it ;) Wouldnt like to use your code (both of you! @aikoven @granmoe) for such purpose without your blessing (would ofc include props for both of you). And later I could convert it to the docs section (as blog post and docs should have a little bit different form).

Also the blog post is nice as it might raise awareness among people if we propagate it through twitter etc, lets be honest - everybody looks to the docs (beside API reference) just once, at the beginning while learning.

granmoe:
> it can happen not only on a startup, and I feel like it should be applied on each sync step of the generator, which is rather a complex thing to do (not overly complex, but still)

Oh...good point. I guess this would be needed, too, if there are situations in which you could have an endless loop of sync errors at some other step in the saga.

> Wouldnt like to use your code (both of you! @aikoven @granmoe) for such purpose without your blessing 

Sure, you can use my code :-) Can you comment with a link to the post here when it's done?

> everybody looks to the docs (beside API reference) just once, at the beginning while learning.

Yep 😏 And most of us only read the 10% we need to start writing code (sometimes overly complex code because we missed some relevant feature(s) of the API within the remaining 90%).

yelouafi:
@granmoe a drawback of using spawn is that it doesnt propgate cancellation down. not a prob. for top watchers but can be for watchers located in mid-hierarchy.

as I said to @Andarist spawn is the opposite extreme  to fork. The issue is essentially related to 2 params: error bubbling and cancellation propagation. Actually there is no api for more fine tunning of the 2 behaviors and thus possibly the core issue is here: need for a gradual top down tunning of fork errors starting with a global onError and more specific/per case config (we could also add config options to takeXXXX helpers).

for the auto retry behavior I assume we are referring here to watchers only. Because we can also fork non watchers and I dont think it makes sense to *always* restart a worker saga. in the latter case I guess this is what @aikoven meant when he said the restarting is a internal concern to the worker saga (possibly combined with a ui dialog)

Also the setTimeout in the example could potentially have some race condition issues. It's unlikely that an async request could return before the timeout but not impossible if I'm not mistaken (cached response, sevice worker).

regarding sync/async distinction I tend also to make the distinction but with diff. terms expected/unexpected (or business errors/bugs). for the 2nd type I agree that it doesnt make sense to catch them b/c the program is already in an unexpected state. but I'm not sure what other alternative could be used in place of simply halting all the things (possibly after executing some post mortem onError handler) because the program could be in an inconsistent state.

granmoe:
@yelouafi 

> @granmoe a drawback of using spawn is that it doesnt propgate cancellation down. not a prob. for top watchers but can be for watchers located in mid-hierarchy.

So if the error propogated up to the top, everything it bubbled through would get cancelled, right? But if it bubbled from a mid-layer to the top, anything BELOW would not get cancelled because cancellation would not propogate down from the top spawn? Is that correct? If so, maybe the best (albeit not perfect) approach is to rely ONLY on upward and not downward cancellation. Obviously, that is a big limitation, but I'm not sure of an alternative, especially given your last paragraph where you seem to have reached the same impasse.

> for the auto retry behavior I assume we are referring here to watchers only. Because we can also fork non watchers and I dont think it makes sense to always restart a worker saga. in the latter case I guess this is what @aikoven meant when he said the restarting is a internal concern to the worker saga (possibly combined with a ui dialog)

Yes, I was speaking only of watchers. Agreed, it doesn't make sense to always restart a worker (should be the worker's own responsibility to restart itself if it needs to for some reason).

> Also the setTimeout in the example could potentially have some race condition issues. It's unlikely that an async request could return before the timeout but not impossible if I'm not mistaken (cached response, sevice worker).

Hmmm...that would suck. I guess the penalty is that sometimes a watcher that you don't want to ever die would die in this narrow range of scenarios.

granmoe:
One more update to how this hypothetical global task restarter should work. A co-worker noticed today that his "once only" saga was running multiple times somehow. The reason is due to a kind of funny oversight in the code. 

```javascript
export default function* root() {
  yield fork(startup)
  const sagas = [nextRedditChange, invalidateReddit]

  yield sagas.map
      let isSyncError = false
      while (!isSyncError) {
        isSyncError = true
        try {
          setTimeout(() => isSyncError = false)
          yield call(saga)
          // After success, we loop again, too!
        } catch (e) {
          if (isSyncError) {
            throw new Error(saga.name + ' was terminated because it threw an exception on startup.')
          }
          yield put(actions.setError(e.message))
        }
      }
})
```

So it should be updated to have a `break` after the saga finishes, like so:

```javascript
export default function* root() {
  yield fork(startup)
  const sagas = [nextRedditChange, invalidateReddit]

  yield sagas.map
      let isSyncError = false
      while (!isSyncError) {
        isSyncError = true
        try {
          setTimeout(() => isSyncError = false)
          yield call(saga)
          break
        } catch (e) {
          if (isSyncError) {
            throw new Error(saga.name + ' was terminated because it threw an exception on startup.')
          }
          yield put(actions.setError(e.message))
        }
      }
})
```

@Andarist

granmoe:
Just commenting to revive this thread and see if anyone has had any further thoughts on it.

Andarist:
Sorry for the lack of response here. Please do not close this PR as I think it is a valuable piece of discussion which is worth to be kept and I'll try to get to summarizing it and adding it to the docs in some form eventually.

Personally after many thoughts on the topic I dont want to recommend your pattern (restarting) for everyone - that doesnt mean its not a valid approach if somebody needs it. Need to rethink what exactly the recommended approach is in my opinion - but it's probably something more like tests + isolated sagas (through `spawn`? maybe there is a need for some better isolation solutions, aint sure yet) and just letting sagas to fail + ofc logging the errors to the analytics, so the bugs can be fixed.

I believe React 16 is moving into the right direction with error handling - https://facebook.github.io/react/blog/2017/07/26/error-handling-in-react-16.html . Let me cite just few things:
- "In the past, JavaScript errors inside components used to corrupt React’s internal state and cause it to emit cryptic errors on next renders. These errors were always caused by an earlier error in the application code, but React did not provide a way to handle them gracefully in components, and could not recover from them."
- "This change has an important implication. As of React 16, errors that were not caught by any error boundary will result in **unmounting of the whole React component tree**."
- "We debated this decision, but in our experience it is worse to leave corrupted UI in place than to completely remove it."

I think it closely relates to the sagas problem. While React 16 introduces "error boundaries" I think its not so easy to apply this concept to the saga due to differences in the formed tree structure (component tree is different than a saga tree) - it behaves in different way. However it might be worthwhile to think about it some more and maybe introduce something similar - no ideas in this field on my side at the moment though.

granmoe:
> "This change has an important implication. As of React 16, errors that were not caught by any error boundary will result in unmounting of the whole React component tree."

Yes, this does seem like a very similar problem. React offers a way to protect your code from unhandled exceptions, so why shouldn't we? I agree, though, implementing this for sagas seems very difficult.

ajwhite:
> Personally after many thoughts on the topic I dont want to recommend your pattern (restarting) for everyone

I agree with this sentiment. A saga might die in a state that it needs to restore back into. For example:

```js
function *sagaThatMayCrash () {
  // wait for something that happens _during app startup_
  yield take(APP_INITIALIZED)

  // assume it dies here
  yield call(doSomethingThatMayCrash)
}
```

If the `sagaThatMayCrash` is restarted, it will restart and wait for an action that only happens once when the application starts up. In this scenario, it restarts, but it never _recovers_.

I've been thinking about this scenario too and have considered providing some meta to the saga. Think:

```js
function *sagaThatMayDie (supervisorState = {firstRun: true, restartCount: 0, ...}) {
  // here we can see if we have restarted or not
  if (supervisorState.firstRun) {
    yield take(INITIALIZE)
  }

  yield call(doSomethingThatMayCrash)
}
```

The caveat here is that this creates a coupling between the actor and the supervisor. Still brainstorming...

Andarist:
Exactly! Unfortunately I do not have any silver bullet for those kinds of problem. I also believe that more fine-grained control should be preferred when it comes to auto-restarting your processes - explicitness over implicitness.

Fortunately saga is really flexible and allows you to implement your own abstractions over the system or create abstraction-library to share with others. Best the redux-saga core can do is offer ([like it does]([addons section](https://redux-saga.js.org/docs/ExternalResources.html))) the addons section for you to share your doings with the community.

Andarist:
**NOTE** If somebody want to propose API and behaviour spec for something like error boundaries in React - I would gladly read it and work out any rough edges. It is not my priority at the moment though, I sincerely invite anyone who would like to collaborate on the project.

ajwhite:
@Andarist we've created a few "supervisor" strategies, but we also haven't come up with a great abstraction. It's a particularly challenging one. I'd be quite interested in looking at this problem with other folks that have also put some thought into it.

We are currently have a few compositional wrappers, like so:

```js
// Keep the fork alive if it crashes. Note - this just simply forks it again.
yield keepAlive(fork(saga))

// Calls the saga and swallows any errors
yield call(safely(saga))

// Tries a few times, then finally fails
yield withRetry(RetryStrategy.attempts(5), RetryStrategy.exponential())(saga)

// Compose rules together, safely call a saga, report any errors, and retry if it fails 
const wrap = compose(
  safely,
  withErrorReporting,
  withRetry(RetryStrategy.attempts(2))
);

// Safely wraps the saga and reports errors
yield takeEvery(action, wrap(saga))
```

I think these are good _ingredients_ to a larger abstraction.

Andarist:
> I think these are good ingredients to a larger abstraction.

Probably, but it seems **really** cool to me already ❤️

granmoe:
@ajwhite The API you have up there (the supervisor strategies) is awesome :D

granmoe:
Yeah there’s a reactiflux channel, but I think this is a significant enough topic to deserve a dedicated channel. Should we create a slack channel and leave the link here, so anyone interested can join? Wdyt?

Andarist:
Fine by me - there is also a gitter channel 😄 In general I would like to have 1 place to hang out, thought it might be reactiflux, but their push notifications cant handle this big server :s so ive kinda stayed on gitter and unfortunately im overlooking discord's one

granmoe:
Gitter works for me :-) I'll open a tab for it now and leave it open perpetually :P

psdao-lw:
I see that this got dropped off last year. Has any progress on this subject been made?

Issue #647, opened 2016-11-22
Undo recipe


HansDP:
Hello

I was looking at the Undo recipe shown on the documentation website of redux-saga. 

I love the idea, it is plain and simple. However, I started wondering whether there are issues with the provided code.

This is what got me thinking (so, please correct me if I'm wrong, which would not surprise me, since I'm not an expert, yet :)

When the main saga starts, a `takeEvery` (which is non-blocking) will listen for all `ARCHIVE_THREAD` actions and immediately continue to the next `take(UNDO)` action, which is blocking. Now, when the user performs an ’archive thread’, an `onArchive` is forked. When the user performs an `UNDO`, the `take(UNDO)` gets unblocked, the next statement will cancel all forks created by the `takeEvery` task and the `while(true)` gets restarted. All is good. 

The same applies when the user does not `UNDO` the archive (this will just keep the `takeEvery` waiting for the next archive action). 

But, what happens when the user triggers two ’archive threads’ in rapid succession? Probably two ’undo UI elements’ will be shown. But when the user now does an `UNDO` on any of the ’undo UI elements’, both archive requests will be undone. Or, maybe, the UI is smart enough to combine both UI elements into one (this depends on the requirements of the application). Then it would make more sense. 

Another issue I see with this solution is when the user triggers an `UNDO` at the exact time where the code is calling the `updateThreadApi` method. This method is probably a Fetch request (or similar). 
So, the HTTP request has been started, and now the `archiveTask` gets cancelled, forcing the `onArchive` to execute the `finally` block and thus marking the thread as not archived. And this results in an invalid state (the thread is archived on the server, but the client thinks it is not archived)
 
All of this made me wonder why this Undo recipe was not provided using a race condition? In my personal opinion, using a race would make the code much more understandable. So, I tried to make a solution (small disclaimer: I did not test this code, I've written it in Notes on my phone :-s

``` javascript 

function* onArchive(action) {
  const { threadId } = action
  const undoId = `UNDO_ARCHIVE_${threadId}`
  
  const thread = { id: threadId,archived: true }
  // show undo UI element, and provide a key to communicate
  yield put(actions.showUndo(undoId))
  // optimistically mark the thread as `archived`
  yield put(actions.updateThread(thread))
  // allow user time to active the undo action
  const { undo, timedout } = yield race({
    undo: take(action => action.type === UNDO && action.undoId === undoId),
    timedout: call(delay, 5000)
  })
      
  // hide undo UI element, in any case
[...]

PR #650, opened 2016-11-22
Update Undo recipe

HansDP:
Modified the Undo recipe to use a race condition
Related issue: #647

### DIFF
--- a/docs/recipes/README.md
+++ b/docs/recipes/README.md
@@ -154,43 +154,54 @@ creates a higher order reducer to do most of the heavy lifting for the developer
 
 However, this method comes with it overheard from storing references to the previous state(s) of the application.
 
-Using redux-saga's `delay` and `cancel` we can implement a simple, one-time undo without enhancing
+Using redux-saga's `delay` and `race` we can implement a simple, one-time undo without enhancing
 our reducer or storing the previous state.
 
 ```javascript
-import {  take, put, call, fork, cancel, cancelled } from 'redux-saga/effects'
-import { takeEvery, delay } from 'redux-saga'
+import { take, put, call, spawn, race } from 'redux-saga/effects'
+import { delay } from 'redux-saga'
 import { updateThreadApi, actions } from 'somewhere'
 
-function* onArchive() {
-  try {
-      const thread = { id: 1337, archived: true }
-      // show undo UI element
-      yield put(actions.showUndo())
-      // optimistically mark the thread as `archived`
-      yield put(actions.updateThread(thread))
-      // allow user time to active the undo action
-      yield call(delay, 5000)
-      // hide undo UI element
-      yield put(actions.hideUndo())
-      // make the API call to apply the changes remotely
-      yield call(updateThreadApi, thread)
-  } finally {
-    if (yield cancelled()) {
-      // revert thread to previous state
-      yield put(actions.updateThread({ id: 1337, archived: false }))
-    }
+function* onArchive(action) {
+
+  const { threadId } = action
+  const undoId = `UNDO_ARCHIVE_${threadId}`
+
+  const thread = { id: threadId, archived: true }
+
+  // show undo UI element, and provide a key to communicate
+  yield put(actions.showUndo(undoId))
+
+  // optimistically mark the thread as `archived`
+  yield put(actions.updateThread(thread))
+
+  // allow the user 5 seconds to perform undo.
+  // after 5 seconds, 'archive' will be the winner of the race-condition
+  const { undo, archive } = yield race({
+    undo: take(action => action.type === 'UNDO' && action.undoId === undoId),
+    archive: call(delay, 5000)
+  })
+
+  // hide undo UI element, the race condition has an answer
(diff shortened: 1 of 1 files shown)


HansDP:
I did some basic smoke testing on the code snippet. It runs and everything seems fine.

PS: this is the first time I created a pull request on github. If I did anything wrong, just let me know

docs/recipes/README.md
```diff
@@ -154,43 +154,54 @@ creates a higher order reducer to do most of the heavy lifting for the developer
...
 function* main() {
   while (true) {
-    // listen for every `ARCHIVE_THREAD` action in a non-blocking manner.
-    const onArchiveTask = yield takeEvery(ARCHIVE_THREAD, onArchive)
-    // wait for the user to activate the undo action;
-    yield take(UNDO)
-    // and then cancel the fetch task
-    yield cancel(onArchiveTask)
+    // wait for an ARCHIVE_THREAD to happen
+    const action = yield take(`ARCHIVE_THREAD`)
```
Andarist: I wouldnt use `` syntax here, as you are not interpolating anything. '' would suffice

docs/recipes/README.md
```diff
@@ -154,43 +154,54 @@ creates a higher order reducer to do most of the heavy lifting for the developer
...
+    // use spawn to execute onArchive in a non-blocking fashion (which also
```
Andarist: you have the opening bracket here in the comment, but no closing one

Andarist:
It looks rly great, I just aint sure if introducing `spawn` here is necessary. It convolutes the example a little bit. Anyway - `fork` vs `spawn` struggle is geting bigger and bigger in the community. Gonna speak about it with @yelouafi 

I've put 2 really minor comments, if you decide to correct them, please do an ammend commit (or squash them) + rebase with the current master - it will help keeping the history clean.

HansDP:
I agree about the `spawn` vs `fork` comment.
But I'm not sure how to handle cancellation in `onArchive` when the `main` saga gets cancelled.

Just 'ignore' this (to keep the example clear) is an option, but in production code, i guess there must be some flags to know whether or not the undo ui should be hidden and whether or not the optimistic update should be reverted.

Something like

```javascript
import { take, put, call, fork, race } from 'redux-saga/effects'
import { updateThreadApi, actions } from 'somewhere'

function* onArchive(action) {

  let needsHideUndo = false
  let needsRevertOptimisticUpdate = false

  const { threadId } = action
  const undoId = `UNDO_ARCHIVE_${threadId}`

  const thread = { id: threadId, archived: true }

  try {

    // show undo UI element, and provide a key to communicate
    yield put(actions.showUndo(undoId))
    needsHideUndo = true

    // optimistically mark the thread as `archived`
    yield put(actions.updateThread(thread))
    needsRevertOptimisticUpdate = true

    // allow the user 5 seconds to perform undo.
    // after 5 seconds, 'archive' will be the winner of the race-condition
    const { undo, archive } = yield race({
      undo: take(action => action.type === 'UNDO' && action.undoId === undoId),
      archive: call(delay, 5000)
    })

    // hide undo UI element, the race condition has an answer
    needsHideUndo = false
    yield put(actions.hideUndo(undoId))

    if (archive) {
      // make the API call to apply the changes remotely
      needsRevertOptimisticUpdate = false
      yield call(updateThreadApi, thread)
    }

  } finally {
    // no need to check for cancellation, just execute this code when the 
    // try-block has finished (it will include cancellation as well. but also
    // the actual undo, error handling, ...)

    if (needsHideUndo) {
      yield put(actions.hideUndo(undoId))
    }

    if (needsRevertOptimisticUpdate) {
      // revert thread to previous state
      yield put(actions.updateThread({ id: threadId, archived: false }))
    }
  }
}

function* main() {
  while (true) {
    // wait for an ARCHIVE_THREAD to happen
    const action = yield take(`ARCHIVE_THREAD`)
    // use spawn to execute onArchive in a non-blocking fashion (which also
    // prevents cancelation when main saga gets cancelled.
    // This helps us in keeping state in sync between server and client
    yield fork(onArchive, action)
  }
}
```

But I'm not sure this is a proper solution. And it is harder to understand.

Anyway, I will update the code sample with your comments

HansDP:
Did I updated this pull request correctly? (because i see 3 commits instead of one)
If so, what should I do? I'm not used to using github (sorry for the trouble)

Andarist:
Only a single commit should be wanting to get merged.

You should:
`git rebase -i HEAD~3`
you will see 3 commits - use `p d s` on them in that order, so 1st will get picked, 2nd dropped, 3rd squashed
you will get asked about new commit message - just keep single `Update Undo recipe`

I aint sure if you did it right, so you might skip next steps if you have already done them:
`git remote add upstream https://github.com/yelouafi/redux-saga.git`
`git fetch upstream`

And finally
`git rebase upstream/master`
`git force --push`

Andarist:
Thanks, real good example, leveraging `take(predicate)` which is rly neat and underpromoted in the docs 🎉

PR #652, opened 2016-11-23
Renamed put.sync to put.resolve (better naming). Keeping put.sync for…

Andarist:
… a while with deprecation warning

As of - https://github.com/yelouafi/redux-saga/issues/632#issuecomment-262471473

Docs should be updated after a release.


yelouafi:
looks better to me. Thanks

PR #653, opened 2016-11-23
Fixed comment in Glossary which indicated wrongfully that put is a bl…

### LINKED ISSUE #626: Contradiction in docs whether `put` is blocking
In https://github.com/yelouafi/redux-saga/blame/c6af0703ba5e8829494583e33c1537e137569ff1/docs/Glossary.md#L41 it says that `put` is blocking, but in https://github.com/yelouafi/redux-saga/blame/c6af0703ba5e8829494583e33c1537e137569ff1/docs/api/README.md#L316 it says that it is non-blocking (and below that only `put.sync` is non-blocking). 

Which of those two is true? The other one should be corrected.

Andarist:
…ocking effect (fixes #626)


yelouafi:
thank :)

PR #659, opened 2016-11-29
Docs cleanup: Removes reference to unused import 

misscs:
Small docs eslint-ish fix. The example code was importing `fork` but the code isn't referencing it.


yelouafi:
thanks

PR #666, opened 2016-11-30
API Reference: remove gendered pronoun

samuelcole:
I'm pretty sure middleware doesn't have a gender...


yelouafi:
of course :) thanks.

PR #673, opened 2016-12-06
Allow to provide custom emitter to saga middleware

pbadenski:
This feature is something I found useful when integrating with [redux-batched-actions](https://github.com/tshelburne/redux-batched-actions). Redux-batched-actions allows to publish and reduce all actions in a batch (which is a separate action), rather than going one by one - useful when you send hundreds or thousands of them. At the same time when used with redux-saga it introduces a bit of a leaky abstraction - as TAKE is concerned what gets published is a completely new action. Actual actions are inside of it's payload, and that's what redux-saga needs to be informed with.

See example below:

```javascript
const emitter_ = emitter();
  const sagaMiddleware = createSagaMiddleware({emitter: {
    subscribe: emitter_.subscribe,
    emit: (action: any) => {
      if (action.type === "BATCHING_REDUCER.BATCH") {
        action.payload.map(emitter_.emit);
      } else {
        emitter_.emit(action);
      }
    }
  }});
```

### DIFF
--- a/test/middleware.js
+++ b/test/middleware.js
@@ -74,3 +74,20 @@ test('middleware.run', assert => {
 
   assert.end()
 })
+
+test('middleware\'s custom emitter', assert => {
+  const emittedActions = [];
+  const action = {};
+  sagaMiddleware({
+    emitter: function (emit) {
+      return function (action) {
+        emittedActions.push(action);
+      }
+    }
+  })({})(action => action)(action);
+
+  assert.deepEqual(emittedActions, [action],
+    'custom emitter needs be executed when specified');
+
+  assert.end();
+});


Andarist:
Could you provide docs update in your PR? Would be much appreciated.

pbadenski:
@Andarist sure I'll take a shot
@yelouafi sorry I don't understand, is there anything you want me to change?

yelouafi:
@pbadenski oops! sorry I was reviewing the ` index.d.ts` file without noticing.  Just ignore my previous comment.

pbadenski:
@Andarist I added information in API docs, let me know what you think

Andarist:
Im on the vacation atm, gonna look into it in a few days.

Andarist:
I just thought that this approach - https://gist.github.com/slorber/c095671a14b2eccf8ad665a97a7b8ad1 could be leveraged to achieve your goal, although it would need much more boilerplate to connect the pieces.

Also I just thought that as `runSaga`'s store-like options dispatch/getState/subscribe are disconnected from the standard middleware emitting from them via subscribe could cause every scheduling problem which was fixed by 'asap' scheduler in the standard middleware recently. That should be addressed I think, WDYT @yelouafi ?

also much simpler way of adding the requested functionality than having to provide whole emitter could be ability to 'hijack' emit with the option, smth like this
```javascript
  function sagaMiddleware({getState, dispatch}) {
    runSagaDynamically = runSaga
    const sagaEmitter = emitter()
    if (options.emit) {
      const realEmit = sagaEmitter.emit
      sagaEmitter.emit = (action) => options.emit(realEmit, action)
    }
```
and then ofc passed in `options.emit` could handle the action however it wants to and still emit easily
@pbadenski WDYT?

pbadenski:
I like narrowing it down to only providing "emit" option instead of replacing the whole "emitter". The other solution did seem a bit clunky. Should I wait for @yelouafi's response or just go ahead and change it?

Andarist:
W8. He'll probably have valuable input :)

yelouafi:
I think using runSaga wont be much convenient. it's mainly intended to connect to non redux envs (eg other flux libs, angular). beside you'd have to write another middleware/enhancer to hookup redux actions.

I'd favour the 2nd option (only provide the emit). maybe we should find a more apprpriate name since the option can have other uses (like filtering out some actions).

@pbadenski for the motivating use case of batched actions. If I'm not mistaken you have to solve the same issue for every other middleware: logger, devtool ...?

Andarist:
@pbadenski 
will you have time in following days to work on the PR? I would like to release a new `redux-saga` version soon and include this PR in it if possible :)

pbadenski:
@yelouafi yeah, using batched actions seem to require small tweaks in other middleware. We just barely started using it so still learning about them.

@Andarist I should have enough time for these changes :) I'll try pushing it today.

pbadenski:
@Andarist I decided to take a slightly different approach, let me know what you think. I started off with a solution allowing to override only "emit" function, but all my attempts felt clunky. I think this new solution better embraces @yelouafi comment on finding "a more appropriate name since the option can have other uses".

*) emit requires access to private field "subscribers", if "emit" override is provided it needs to accommodate that

src/internal/middleware.js
```diff
@@ -44,9 +44,14 @@ export default function sagaMiddlewareFactory(options = {}) {
...
   }
 
+  if(options.processAction && !is.func(options.processAction)) {
+    throw new Error('`options.processAction` passed to the Saga middleware is not a function!')
+  }
+
   function sagaMiddleware({getState, dispatch}) {
     runSagaDynamically = runSaga
     const sagaEmitter = emitter()
+    const processAction = options.processAction || ((action) => action)
```
Andarist: `identity` is already implemented in `utils` so you can import it from there - https://github.com/yelouafi/redux-saga/blob/250e722419f9fcaade729c51aaf5d97a3de2b16a/src/internal/utils.js#L11

src/internal/channel.js
```diff
@@ -13,9 +13,21 @@ export function emitter() {
...
   }
 
+  function emitAll(items) {
+    for (var i = 0, len = items.length; i < len; i++) {
+      emit(items[i])
+    }
+  }
+
   function emit(item) {
+    if (item === undefined) {
```
Andarist: its only guarding from the `undefined`, but I think `null` should filtered out too, cant use simple `!item` though as we need to pass i.e. `0` (in case somebody emit numbers)

docs/api/README.md
```diff
@@ -56,6 +56,11 @@ Creates a Redux middleware and connects the Sagas to the Redux Store
 
   - `sagaMonitor` : [SagaMonitor](#sagamonitor) - If a Saga Monitor is provided, the middleware will deliver monitoring events to the monitor.
 
+  - `processAction` : Function - all dispatched actions will be transformed by this function before delivering to the Sagas. Custom implementation needs to make sure to never mutate the action. If custom implementation:
+    - returns `undefined` for an action, that action will be ignored by the Sagas,
```
Andarist: should be "`undefined` or `null`"

src/internal/channel.js
```diff
@@ -13,9 +13,21 @@ export function emitter() {
...
+      return
+    }
+    if (Array.isArray(item)) {
+      emitAll(item);
```
Andarist: should be `return`ed after `emitAll` as it would emit twice in your scenario
1. batch action
1. unpacked batch actions

src/internal/channel.js
Andarist: wondering also if `processAction` could be made a concern of `emit` itself, right now you are passing already processed actions into `emit`

src/internal/channel.js
```diff
@@ -13,9 +13,21 @@ export function emitter() {
...
+    for (var i = 0, len = items.length; i < len; i++) {
+      emit(items[i])
+    }
+  }
+
   function emit(item) {
+    if (item === undefined) {
+      return
+    }
+    if (Array.isArray(item)) {
```
Andarist: thats a breaking change as even without `processAction` option is theoretically possible to dispatch an array and with that change it would suddenly start to emit each item in it instead of the original item

src/internal/channel.js
yelouafi: agree on the null

yelouafi:
I feel like the contract of emit is broken into 2 pieces: 

1-the custom emit dispatches an array or null/undfined
2- the builtin emit interprets the array/null or undefined

I don't feel much putting the constraint on the builtin emit to interpret the passed result. IMO, could be better if the custom emit receives the builtin emit as parameter and decides itself fully on how to interpret the action.

Besides avoiding the breaking change it could also make the custom emit more flexible, (eg could potentially allow async transoformations)

Andarist:
do you have some better naming for this option than `emit` or `customEmit`? I guess `customEmit` is not the worst name in the world, but always hated `custom`, `original` etc

yelouafi:
dont know, `processAction`, `mapAction` looks more appropriate to me. As here we're not really emitting actions but just transforming the stream of actions already dispatched

Andarist:
But emitting them by hand to the saga internals after processing the
original one

yelouafi:
yeah that's right, but I used map because we're like in the middle of a chain where we forward dispatched actions to sagas

Andarist:
Process is better than as map normally indicate that there will be some
outcome of the operation - and its up to the user, he can prevent from emit

pbadenski:
@yelouafi @Andarist what do you suggest is the next step? I can take another shot at the implementation based on suggestions?

Andarist:
Yeah, sure, I would see it as loose vatiation based on the snippet Ive
posted above -
https://github.com/yelouafi/redux-saga/pull/673#issuecomment-265712490 .
With other naming and possibly other internal implementation, but the API
idea should stay similar. Ofc if you see it differently, please raise your
concerns!

yelouafi:
API suggestion: make the option a higher order function, which takes the builtin emitter and returns another emitter. For example

```js
createSagaMiddleware({
  [whatevername](emitter) {
    return action => {
      if(action == null) return // do nothing
      else if(Array.isArray(action) ...
      else emitter(action)
    }
  }
}
```

In the middleware code then, the default emitter option will be just the identity function

```js
const sagaEmitter = (options.[whatevername] || identity)(emitter)
```

This way we don't impose any superfluous semantics on the builtin emitter (ie interpreting array or null actions). All interpretation can be done on user land by the custom emitter

pbadenski:
Done. (took me a while.. Christmas). Let me know what you think.

Andarist:
It looks fine, but the feature lacks the tests. If you want I can write them up, just let me know. Other than that it is ready to merge in.

pbadenski:
I added a simple test for custom emitter, feel free to extend. I'm struggling to come up with good scenarios so would certainly appreciate your help.

Andarist:
Well, I'm gonna use your origin problem as a test case (batched actions). Gonna merge it now. Thanks for your great work and iterating on the idea with us!

Andarist:
if you are interested [here](https://github.com/redux-saga/redux-saga/commit/7be2a8bafd8c8ff45e65692b4931f94d0ce23023) is the commit with my changes

pbadenski:
thanks! looking forward to next release :)

Andarist:
its already released! https://github.com/redux-saga/redux-saga/releases/tag/v0.14.2

pbadenski:
neat, thank you! :D

tugorez:
If any of you guys are interested [redux-batcher](https://github.com/tugorez/redux-batcher) is shipped with  the emmiter (ready to play well with sagas). 

``` javascript
import { emmiter } from 'redux-batcher';

createSagaMiddleware({ emmiter });
```

PR #681, opened 2016-12-09
Add example test with createMockTask

henrikra:
I have added example test with `createMockTask` because `fork` is often needed but examples on how to test it weren't available. Also there is **nothing** on `createMockTask` in the docs. Tell me what to improve :)


Andarist:
Thanks!

PR #687, opened 2016-12-12
Pushed downwards the stack scheduling of dispatched actions to channe…

Andarist:
…l's subscribe function, so store-like objects used by runSaga benefit the same scheduling fixes done previously in redux middleware

1. I was wondering where to put the test - I would lean towards changing its location to `test/proc/takeSync.js` + exporting `storeLike` from `test/runSaga.js` and then DRY both tests (the one which got fixed by scheduler changes), as they are exactly the same, just one using standard store and one storeLike

2. also I was wondering where to put scheduling now, should it be handled in `stdChannel`'s subscribe like now or should maybe it be handled in `eventChannel`'s subscribe (commented out code)? I think the latter would only fix rly far fetched scenarios, cause its **technically** possible to reference `store.dispatch` inside `eventChannel`'s listener, but that could be said about pretty everything, as there is always a possibility for developer to shoot himself in the foot. Wouldnt worry about it too much. I think that only store (or store-like) emits should be handled in a special way.

3. looking on `runSaga` API it was really confusing for me that it behaved pretty much like store, but not exactly. Real store's `subscriber` function is not receiving any arguments, its just notified about the change, but nothing more. Dont you think it's a slight problem? Maybe the difference should at least be expressed better in the docs?

WDYT?


yelouafi:
>I was wondering where to put the test - I would lean towards changing its location to test/proc/takeSync.js + exporting storeLike from test/runSaga.js and then DRY both tests (the one which got fixed by scheduler changes), as they are exactly the same, just one using standard store and one storeLike

Probably better to isolate the test in a separate file: i.e. exports (from the test file) a function which takes a store object with methods necessary to run the test `{ subscribe, dispatch, runSaga }`. The test function could then be imported from both takeSync and runSaga which can run the tests by providing the store object.

>also I was wondering where to put scheduling now, should it be handled in stdChannel's subscribe like now or should maybe it be handled in eventChannel

My approach to scheduling until now was to handle issue (i.e. edge cases) as they surface. So progressively we gain better understanding of the whole scheduling problem. So I'd let the scheduling in stdchannel for now

>I think the latter would only fix rly far fetched scenarios, cause its technically possible to reference store.dispatch inside eventChannel's listener, but that could be said about pretty everything, as there is always a possibility for developer to shoot himself in the foot

I think it's unlikely to happen (although not impossible) because eventChannel is typically used to interface external events like websockets. 

The scheduling issue appears [only] when there are synchronous effects called inside within the stack frame of others. All external events (browser, websockets) are dispatched in separate event queues (i.e. after the current stack is exhausted). And in edge cases of eventChannel dispatching events synchronously (eg app specific events) I think the scheduling issues can be easily avoided by using `Promise.resolve().then(() => emit())` which should give the developer a more predictable order of events.

>looking on runSaga API it was really confusing for me that it behaved pretty much like store, but not exactly. Real store's subscriber function is not receiving any arguments, its just notified about the change, but nothing more. Dont you think it's a slight problem? Maybe the difference should at least be expressed better in the docs?

Not sure to understand what you mean. runSaga's subscribe argument is different from store's subscribe. We need the input argument in order to fullfill take effects, in the middleware we get this from the middleware's action argument.

Andarist:
> >looking on runSaga API it was really confusing for me that it behaved pretty much like store, but not exactly. Real store's subscriber function is not receiving any arguments, its just notified about the change, but nothing more. Dont you think it's a slight problem? Maybe the difference should at least be expressed better in the docs?

> Not sure to understand what you mean. runSaga's subscribe argument is different from store's subscribe. We need the input argument in order to fullfill take effects, in the middleware we get this from the middleware's action argument.

Yes, I know thats the input argument is needed somewhere, just wondering if maybe it could be done in some other place, so the store like API remains the same as the real store. Didnt spent much time on that thought though, just it caused me a little confusion at first as `dispatch` and `getState` are 100% the same and suddenly `subscribe is not` so I had to stop and think how it works, how its implemented and why it is implemented this way.

Andarist:
@yelouafi 
I've pushed the fresh version of this branch and it should be ready to merge, however would like you to take a look before it.

test/runSaga.js
```diff
@@ -67,5 +63,13 @@ test('runSaga', assert => {
...
 
   task.done.catch(err => assert.fail(err))
-
 })
+
+test('put causing sync dispatch response in store-like subscriber', assert => {
+  const reducer = (state, action) => action.type
+  const store = storeLike(reducer, {})
+
+  const { rootSaga, runTest } = syncDispatchTest(assert, store)
```
yelouafi: wouldn't be simpler to pass `runSaga` as argument to `syncDispatchTest`. it can then run sagas and tests directly

Andarist: i didnt do that because of the API differences, I've changed it to to accept 3rd parameter which is a callback receiving a saga from the `syncDispatchTest ` and need to run a saga with appropriate API

yelouafi: ah right! missed that

yelouafi:
seems good to me.

PR #688, opened 2016-12-12
Added debounce helper

Andarist:
it should be roughly equivalent of:
```javascript
function* debounce(ms, pattern, worker, ...args) {
	let action
	const setAction = ac => action = ac

	while (true) {
		setAction( yield take(pattern) )

		while (true) {
			const { debounced, _action } = yield race({
				debounced: call(delay, ms),
				_action: take(pattern)
			})

			if (debounced) {
				yield fork(worker, ...args, ac)
				break
			}

			setAction(_action)
		}
	}
}
```

In my reasoning this type (forking) of debouncing versus `takeLatest` + `delay` on the beginning is more appropriate for general use cases. As it wont interrupt the asynchronous forked task in the middle which could potentially land the user in some inconsistent state. WDYT?

I see `takeLatest` + `delay` as other useful pattern to use, but the one that shouldnt be labeled as `debounce`.

Also I didnt write tests for it as this one https://github.com/yelouafi/redux-saga/issues/582 should be fixed first, cause `debounce` is ofc time-dependant.

In addition I've tried to simplify helpers creation and moved `END` handling to the `fsmIterator`, but it still stays an open question and the thing to be tested and checked.

### DIFF
--- a/src/internal/sagaHelpers.js
+++ b/src/internal/sagaHelpers.js
@@ -1,9 +1,9 @@
-import { END } from './channel'
 import { makeIterator, delay, is, deprecate } from './utils'
-import { take, fork, cancel, actionChannel, call } from './io'
+import { take, fork, cancel, call, actionChannel, race } from './io'
 import { buffers } from './buffers'
 
 const done = {done: true, value: undefined}
+const nextState = {done: false}
 const qEnd = {}
 
 function fsmIterator(fsm, q0, name = 'iterator') {
@@ -44,6 +44,33 @@ function safeName(patternOrChannel) {
   }
 }
 
+export function debounceHelper(delayLength, pattern, worker, ...args) {
+  let action, raceOutput
+
+  const yTake = {done: false, value: take(pattern)}
+  const yRace = {
+    done: false,
+    value: race({
+      action: take(pattern),
+      debounce: call(delay, delayLength)
+    })
+  }
+  const yFork = ac => ({done: false, value: fork(worker, ...args, ac)})
+
+  const setAction = ac => action = ac
+  const setRaceOutput = ro => raceOutput = ro
+
+  return fsmIterator({
+    q1() { return ['q2', yTake, setAction] },
+    q2() { return ['q3', yRace, setRaceOutput] },
+    q3() { return raceOutput.debounce
+      ? ['q1', yFork(action)]
+      : ['q4', raceOutput.action, setAction]
+    },
+    q4() { return ['q2'] }
+  }, 'q1', `debounce(${safeName(pattern)}, ${worker.name})`)
+}
+
 export function takeEveryHelper(patternOrChannel, worker, ...args) {
   const yTake = {done: false, value: take(patternOrChannel)}
   const yFork = ac => ({done: false, value: fork(worker, ...args, ac)})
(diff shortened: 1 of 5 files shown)


src/internal/sagaHelpers.js
```diff
@@ -21,7 +21,7 @@ function fsmIterator(fsm, q0, name = 'iterator') {
       let [q, output, _updateState] = fsm[qNext]()
       qNext = q
       updateState = _updateState
-      return qNext === qEnd ? done : output
+      return qNext === qEnd || output === END ? done : output
```
yelouafi: I think that the END test is not required anymore  b/c `proc` will anyway terminate the generator if it detects an END (END is returned explicitly only when using `takem` effects ). see

https://github.com/yelouafi/redux-saga/blob/master/src/internal/proc.js#L412
https://github.com/yelouafi/redux-saga/blob/master/src/internal/proc.js#L247-L250

I added this test when there wasn't yet automatic handling of END in proc but forgot to remove it later.

yelouafi:
Good catch on the cancel issue, didn't thought of it. This seems ok to me. 

OTOH [thinking out loud] this could be also done with actionChannel + custom buffer.

Andarist:
Well, as of `actionChannel + custom buffer` solution - that was my first thought (maybe not totally custom buffer, but utilising existing ones), but the problem is that you `take` the first item immediately and in the sequence of 1 `debounce` call you need to behave differently than when waiting for the leading call (or I might didnt come up with the better solution... yet ;) )

tried to work around this snippet which I use in my app, but unfortunately changing its logic to `debounce` is not straight-forward
```javascript
export function* trailingThrottle(ms, pattern, worker, ...args) {
	const throttleChannel = yield actionChannel(pattern, buffers.sliding(1))

	for (;;) {
		const leadingAction = yield take(throttleChannel)
		yield call(delay, ms)
		const [ action ] = yield flush(throttleChannel)
		yield fork(worker, ...args, action || leadingAction)
	}
}
```

Andarist:
smth like `take.preserve` could be added or `buffers.preserving` and maybe those could be leveraged to make a nicer implementation for `debounce` and at the same time be a general purpose API for everybody to use

yelouafi:
>smth like take.preserve could be added or buffers.preserving

Could you provide an example? 

After though I think in its current form, buffers alone can't be used for time dependent things (since they can't 'push' actions back to channels, currently they are passively managing storage).

We can add the 'push' capability to buffers but not sure if this is the right solution (keeping the buffers 'simple' as they are now is better IMO). 

Another way is implementing time transformations directly in channels: e.g. the channel function could accept a 'time-buffer' (sort of scheduler) as argument which could then manage time transformations. 

Also being thinking of adding an `observableChannel` (like eventChannel but get message from Rxjs-like observable sources)

Andarist:
When I think about it now `buffers.preserving` doesnt make any sense.

`take.preserve` would work like this:
```javascript
const chan = channel()
yield put(chan, 'test')
yield take.preserving() // 'test'
yield flush(chan) // ['test']
```
its only useful for channels and not patterns ofc 

Didnt think yet how to leverage it in the `debounce` helper, but it could be also useful for sending 'signals' from the channels, when there are multiple takers on them and all of them needs to be notified about new item.

Andarist:
> Another way is implementing time transformations directly in channels: e.g. the channel function could accept a 'time-buffer' (sort of scheduler) as argument which could then manage time transformations.

Could you provide the example snippet for this?

yelouafi:
dont have a precise idea, but basically smth like

```js
// simple scheduler
function delayScheduler(delay) {
  return put => item => {
    setTimeout(() => put(item), delay )
  }
}
const chan = channel(buffer, scheduler)
```

Another option could be to implement an observableChannel which could get events from an Rx like observable. Since Observables are really great to handle time based transformations

restrry:
closed in favour of https://github.com/redux-saga/redux-saga/pull/1454

PR #690, opened 2016-12-12
Fixes CHANNEL_END public export (fixes #689)

### LINKED ISSUE #689: Bundling with Rollup fails due to missing export
Hello,

i'm trying to bundle my project using redux-saga and stumbled upon an issue.

using the node-resolve plugin for rollup and the `{ jsnext: true }` option gives me the following error:

`'CHANNEL_END' is not exported by '$PATH_TO_PROJECT$\node_modules\redux-saga\es\internal\io.js' (imported by '$PATH_TO_PROJECT$\node_modules\redux-saga\es\utils.js')
Error: 'CHANNEL_END' is not exported by '$PATH_TO_PROJECT$\node_modules\redux-saga\es\internal\io.js' (imported by '$PATH_TO_PROJECT$\node_modules\redux-saga\es\utils.js')
    at Module.traceExport ($PATH_TO_PROJECT$\node_modules\rollup\src\Module.js:376:17)
    ...`

Looking at the `es/internal/io.js` module there is no export for this but the `es/utils.js` re-exports this constant.

**es/utils.js**
`export { CHANNEL_END, asEffect } from './internal/io';`

Setting the `{ jsnext: false }` option solves the issue as a workaround but also disables all other es6 imports.

Did I miss something or is this a small bug?

Thanks in advance.

Andarist:
The issue was reported, as `redux-saga/utils` tries to reexport non-existent import. From what I see this reexport was introduced solely for redux-saga-devtools and is used by it here - https://github.com/yelouafi/redux-saga-devtools/blob/6721e6cd4f561973b8b80c6ac4b1b1a8df139607/src/components/SagaValue/SagaValue.js#L10 .

I think I've chose correct import, but please verify the intention here as there are 2 similar things in the codebase.

1. https://github.com/yelouafi/redux-saga/blob/283cc6c168e5773854e44dce66cda70bb1a1d0a9/src/internal/proc.js#L9 - I'm wondering about theese lines - Whats the purpose of this being an object with `toString` method?

2. https://github.com/yelouafi/redux-saga/blob/283cc6c168e5773854e44dce66cda70bb1a1d0a9/src/internal/channel.js#L5

Im curious if they should be unified? And also which one should be used as public export?


yelouafi:
Oops! sorry for this mess.

Actually it's the channel export which should be exposed. let me explain

The channel's `END` is used to resolve takes from closed channels (call to channel.take). In proc, the `END` is interpreted depending on the `maybe` flag of take effect (`takem` creates a take effect with maybe = true). 

by default by receiving END, proc will terminate the generator this is done by resolving the take with a special value `CHANNEL_END` (cf https://github.com/yelouafi/redux-saga/blob/master/src/internal/proc.js#L412). it's a special value interpreted by next (cf https://github.com/yelouafi/redux-saga/blob/master/src/internal/proc.js#L247-L250).

if maybe=true (using takem) then the dev. is willing to trap the END and handles it by himeself, in this case the take will be resolved with the (channel's) END. In the saga devtools we check for it in order to show a custom display (instead of the default `{type: CHANNEL_END_TYPE}`)

`CHANNEL_END` is internal to proc (can't remember why it's exported, perhaps for possible usage in tests although I think it's not used there).

Andarist:
couldnt `proc`'s CHANNEL_END just be imported from the `./channel`? Is there need for both?

yelouafi:
>couldnt proc's CHANNEL_END just be imported from the ./channel? Is there need for both?

No, because in proc `next` function couldn't interpret the END as either terminating the generator or resolving the effect

Andarist:
Ok, gonna look more into that, but merging for now as the original issue is resolved.

Also still aint sure about this
> https://github.com/yelouafi/redux-saga/blob/283cc6c168e5773854e44dce66cda70bb1a1d0a9/src/internal/proc.js#L9 - I'm wondering about theese lines - Whats the purpose of this being an object with toString method?

PR #691, opened 2016-12-12
Make it clear in the docs that "Delay" is built-in

hmillison:
Building on #335 

I noticed when reading through the Redux Saga docs that the `delay` function is built-in to Redux Saga.

I think it would be helpful to make that more clear in the examples for throttling and debouncing. The only time that it is mentioned before is the final example in this section of the docs.

### DIFF
--- a/docs/recipes/README.md
+++ b/docs/recipes/README.md
@@ -20,11 +20,11 @@ By using this helper the `watchInput` won't start a new `handleInput` task for 5
 
 ## Debouncing
 
-To debounce a sequence, put the `delay` in the forked task:
+To debounce a sequence, put the built-in `delay` helper in the forked task:
 
 ```javascript
 
-const delay = (ms) => new Promise(resolve => setTimeout(resolve, ms))
+import { delay } from 'redux-saga'
 
 function* handleInput(input) {
   // debounce by 500ms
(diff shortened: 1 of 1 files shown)


docs/recipes/README.md
```diff
@@ -44,13 +44,13 @@ function* watchInput() {
 }
 ```
 
-In the above example `handleInput` waits for 500ms before performing its logic. If the user types something during this period we'll get more `INPUT_CHANGED` actions. Since `handleInput` will still be blocked in the `delay` call, it'll be cancelled by `watchInput` before it can start performing its logic.
+The `delay` function implements a simple debounce using a Promise. In the above example `handleInput` waits for 500ms before performing its logic. If the user types something during this period we'll get more `INPUT_CHANGED` actions. Since `handleInput` will still be blocked in the `delay` call, it'll be cancelled by `watchInput` before it can start performing its logic.
```
Andarist: You could put its implementation here between the lines, so people dont think its some kind of magic, but just a simple function returning a promise.

Andarist:
Thanks! That's a great input easing a newcomers life. Ive put a tiny comment, would you like to address it before merging?

hmillison:
Thank you for the prompt feedback! I updated my change with your suggestion.

Andarist:
Thanks! looks good :)

PR #692, opened 2016-12-12
Update docs to avoid confusion over saga helper implementations

odziem:
This is a small documentation update that attempts to reduce confusion about the implementation of the saga helpers. They are no longer actually implemented as generators. However, my understanding is that the examples in the docs are meant more as higher level illustrations of how the saga helpers could be implemented.


Andarist:
Hm, I guess it is more strict now. Thanks for a PR!

Issue #665, opened 2016-11-30
Add a cheatsheet to the docs for what effects are blocking/non-blocking


granmoe:
I see a lot of confusion from newcomers regarding what effects are blocking/non-blocking. I think there should be a "cheatsheet" for this in the docs. The definitive, concise source of truth on this topic.

granmoe:
Currently, there is a section in the glossary that addresses this, but it isn't complete and it's kind of hidden away.

Andarist:
Would you be willing to create an API section with this cheatsheet?

granmoe:
I'd be happy to :-) I'll open a PR for it probably some time this week

granmoe:
Closing per PR #693

PR #693, opened 2016-12-13
Added a blocking / non-blocking cheatsheet to the API reference section of the docs.

granmoe:
Please let me know if anything is incorrect, or if you have any suggestions at all. Also, I'm not totally clear on how GitBooks generates the markup to make name attributes for the # links.

Should the helpers, select, actionChannel be included here?

See #665 

@Andarist

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -46,6 +46,10 @@
   * [`buffers`](#buffers)
   * [`delay(ms, [val])`](#delayms-val)
 
+# Cheatsheets
+
+* [Blocking / Non-blocking](#blockingnonblocking)
+
 ## Middleware API
 
 ### `createSagaMiddleware(options)`
@@ -959,3 +963,26 @@ Provides some common buffers
 ### `delay(ms, [val])`
 
 Returns a Promise that will resolve after `ms` milliseconds with `val`.
+
+## Cheatsheets
+
+### Blocking / Non-blocking
+
+| Name | Blocking |
+| -- | -- |
+| take | Yes |
+| takem | Yes |
+| put | No |
+| put.sync | Yes |
+| put(channel, action) | Sometimes (see API reference) |
+| call | Yes |
+| apply | Yes |
+| cps | Yes |
+| fork | No |
+| spawn | No |
+| join | Yes |
+| cancel | Yes |
+| flush | Yes |
+| cancelled | Yes |
+| race | Yes |
+| [...effects] | Yes |
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -970,11 +970,15 @@ Returns a Promise that will resolve after `ms` milliseconds with `val`.
 
 | Name | Blocking |
 | -- | -- |
+| takeEvery | No |
+| takeLatest | No |
+| throttle | Yes |
 | take | Yes |
+| take(channel) | Sometimes (see API reference) |
 | takem | Yes |
 | put | No |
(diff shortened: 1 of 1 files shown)


granmoe:
screen shot 2016-12-12 at 8 56 58 pm

Andarist:
`put.sync` got renamed to `put.resolve` (not released yet, but already on the `master`)

> Should the helpers, select, actionChannel be included here?

Yeah, I think its useful to keep all effects here (helpers are gonna be effects as well with the new release), all of them being non-blocking.

`put(channel, action)` is never blocking I think (only scheduler by `asap` internally), although the above comment ("sometimes, see API reference") should be for `take(channel)`

Also I would like for @yelouafi to take a look on it, before merging

docs/api/README.md
```diff
@@ -959,3 +963,26 @@ Provides some common buffers
...
+## Cheatsheets
+
+### Blocking / Non-blocking
+
+| Name | Blocking |
+| -- | -- |
+| take | Yes |
+| takem | Yes |
+| put | No |
+| put.sync | Yes |
```
restrry: worth adding `put.resolve` here?

granmoe:
OK, updated per your comments. I'm not sure what, if anything, to say for `select` and `actionChannel`. Should those be included? And what should they say?

granmoe:
Uggghhh...I needed to change my commit email, and I think I messed up my branch. Once we finalize the changes, I can create a clean branch and open a new PR. Sorry

Andarist:
both of `select` and `actionChannel` are non-blocking as they simply run a functions, but are wrapped as effects so its easy for the saga to distinguish them and run appropriately, so as long as they are effects they should be included in the cheat sheet.

about the messed up branch - you could probably just reset till your good commits, just keep a copy so you dont lose your changes by accident ;)

granmoe:
No I cloned a bare repo of my forked repo, then rewrote commit history with a script, then force pushed it. So actually my whole fork is bad :-(

Hmmm...the diff looks correct and there are no conflicts. Could we potentially just keep this branch? I think I just have an extraneous merge commit in here, but my actual diff is good--it only has the stuff that I changed.

I'll add `select` and `actionChannel`

docs/api/README.md
```diff
@@ -959,3 +963,32 @@ Provides some common buffers
...
+| fork | No |
+| spawn | No |
+| join | Yes |
+| cancel | Yes |
+| select | No |
+| actionChannel | No |
+| flush | Yes |
+| cancelled | Yes |
+| race | Yes |
+| [...effects] | Yes |
```
Andarist: hm, it should have a comment of smth like "depends on the effects, if at least 1 is blocking then it will be blocking, no-blocking otherwise" - could be phrased much better ofc

Andarist:
Although the diff is good, I wouldnt like to merge this 'broken' branch as it would bloat history too much with unneccessary duplicated commits.

Do not close the PR though, I'll try later to play with your branch and its history and maybe we could recover it.

Andarist:
```sh
git reset --mixed HEAD~3
git stash
git reset --hard HEAD~1
git stash pop
git commit -am 'YOUR NEW MESSAGE'
git push --force
```

granmoe:
Thanks @Andarist! Your git fu is strong :-)

Branch looks clean now

Andarist:
Thanks for your work!

PR #696, opened 2016-12-13
Fixed sagas.js example in README.md

sublimeye:
takeEvery, takeLatest must be imported from 'redux-saga' module, not 'redux-saga/effects'

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,8 @@ The Component dispatches a plain Object action to the Store. We'll create a Saga
 #### `sagas.js`
 
 ```javascript
-import { call, put, takeEvery, takeLatest } from 'redux-saga/effects'
+import { takeEvery, takeLatest } from 'redux-saga'
+import { call, put } from 'redux-saga/effects'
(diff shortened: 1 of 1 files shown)


Andarist:
Actually we are in the process of migrating those helpers to the `effects` and you caught us unguarded ;) Within a week I think the new `redux-saga` version should be released with that changes. Probably docs were adjusted prematurely, sorry for the confusion.

sublimeye:
@Andarist Thanks for the update. Yeah, I was wondering myself why those helpers were out of `effects` at all ... I'm new to redux-saga and spend like half a day on trying to figure out what's wrong (the error was really weird and didn't imply anyhow that a module or object wasn't imported or undefined)

Andarist:
@sublimeye sorry for this mess, I forgot that the main `README.md` is accessible just like that on the repo site, other docs were more safe to update even before releasing the new version and the docs site needs to be published manually and I've planned to do it after the release

That being said I'll definitely release the newest version before the weekend

PR #702, opened 2016-12-16
Reworked throttle test to be dependant on the custom mock clock isnte…

### LINKED ISSUE #582: throttle test is randomly failing
image

This happens randomly it seems, and I'm guessing it's timing dependent.

Andarist:
…ad of the native one (fixes #582)


yelouafi:
looks much better

PR #704, opened 2016-12-16
[TypeScript] Added actionDispatched to Monitor interface

### LINKED ISSUE #700: [TypeScript] Update Monitor Interface for 0.13 changes
[Monitor Interface](https://github.com/yelouafi/redux-saga/blob/master/index.d.ts#L18) is missing `actionDispatched` from #609

CarsonF:
Per #609

Fixes #700

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,4 @@
-import {Middleware} from "redux";
+import {Action, Middleware} from "redux";
 import {Effect, Pattern} from "./effects";
 import {Task, Buffer, Channel, Predicate} from "./types";
 
@@ -26,6 +26,7 @@ export interface Monitor {
   effectResolved(effectId: number, res: any): void;
   effectRejected(effectId: number, err: any): void;
   effectCancelled(effectId: number): void;
+  actionDispatched<A extends Action>(action: A): void;
 }
 
 


Andarist:
cc @aikoven

aikoven:
👍

aikoven:
Oh wait, more accurate would be `actionDispatched<A extends Action>(action: A)`

CarsonF:
@aikoven better now?

aikoven:
It's good now. Thanks

CarsonF:
I'm curious what the difference is there?

aikoven:
That's just how it's done in Redux: https://github.com/reactjs/redux/blob/v3.6.0/index.d.ts#L46

I'm not sure if it adds any strictness right now, but I think that it describes the function contract more precisely. In TypeScript saying that something has interface `T` doesn't always mean that it has properties from `T`. Sometimes it means that there are no other properties that aren't in `T`, e.g.:
```ts
interface T {
  prop: string;
}

// this would fail:
const t: T = {prop: 'foo', otherProp: 'bar'};
```
However, I can't think of any non-exotic case like this for `Monitor.actionDispatched`.

PR #708, opened 2016-12-18
Fix imports in README examples

shawngustaw:


shawngustaw:
I'm not sure if this changed recently but the examples in the README don't match the examples in the documentation.

Andarist:
Thanks for the PR, actually README got prematurely updated - both `import`s will be available with the new release, which gonna happen in following week. 

Sorry for this mess!

PR #709, opened 2016-12-19
Renamed takem to take.maybe for clarity, also it pairs up nicely with…

Andarist:
… convention used for put.resolve

Its just a proposal ofc, but I find it more descriptive naming as `takem` is kinda cryptic and I always forgot what it does, as Im not using it. Whereas `take.maybe` should be less forgotable.


yelouafi:
agree; it's more consistent

PR #711, opened 2016-12-20
[WIP] Code improvement in the beginner tutorial

ummahusla:
Initially, I was following the beginner tutorial and I've noticed that my code doesn't work. Obviously, later on, I figured out that it was a minor typo I probably removed more code when I supposed to do. I still think that adding `value={store.getState()}` to `<Counter />` will remove that issue and the code will work even if the user is going just simply to copy and paste it.

### DIFF
--- a/docs/introduction/BeginnerTutorial.md
+++ b/docs/introduction/BeginnerTutorial.md
@@ -248,9 +248,10 @@ Well, `redux-saga` provides a way to make the above statement possible. Instead 
 
 ```javascript
 // ...
+import { delay, takeEvery } from 'redux-saga'
 import { put, call } from 'redux-saga/effects'
-import { delay } from 'redux-saga'
 
+// Our worker Saga: will perform the async increment task
 export function* incrementAsync() {
   // use the call Effect
   yield call(delay, 1000)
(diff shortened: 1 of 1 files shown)


Andarist:
Hey,
thanks for the PR - its probably the good idea to include in this tutorial even more of the code involved in real working example - not the `value` props only. It in fact only shows the relevant parts of the code, but there is no indication that there should be more in the final code.

Could you adapt your PR to include all of the `Counter` and `render` code with indication what was added on each step?

ummahusla:
@Andarist Hey, yeah I will think what we can do with that Beginner tutorial and will try to take care of it. Moreover, I've seen that there are issues in the repository for that tutorial. I will evaluate the errors and find anything else we might need to add here.

ummahusla:
@Andarist Hi, can you please review my work. I guess it's ready to be merged :rocket:

docs/introduction/BeginnerTutorial.md
```diff
@@ -247,8 +250,8 @@ Well, `redux-saga` provides a way to make the above statement possible. Instead
 
 ```javascript
 // ...
+import { delay, takeEvery } from 'redux-saga'
```
Andarist: could you import `takeEvery` also from the `'redux-saga/effects'`? Its upcoming change and will be released before the weekend

ummahusla:
@Andarist Thanks 👍

PR #716, opened 2016-12-22
blocking/non-blocking effect. fix table and link in docs

restrry:
according to [docs](https://help.github.com/articles/organizing-information-with-tables/#creating-a-table) table header row should has at least 3 hyphens. also link in docs is broken


Andarist:
Thanks for the fix!

PR #721, opened 2016-12-27
check buffer correctness 

restrry:
we check `notUndef ` twice, so seems to be a typo


Andarist:
Thanks for noticing! Gonna merge right away.

PR #724, opened 2016-12-27
Reworked take/takem handling internally a little bit, so its more cle…

### LINKED ISSUE #689: Bundling with Rollup fails due to missing export

Andarist:
…ar. Also it should fix #689 

Im not completely sold on my approach with `TERMINATE` flag added to the `END` in case of `take`. But I felt that having really similar names and concepts across 2 files which were not the same was really confusing and at least some more clearer names should be used. 

Anyway - `TERMINATE` is just a proposition which can be reverted easily, rest of the PR stands on its own (just stylistic changes and 1 function extraction I think)

Also with the `TERMINATE` approach I've got rid public `CHANNEL_END` export so [this](https://github.com/redux-saga/redux-saga-devtools/blob/6721e6cd4f561973b8b80c6ac4b1b1a8df139607/src/components/SagaValue/SagaValue.js#L10) could be testes just like this:
```javascript
import { is, END } from 'redux-saga/utils'
// ...
  else if(value === END.type) { 
// ...
```
or alternatively `isEnd` could be exported publicly (which might be the better approach).

Still wondering why such construct is being used - `const TASK_CANCEL = {toString() { return sym('TASK_CANCEL') }}`. Does it having `toString` property have some hidden purpose I miss?


yelouafi:
You can view CHANNEL_END as type of completion for sagas. Actually in proc, the continuation fn (next) function accept 2 args (value and isErr flag) with this we represent 2 types of completion: success & error.

The 3rd type is cancellation which is denoted by TASK_CANCEL special value. this value is 'contagious' meaning any saga that gets this value as argument in its next cont. will also be cancelled (this how cancellation is propagated to joiners).

CHANNEL_END is the 4rth type, which results by taking from a closed channel, but the value is not 'contagious': ie a calling/joining a saga that have completed with CHANNEL_END won't terminate the caller Saga (b/c for now we just 'return' the completed saga with undefined). TBH I wasn't (and still not) sure about this one (to propagate or not). But if make this value propagate as well then it means it's the same type as TASK_CANCEL and the 2 type should get merged.

I think I've put the toString to show the completion values of sagas on devtools in a user friendly way.

yelouafi:
Actually in a language with Disjoint union types like Haskell or Elm, I'd have represented this with some data type like

```hs
data Completion a = Next a | Cancel CancelReason | Error Error | CHANNEL_END 
```

Andarist:
Yeah, I grasped the reasoning behind the code, just found it confusing to have seperate `END` from [`src/internal/channel`](https://github.com/redux-saga/redux-saga/blob/23f1e0f6e226d3e05897f2709705c0c65cfd3eb1/src/internal/channel.js#L5-L7):
```javascript
const CHANNEL_END_TYPE = '@@redux-saga/CHANNEL_END'
export const END = {type: CHANNEL_END_TYPE}
```
and the one in [`src/internal/proc`](https://github.com/redux-saga/redux-saga/blob/23f1e0f6e226d3e05897f2709705c0c65cfd3eb1/src/internal/proc.js#L9):
```javascript
export const CHANNEL_END = {toString() { return '@@redux-saga/CHANNEL_END' }}
```

This PR is solely for unifying those, however the validity of the approach used (or even need to do it) must be assessed by you

yelouafi:
Dont know. Personally I'd prefer to have separate values for the 2 since each one

Andarist:
maybe the proc's one should be called `CHANNEL_END_TERMINATION` then?

yelouafi:
Yeah, it's better to avoid confusion. We can also adopt an explicit naming convention for termination values: SAGA_TERMINATION_CANCELLED, SAGA_TERMINATION_CHANNEL_END ...

Andarist:
or maybe `SAGA_TERMINATION` with `reason` field? would probably come handy for tracking cancellations

I was trying to grasp how cancellation cause of uncaught error and cause of the `cancel` effects work and propagate through tasks to track the chain so it (canceled tasks) can be printed [here](https://github.com/redux-saga/redux-saga/blob/5858a7ce104c5b9d445e200af5922bb21e376d21/src/internal/proc.js#L278) and also omitted when cancellation happened intentionally (by a `cancel` effect)

But what I thought would take only little of time, happened to be not so easy and hadnt yet time to go back to this idea and debug further.

yelouafi:
>or maybe SAGA_TERMINATION with reason field? would probably come handy for tracking cancellations

I'd prefer to separate type info from payload info. Perhaps we could use smth like the Effect siganature

```js
{ [SAGA_TERMINATION]: CANCEL, reason }
{ [SAGA_TERMINATION]: CHANNEL_END }
...
```

In the earlier versions of redux-saga (when cancellations were modeled as special type of Errors) the reason info was stored in the message field on the Error class. After redux-saga switched to Cancellation as a third state, the reason info was discarded (also no one has reclaimed it)

Andarist:
wouldnt in this case `{ [SAGA_TERMINATION]: CANCEL, reason }` reason be kinda already included there? I mean the `CANCEL` value, it would indicate the already a reason - explicit `cancel` from the user

yelouafi:
`[SAGA_TERMINATION]` indicates the type of completion, because in [`proc#next`](https://github.com/redux-saga/redux-saga/blob/master/src/internal/proc.js#L230) we handle each type differently

Andarist:
yes, but still with mentioned example (`{ [SAGA_TERMINATION]: CANCEL, reason }`) im not sure if there is a purpose of both value of `[SAGA_TERMINATION]` field (here `CANCEL`) + the `reason` field.

Arent they the same in the end? 

With the proposed `SAGA_TERMINATION` I would now implement this as similarly to the `runEffect`. So something like:
```javascript
let terminationType
if (terminationType = asTermination(arg)) {
	// delegation to each method, for various termination types:
	// err ?
	// TASK_CANCEL, maybe 2 types here - explicit (user-induced) and implicit (saga-induced)
	// CHANNEL_END
	return
}
// rest of the `.next()` logic
```

yelouafi:
>es, but still with mentioned example ({ [SAGA_TERMINATION]: CANCEL, reason }) im not sure if there is a purpose of both value of [SAGA_TERMINATION] field (here CANCEL) + the reason field.

Ah, perhaps I didn't understand properly what you meant by 

>or maybe SAGA_TERMINATION with reason field? would probably come handy for tracking cancellations

I thought you're talking about having additional infos about the origins of cancellations (manual, automatic in race or parallel)

Andarist:
> > or maybe SAGA_TERMINATION with reason field? would probably come handy for tracking cancellations
> I thought you're talking about having additional infos about the origins of cancellations (manual, automatic in race or parallel)

Yeah, that was an idea once, forgot a little about it :P sorry for the confusion. Ive missed that in my understanding now. Anyway - I'll probably just try to implement this what we were talking about and such cases will just show up during the implementation I guess - whats needed and whats not.

PR #728, opened 2016-12-28
Updated TS definition to include saga helpers in redux-saga/effects

### LINKED ISSUE #720: Typescript - Saga 0.14 - Mismatch between console warning & definition files
With version 0.14 there is a new console warning:
```
import takeEvery from 'redux-saga' has been deprecated in favor of import takeEvery from 'redux-saga/effects'.
The latter will not work with yield*, as helper effects are wrapped automatically for you in fork effect.
Therefore yield takeEvery will return task descriptor to your saga and execute next lines of code. 
```
If using Typescript, changing the import to `redux-saga/effects` is not supported by included Typescript definition files.

Should we move the helpers across from index.d.ts to effects.d.ts, or copy it until the deprecation is enforced?

Happy to provide PR or assist.

Mark.

mcrawshaw:
This was required to match library changes in version 0.14.

Resolves #720.

### DIFF
--- a/effects.d.ts
+++ b/effects.d.ts
@@ -192,3 +192,62 @@ export type Effect =
   RaceEffect | CallEffect |
   CpsEffect | ForkEffect | JoinEffect | CancelEffect | SelectEffect |
   ActionChannelEffect<any> | CancelledEffect;
+
+
+type HelperFunc0<A> = (action?: A) => any;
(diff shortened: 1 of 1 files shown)


Andarist:
cc @aikoven

aikoven:
👍

Issue #518, opened 2016-09-01
Testing is Done Right


3LOK:
Hi,

My comment is about: https://redux-saga.github.io/redux-saga/docs/advanced/Testing.html

I think the way you see testing sagas is wrong and puts emphasis on exact 'implementation', and not 'result'. 

The way you do tests: Check the call for each 'yield', and muck a return value to that yield. In this case if I change the implementation of my saga (in such a way that doesn't affect the results), I need to change me tests. Any refactoring of code will necessarily mean changing the tests.

My view is that your tests check that you implemented your saga in a very very specific way and not that the saga is doing what it needs to be doing. 

The way we write test for our sagas is:
1. Set up a test store with a specific scenario. Test store includes a reducer that interacts with actions that are fired by the saga.
2. Run the saga using the standard redux-saga framework.
3. See the side effects (actions that were sent).

I think this way of testing puts the emphasis where it needs to be. If we refactor the saga, and it still does exactly what it needs to do, the tests won't need to change (making the tests helpful for refactoring).

WDYT?

Andarist:
I kinda feel the same. In my tests I wouldnt like to focus on underlaying implementation, just like in the 'normal' code. For that reason I'm not even testing my sagas right now. Dont want to have my tests broken if I rearrange 2 lines which dont even change the overall behaviour

slorber:
It's like other tests: do you want to unit-test the details of your code, or do you want to do integration tests of the behavior of the saga and your store, without having to deal implementation details.

I've always felt doing blackbox testing is a superior idea, as testing implementation details means that changing implementation details will likely break tests.

For example, do you want to test your classes that are used to build a REST API, or do you want to test the API directly? Personnally I think it's better to test the API directly, and don't care that much about unit tests, except rare cases where it's worth it because the function to test is likely complex and not going to change (like date/math utilities...)

I guess the subject is also quite related to new debates we've seen on twitter that say it's not really worth it to test React component outputs because when changing React component it will always break the test. That's probably why Jest's snapshot testing feature is gaining interest of the community


If redux-saga gives you good testability for yielded effects, it does not mean you have to test all your code in this fashion.

3LOK:
I'll make sure to open-source our redux-saga testing infrastructure soon. 

Basically what we wrote is a component that you construct with full initial "state", an optional list of reducers (for more complex tests), and the sagas you need (including the tested one). Then our tests look like:

```javascript
[...]

Issue #586, opened 2016-10-14
[RFC] Effect Interceptors


yelouafi:
related #518

Redux Saga allows unit testing Sagas by using the Iterator approach. We iterate over the generator getting next values, and inspect the returned effects by some form of equality testing (eg deepEqual)

The approach allows testing a workflow without mocking APIs but some developpers raised an understandable concern: it couples the test to the internal implementation. Some may find this as desirable (test acts as specification of the Saga) but I think the lib can (and should) allow more flexibility around testing story.

Actually the lib offers only a choice between 2 extremes
1. Unit testing Saga by iterating over all the yielded effects
2. Integration testing using the classic mock approach (mocking the Redux store, APIs ...)

So I propose a new feature I called: _Effect Interceptors_. The idea is to make possible a fine-grained control over Effect resolution when running Sagas (using `runSaga` or through `sagaMiddeware.run`) we can provide an object which can intercept yielded effects and either
1. return a custom resolution value
2. execute the redux-saga's builtin interpreter
3. or combine both approaches

Example. Considering the Saga below

``` js

function* watchActions() {
  yield takeEvery('ACTION', handleActions)
}

function* handleActions(action) {
  const {data, error} = yield call(api.fetchData, action.payload)
  if(data) {
    yield put({type: 'SUCCESS', data})
  } else {
    yield put({type: 'ERROR', error})
  }
}
```

We can either test the whole `watchActions` workflow without mocking the `api.fetchData` function. We can let the middleware execute all the Effects normally and only intercept the execution of `call` effects to the API.

``` js
const mockApiResults = [
  { data: 'mock data' },   // the 1st call will return this result
  { error: 'mock error' }, // the 2nd call will return this result
  ... 
]

const dispatchedActions = []

sagaMiddleware = createSagaMiddleware({
  interceptors: {
    call({context, fn, args}, execBuiltin, cb) {
      if(fn === api.fetchData) {
        cb(mockApiResults.shift())
      } else {
        execBuiltin(cb)
      }
    },
    put({action}, execBuiltin, cb) {
       // exec both builtin (dispatch to store) && custom (store dispatched actions)
       //  We could have monkey-patched the store.dispatch function but this would give us
       //  all dispatched actions (both dispatched by the Saga and from outside)
       // With this we're only interested in the actions dispatched by this Saga

       dispatchedActions.push(action)
       execBuiltin(cb)
    }
  } 
})

[...]

PR #729, opened 2016-12-29
Feat/run saga middleware

eloytoro:
Relates to https://github.com/redux-saga/redux-saga/issues/518#issuecomment-269558668

Allow for `runSaga` to take an extra `middleware` param thus allowing to "hot-replace" effects just before they're run by the saga middleware.

This would allow for integration testing while still being able to mock effects that interact with other logic


src/internal/proc.js
```diff
@@ -420,6 +408,28 @@ export default function proc(
...
+      : (is.notUndef(data = asEffect.call(effect)))          ? runCallEffect(data, effectId, cb)
+      : (is.notUndef(data = asEffect.cps(effect)))           ? runCPSEffect(data, cb)
+      : (is.notUndef(data = asEffect.fork(effect)))          ? runForkEffect(data, effectId, cb)
+      : (is.notUndef(data = asEffect.join(effect)))          ? runJoinEffect(data, cb)
+      : (is.notUndef(data = asEffect.cancel(effect)))        ? runCancelEffect(data, cb)
+      : (is.notUndef(data = asEffect.select(effect)))        ? runSelectEffect(data, cb)
+      : (is.notUndef(data = asEffect.actionChannel(effect))) ? runChannelEffect(data, cb)
+      : (is.notUndef(data = asEffect.flush(effect)))         ? runFlushEffect(data, cb)
+      : (is.notUndef(data = asEffect.cancelled(effect)))     ? runCancelledEffect(data, cb)
+      : /* anything else returned as is        */              cb(effect)
```
Andarist: this one should not be a part of `runDeclarativeEffect` i think

src/internal/proc.js
```diff
@@ -420,6 +408,28 @@ export default function proc(
     cb.cancel = takeCb.cancel
   }
 
+  function runDeclarativeEffect(effect, effectId, cb) {
+    if (options.middleware) {
```
Andarist: this one could have been checked before entering `runDeclarativeEffect` and `runDeclarativeEffect` should run only if this returns something other than `effect` itself

src/internal/runSaga.js
```diff
@@ -24,7 +25,7 @@ export function runSaga(
     subscribe,
     dispatch,
     getState,
-    {sagaMonitor, logger},
+    {sagaMonitor, logger, middleware},
```
Andarist: also aint sure if I feel naming this a `middleware` might be confusing for people, maybe it could be something more specific, like `effectMiddleware`? OTOH crippling the API to just `declarativeEffects` would be kinda crippling, why one couldnt mock other things with this?

eloytoro: agree

src/internal/proc.js
eloytoro: will do my best to move it outside

src/internal/proc.js
eloytoro: what do you mean? this function is now taking care of all the declarative effects, if the middleware returns the same it should still handle the same logic.

The reason im doing this here and separating the declarative effects from the rest is that i dont want to pass the middleware arrays of effects or pending promises or internal stuff

eloytoro: Oh wait nvm I think I get the point. In case the effect gets swaped by something thats not declarative.

Will take care of this

eloytoro:
@Andarist take another look ;)

Andarist:
code itself looks good, but I would like @yelouafi's feedback first before merging

src/internal/proc.js
```diff
@@ -365,16 +365,29 @@ export default function proc(
...
-      : is.helper(effect)   ? runForkEffect(wrapHelper(effect), effectId, currCb)
-      : is.iterator(effect) ? resolveIterator(effect, effectId, name, currCb)
+        is.promise(effect)                                                ? resolvePromise(effect, currCb)
+      : is.helper(effect)                                                 ? runForkEffect(wrapHelper(effect), effectId, currCb)
+      : is.iterator(effect)                                               ? resolveIterator(effect, effectId, name, currCb)
 
       // declarative effects
-      : is.array(effect)    ? runParallelEffect(effect, effectId, currCb)
-      :                       runDeclarativeEffect(effect, effectId, currCb)
+      : is.array(effect)                                                  ? runParallelEffect(effect, effectId, currCb)
```
yelouafi: why array effects are excluded from interpretation?

Andarist:
I guess its because each of them will get evaluated by runEffect anyway

yelouafi:
@Andarist @eloytoro  if we dont interpret combinators then `race` should be excluded as well?

Also I think the signature doesn't allow resolving effects asynchronously, I know this is mainly for testing purpose, but in integration tests it could be useful to test race conditions.

Another thing is I think this PR is closely related to #586 (Effect interceptors proposal). And we could perhaps think of a more generic API which allows chaining interpreters. What follows is just a braindump and would like to have your feedback.

What if we make a redux-like middleware api:

```js
runSaga(mySaga(), {
  ...,
  middleware: (next) => (effect, cb) => {
    if (isEqual(effect, call(serverRequest)) {
      cb('mocked value')
    }
    // forward to the next interpreter
    next(effect);
  }
});
```

multiple middlewares could be chained together, and also things like logging, monitoring, custom error handling could [possibly] be implemented as simple middleware.

The downside of using a callback is that we can't capture the moment after the effect execution. Perhaps using generators as middleware (koa style) can solve this

eloytoro:
@yelouafi I made some progress doing what youre asking using `co` and other fancy middleware composition stuff (inspired off koa like you said) but I realized that this would force an extra tick into every effect call because `co` is always async. So the async middleware promise produced by `co` has to be resolved on every call to `runEffect` because we must check if the effect changed after the middleware went through, is that ok?

Andarist:
> @Andarist @eloytoro if we dont interpret combinators then race should be excluded as well?

Hm, not quite - whole `race` could be resolved with this (that would require automatic cancelation of the sub-effects) and in the same time a sub-effect would be able to resolve itself via this middleware stuff, so the `race` would be resolved automatically (because winner was found).

yelouafi:
> I made some progress doing what youre asking using co and other fancy middleware composition stuff

I think using a 3rd party lib like co would add much size overhead. Anyway, we can in a first time use the simple callback api in my previous example to get this PR merged (would be sufficient to enable integration testing). We can then have another PR to discuss further a generator based API

@Andarist If i'm not mistaken, the same argument applies to parallel effects, no?

Andarist:
Hm, it depends when Im looking at it :D but yeah, argument can be easily applied to the parallel effect too. Maybe we should have `parallel([...effects])` too and deprecate `yield`ing arrays?

Would be more explicit what it does and educating about it would be easier and also would be easier to feed such a middleware as from this PR with this.

yelouafi:
>Maybe we should have parallel([...effects]) too and deprecate yielding arrays?

Hmm.. dont know, seems like a 'significant change' (regarding the number of LOCs likely using the `yield [...]`  form in real code)

Andarist:
For sure yielding arrays wouldnt be removed at once, just deprecated at first. But it would give more explicitness for this effect, which is in my opinion better in most cases. Implicitness causes confusion.

yelouafi:
>But it would give more explicitness for this effect, which is in my opinion better in most cases. Implicitness causes confusion.

Fair point. Another  argument is an explicit effect could support object keys like `race`.

```js
  // or whatever name it gets
  yield all({
    prop1: call(...),
    prop2: take(...)
  })
```

Andarist:
hm, `all` is a good name! nice relation to the well-known Promise API, although I got used to `parallel` already 😢 

we made quite an offtopic here about this one, but what about the PR as the whole? Should something be done here or should we merge in?

yelouafi:
I think [the issue here](https://github.com/redux-saga/redux-saga/pull/729#issuecomment-270117116) is still unresolved.

eloytoro:
@yelouafi did some progress on this trying to stick to the API you proposed.

I'm not sure what the `cb` in the middleware signature you propose means, so I'm leaving it out.

You can now chain middlewares if you do this
```js
effectMiddleware: [middleware1, middleware2]
```

However it seems that delaying effects from triggering will break the lib for an unknown reason

```js
effectMiddleware: (next) => (effect) => {
  // breaks task.done promise
  setTimeout(() => next(effect), 100);
}
```

Andarist:
`cb` is a resolver, I guess in your implementation the effect needs to be interpreted by the whole chain and final `next` is the `cb` itself.

Proposed API was to pass `next(effect, cb)` to the next item in the chain (`cb` missing in original comment because of the mistake I guess). In that manner you could bypass leftovers in the chain if u know you can return value already right into the saga's interpreter.

Andarist:
@eloytoro the problem was that in your test by delaying the `forkB` effect it didnt get a chance to set up a listener for `ACTION-3` (with `take('ACTION-3')`) and it just missed it and couldn't proceed further because the action was dispatched too early

I've merged your branch here #1246 , need to review it and gonna just do some minor tweaks, so it can land on `master` soon. Sorry that it took so long time to take this further.

eloytoro:
It's actually my fault since I found that mocking depencies was a simpler
fix for my needs of mocking effects.

Thanks for the review!

PR #733, opened 2016-12-30
Add support for stringable action creators as a take pattern (closes #732)

### LINKED ISSUE #732: take() should work with action creators with a custom `.toString` property
`take()` should respect action creators that use `.toString()` to output their type—from libraries like *redux-actions* and *redux-act*—so we don't need to maintain a list of constants.

For anyone who hasn't seen it, `redux-actions` sets a `.toString()` method on it's action creators, thus eliminating the need for action type constants. With these libraries, you just use the action creator as a computed property key for an object and it automatically calls `.toString` and finds the type of the action on it's own. This reduces a lot of boilerplate code and makes for a cleaner-looking codebase (there's much less `SNAKE_CASED_CAPITALIZED_CONSTANTS` sprinkled around the code).

`take()` (or really, the `matcher` method in proc.js) should automatically call `.toString()` on these action creators. If this was implemented we could completely drop action type constant files (since we already don't need them for reducers with these action helper libraries).

I know that simply doing this for ALL patterns that are functions would break the predicate matcher. One way we could get around this is to see if the pattern being passed to `take()` has a "toString" property of it's own (not from prototype). This could be tested with `is.func(pattern) && pattern.hasOwnProperty('toString')` (or we could define a new helper to test for it). Action creators from these utility libraries pass this test while standard functions wouldn't.

IMO, nobody using these libraries wants to have to write a wrapper for `take()` and add _Yet Another Import™_ to the top of their files just to have this functionality. I also personally don't want to have to type `.toString()` after every action when I use `take()`.

I could make a PR to update the definition(s) of `matcher()` and/or `matchers()` from _src/internal/proc.js_ to have it support this.

thezanke:
Allows action creator functions from libraries like redux-actions and redux-act to be used as the pattern in `take(pattern)` by checking for and calling `toString()` on them. Also updates the array matcher to handle an array of these action creators. (see #732)


src/internal/proc.js
```diff
@@ -12,15 +12,17 @@ export const TASK_CANCEL = {toString() { return '@@redux-saga/TASK_CANCEL' }}
...
 }
 
 function matcher(pattern) {
   return (
-      pattern === '*'   ? matchers.wildcard
-    : is.array(pattern) ? matchers.array
-    : is.func(pattern)  ? matchers.predicate
+      pattern === '*'          ? matchers.wildcard
+    : is.array(pattern)        ? matchers.array
+    : is.stringableFn(pattern) ? matchers.stringable
```
Andarist: im wondering if we should restrict this to functions only, also as naming convention suggests (`is.func`) the name should be `is.stringableFunc`

src/internal/proc.js
```diff
@@ -12,15 +12,17 @@ export const TASK_CANCEL = {toString() { return '@@redux-saga/TASK_CANCEL' }}
 const matchers = {
   wildcard  : () => kTrue,
   default   : pattern => input => input.type === pattern,
-  array     : patterns => input => patterns.some(p => p === input.type),
+  stringable: pattern => input => input.type === pattern.toString(),
```
Andarist: this one could be completely avoidable with combo of
`default   : pattern => input => input.type === String(pattern),`
+
`: is.stringableFunc(pattern) ? matchers.default`

src/internal/utils.js
```diff
@@ -20,19 +20,20 @@ export function check(value, predicate, error) {
...
+  array       : Array.isArray,
+  promise     : p => p && is.func(p.then),
+  iterator    : it => it && is.func(it.next) && is.func(it.throw),
+  task        : t => t && t[TASK],
+  observable  : ob => ob && is.func(ob.subscribe),
+  buffer      : buf => buf && is.func(buf.isEmpty) && is.func(buf.take) && is.func(buf.put),
+  pattern     : pat => pat && ((typeof pat === 'string') || (typeof pat === 'symbol') || is.func(pat) || is.array(pat)),
+  channel     : ch => ch && is.func(ch.take) && is.func(ch.close),
+  helper      : it => it && it[HELPER],
+  stringableFn: f => is.func(f) && f.hasOwnProperty('toString')
```
Andarist: hm, often I see libs taking precautions and store first `{}.hasOwnProperty` so it can be called later as your code would fail on this
```javascript
function func() {}
func.toString= () => 'func'
func.hasOwnProperty= () => false
```

thezanke: Hello @Andarist! Thank you for taking the time to review this for me!

I have the other changes in place but I am confused by what you mean here. Could you elaborate? Are you saying it would choke if someone defined a custom `hasOwnProperty` method the input pattern so we should store it as a util?

Andarist: yeah, basically yes, I know that it is uncommon for people doing that etc and thats just being me too cautious, especially that people could always shoot themselves in the foot (i.e. `Object.prototype.hasOwnProperty` overriding), but when `lodash` do smth I feel confident there is a reason to that and maybe we should apply the same thing here:
[1](https://github.com/lodash/lodash/blob/ef618992b5c206fd07202d6234807085fe3b58d7/lodash.js#1453) [2](https://github.com/lodash/lodash/blob/ef618992b5c206fd07202d6234807085fe3b58d7/lodash.js#L3105)

src/internal/utils.js
```diff
@@ -19,6 +19,11 @@ export function check(value, predicate, error) {
   }
 }
 
+const hasOwnProperty = Object.prototype.hasOwnProperty
+export function hasProperty(object, property) {
```
thezanke: I went ahead and just added this to utils as an export so it can be reused if needed down the road. If you think this unnecessary I can change it.

It annoys me that I am using `is.notUndef` before the object `is` gets defined in the file. but if I put the helper _below_ the definition of `is`, it would be falling `hasProperty` above the definition of it. I hate moments like this.

utils.d.ts
```diff
@@ -24,6 +24,7 @@ export const is: {
   observable: Predicate<any>;
   buffer: Predicate<any>;
   pattern: Predicate<any>;
+  stringableFn: Predicate<any>;
```
Andarist: you've missed correcting the name of this one, should be `stringableFunc`

src/internal/utils.js
Andarist: dont worry about this ordering, circular dependencies are fun ;)

though I feel like `hasOwn` would be better name for this utility, from `hasProperty` somebody could read that it should return `true` also for inherited ones

src/internal/proc.js
```diff
@@ -11,16 +11,17 @@ export const TASK_CANCEL = {toString() { return '@@redux-saga/TASK_CANCEL' }}
...
 }
 
 function matcher(pattern) {
   return (
-      pattern === '*'   ? matchers.wildcard
-    : is.array(pattern) ? matchers.array
-    : is.func(pattern)  ? matchers.predicate
+      pattern === '*'            ? matchers.wildcard
+    : is.array(pattern)          ? matchers.array
+    : is.stringableFunc(pattern) ? matchers.default
```
Andarist: still wondering if this should be restricted to just stringable functions, although when I wonder about this:
```javascript
{
  type: 'TYPE_A',
  toString: () => 'TYPE_B',
}
```
I have cold chills, so maybe only functions after all, WDYT? do u know about any other libs with different approach to actions than basic one?

utils.d.ts
thezanke: Oh shoot, thanks, sorry about that!

src/internal/proc.js
thezanke: Hmmm, I don't know of any. Would this be something someone would put `take()` do you think? That wouldn't be a valid pattern either way, would it?

Andarist: ye, it would be now... im just wondering how broad use cases should we cover here and what other libs should be considered in this context

Andarist:
will let it stay in PRs for few days, if somebody has some comments before merging it in, but after that it will land into the `master`

thezanke:
Hey @Andarist, I'm just checking in to see if there's any movement on this so I can stop using my fork in my projects :D

Andarist:
Hehe, sorry for the delay :) Gonna merge after the weekend. However if I dont, please just ping me and I will do it.

thezanke:
"Good Monday Morning", @Andarist 😂

Andarist:
merged

cc @yelouafi - code review would still be much appreciated

PR #735, opened 2016-12-31
Added getContext/setContext effects, where context is acting like dyn…

### LINKED ISSUE #713: Connecting to both Redux and non-Redux store
I have some stuff that I can't put into Redux store because it is not serializable and not immutable. More precisely, a Connection object that is created and managed by a third-party lib. To avoid passing connection via Redux action, I use Channels. It works quite well, despite the need to pass channels to every saga that needs the connection. 

However, a lot of generators aren't started on application startup, so they don't have a chance to catch the connection from channel. To workaround this, I have to pass connection along with channel to lots of generators which is not convenient at all.

I have one idea how to solve this. It's a common practice to assign third-party clients to Redux store like this:
```js
const store = createStore(...);
store.thirdPartyClient = new ThirdPartyClient(...);
```
Then you're able to use the client in React components.

Saga runtime is connected to the store, although only to `getState` and `dispatch`. But if we connect runtime to the whole store, we could introduce custom effect creators like this:
```js
function getConnection() {
  return store => store.thirdPartyClient.connection;
}
```
And then use it in any generator without the need to pass connection as argument:
```js
const connection = yield getConnection();
```

Maybe there's a more elegant way, like introducing something similar to React Context.

Andarist:
…amic scoping between parent/child tasks (resolves #713)

Just a quick draft of this proposed API. I've put several `TODO:`s in the code, they act as thought-placeholders and I would appreciate commenting each one of them.

Also there I've spotted a difference between standard `middleware` API and `runSaga`.
- `middleware` - options are passed once, so encapsulated by the factory) and each subsequent `middleware.run(saga)` will receive them (also `context` now)
- `runSaga` - each call expects the `options` all over again, so subsequent calls do not share `context` and it cannot be shared among 'top sagas' (at least not without creating a `topContext` within `runSaga` module)

cc @aikoven


Andarist:
also I think like some way for removing the context should be available as for now it wouldnt be possible to 'uncover' higher scope
```javascript
function* saga() {
	yield setContext({ a: 1 })
	yield fork({
	  yield setContext({ a: 2 })
	  yield setContext({ a: undefined })
	  yield getContext('a') // undefined, no way to reach { a: 1 } at this point
	})
}
```
hating to put 3 effects for this rather simple feature - `getContext`, `setContext` and `unsetContext` (?)

src/internal/proc.js
```diff
@@ -626,6 +640,18 @@ export default function proc(
     channel.flush(cb)
   }
 
+  function runGetContextEffect(prop, cb) {
+    // TODO: if object returned -> return shallow copy?
+    // should yield getContext() be available to access context?
+    // if yes then shallow copying it is a must to avoid unexpected mutations
+    cb(taskContext[prop])
```
aikoven: Context property could be anything, not just plain object, so we can't safely copy it.

Andarist: eh, you static typing habits!

Andarist:
> Should generator have access to context properties set by itself? IMO it shouldn't, because you can just use variables for this. So maybe `setContext` should only have effect on children? Maybe even rename `setContext` with `setChildContext` (I don't really like the verbosity though)?

I dont think it will be a problem as this context feature is less convenient to use, so people will always prefer accessing variables than this within task which uses `setContext`

> For removing context we can make `setContext` replace all *own* properties instead of assigning. This way we can keep API surface smaller.

I think that could lead to unexpected bugs and also updating the `context` would be quite unwieldy

> Should `c` be able to uncover context set in its grandparent `a`? I think not, because it will break encapsulation. 

Nah, didnt think about uncovering like this, just 1 step into the upper `context` like in the example posted [here](https://github.com/redux-saga/redux-saga/pull/735#issuecomment-269887969)

aikoven:
> > For removing context we can make setContext replace all own properties instead of assigning. This way we can keep API surface smaller.
> 
> I think that could lead to unexpected bugs and also updating the context would be quite unwieldy

Another way to avoid `unsetContext` is to use symbols for context keys on the library user's side.

> > Context property could be anything, not just plain object, so we can't safely copy it.
>
> ofc it would be checked beforehand if it is a plain object to see if we should make a shallow copy

Is this really a library responsibility? We don't shallow copy `select`ed state and actions returned from `take`.

aikoven:
Here's a counterexample:
```js
const obj = {
  foo: null,
  setFoo(foo) {this.foo = foo},
  getFoo() {return this.foo}
};

yield setContext({obj});
// ...
const obj1 = yield getContext('obj');
yield call([obj1, obj1.setFoo], 1);
// ...
const obj2 = yield getContext('obj');
const foo = yield call([obj2, obj2.getFoo]);  // null
```
Here, the value of `obj.foo` is lost due to shallow copying. In real world, this could happen when you store observable objects in context and later add callbacks to them.

Andarist:
Hm, interesting. Its really a scary situation and compelling one at the same time to drop shallow copying. Wondering only which usage would be more often and therefore which approach would give us less footguns in the end.

In theory we could make 2 versions of this effect getContext and getContext.impure

In general im in favor of immutability and this-less programming, aint saying though that everyone should. This feature would be experimental anyway and rather an escape hatch which should noy be used in trivial cases anyway, so... i dont know which is better now as a default

aikoven:
> In general im in favor of immutability and this-less programming

I absolutely agree with you, but my reasoning is this: for pure things we already have Redux store and actions which we can access in sagas at any level using appropriate effects. We would use Context to access non-pure third-party library objects.

Another use case is to share Channels between sagas, however I guess shallow-copying a Channel would be safe.

src/internal/proc.js
```diff
@@ -481,7 +485,17 @@ export default function proc(
 
     try {
       suspend()
-      const task = proc(taskIterator, subscribe, dispatch, getState, options, effectId, fn.name, (detached ? null : noop))
+      const task = proc(
+        taskIterator,
+        subscribe,
+        dispatch,
+        getState,
+        (detached ? null : taskContext), // TODO: topContext should be passed in case of spawn?
```
yelouafi: why cant we just pass the parent contextN

Andarist: reasoning behind it - `spawn` creates a `detached` task, which is acting like a top task, so would it be good to propagate `taskContext` (parent context) to it? behaviour would differ from other propagations (error/cancellation) in terms of detached (`spawn`) and non-detached tasks (`fork`)

yelouafi: yeah but you could also view `spawn` as a simple child task with different behavior regarding cancellation/errors? And also with analogy to lexical scope, it gets its params from its parent task, so could be the same from dynamic scope (context)

Andarist: could be ofc, but what i have proposed felt more natural when i was implementing this, however that was just a question for reviewers (so mostly you :P) left in the source code

also a little bit connected is this - https://github.com/redux-saga/redux-saga/issues/763#issue-200765663

in regards of spawn/fork debate, cancel propagation and maybe even need to distinguish rootTask from regular tasks for code-splitting purposes and proper cancelation logic on route changes, would appreciate your input about those matters there greatly!

src/internal/io.js
```diff
@@ -153,6 +155,16 @@ export function flush(channel) {
   return effect(FLUSH, channel)
 }
 
+export function getContext(prop) {
+  check(prop, is.string, `getContext(prop): argument ${ prop } is not a string`)
+  return effect(GET_CONTEXT, prop) // TODO: maybe accept functions also?
+}
+
+export function setContext(props) {
+  // TODO: check if props is object/plain object?
```
aikoven: Yep, I think we should restrict `props` to be a plain object.

src/internal/middleware.js
```diff
@@ -91,5 +93,14 @@ export default function sagaMiddlewareFactory(options = {}) {
     return task
   }
 
+  sagaMiddleware.setContext = (props) => {
```
aikoven: What's the purpose of this API? I can see cases where you need to change the root context as a result of some event, but I think it's more convenient to do this in the root saga.

src/internal/middleware.js
```diff
@@ -91,5 +93,14 @@ export default function sagaMiddlewareFactory(options = {}) {
...
+    // TODO: or maybe `setContext` should be available on task and this one should act only as proxy method?
+    // would that be useful?
+    // const task = yield takeEvery(...)
+    // const task.setContext({...}) or maybe even - yield setContext(task, {...})?
```
aikoven: IMO the context of a task should be a private thing, just like `state` in React components. The component itself is an *owner* of its state and it shouldn't be accessible from its parent.

src/internal/proc.js
aikoven: Agree with @yelouafi, if we disable propagating context on `spawn`, it would make `context` a much less convenient mechanism for dependency injection.

That said, I think that `runSaga` should behave just like you said — starting with an empty context by default. Another reason is that otherwise we'd have to somehow link `runSaga` with `sagaMiddleware` instance.

src/internal/proc.js
aikoven: Ok, I tend to agree with you that throwing is not really necessary here as any relevant checks could be easily performed on the user side.

src/internal/proc.js
```diff
@@ -626,6 +640,18 @@ export default function proc(
...
 
+  function runGetContextEffect(prop, cb) {
+    // TODO: if object returned -> return shallow copy?
+    // should yield getContext() be available to access context?
+    // if yes then shallow copying it is a must to avoid unexpected mutations
+    cb(taskContext[prop])
+  }
+
+  function runSetContextEffect(props, cb) {
+    Object.assign(taskContext, props)
```
aikoven: A check for `props` being a plain object should be added here as well.

src/internal/middleware.js
aikoven: Just found out this was proposed by myself: https://github.com/redux-saga/redux-saga/issues/713#issuecomment-268750352

Although I think only one of `options.context` or `sagaMiddleware.setContext()` is enough.

Andarist: well, one is needed upon creation and latter might be used in a runtime, could be ofc in response for some action but for more complex contexts it would require action to contain ur dependency which could be not serializable fully, which i dont like - if you for example need to pass some event bus with methods on it

src/internal/middleware.js
Andarist: actually as React's `setState` is available on the instance of the component - it could be used outside and therefore its not so private

src/internal/proc.js
Andarist: such checks are usually performed in effect creators, it would only double the same check if we put it here also, as its not possible to reach this - `runSetContextEffect` in any other way that `yield setContext(...)`

src/internal/io.js
```diff
@@ -153,6 +155,16 @@ export function flush(channel) {
   return effect(FLUSH, channel)
 }
 
+export function getContext(prop) {
+  check(prop, is.string, `getContext(prop): argument ${ prop } is not a string`)
+  return effect(GET_CONTEXT, prop) // TODO: maybe accept functions also?
```
Andarist: cc @aikoven

src/internal/middleware.js
aikoven: Good point. 👍

src/internal/io.js
aikoven: Since we do `Object.assign` and not deep merge, I'd view `context` as a flat object. Thus not sure if complex accessors are needed. Maybe I'm missing some use cases, but I'd rather add such features on demand.

src/internal/middleware.js
aikoven: Yep, also I can see some parallel with `sagaMiddleware.setContext`. Maybe then we just leave `task.setContext` and remove `sagaMiddleware.setContext`, as the latter can be done with `sagaMiddleware.run(...).setContext`?

Andarist: hm... at first I thought that this is a valid point and should be done, but after a while i think in the world of code-splitting its nice to have a possibility to share this top-level middleware context among root sagas, as there could be many of those, and when puting this only on task itself it would require such verbose code in order to share it 'globally':
```javascript
const context = {...}
sagaMiddleware.run(...).setContext(context)
// ... later
sagaMiddleware.run(...).setContext(context)
```

aikoven: You're right, I've overlooked the case when `sagaMiddleware.run` is called multiple times.

Andarist:
@aikoven 
Last comments made I think (ofc still waiting for your input, if you disagree - or agree :trollface: - somewhere) and I think im gonna update the PR soon so it can get finally merged in

src/internal/middleware.js
Andarist: the only thing im not so happy about is that `sagaMiddleware` is kinda superior to the `runSaga` here, as the latter do not have possibility to share such things in an elegant way, but I guess one can always write some wrapper function up to his needs

Andarist:
I think the code is final now, so you can have last look on it - whats missing before merging is some tests + docs, ill get them done in this week and merge whole thing to the master.

src/internal/middleware.js
```diff
@@ -1,9 +1,9 @@
-import { noop, is, check, uid as nextSagaId, wrapSagaDispatch, isDev, log } from './utils'
+import { noop, is, check, uid as nextSagaId, wrapSagaDispatch, isDev, log, object, createSetContextWarning } from './utils'
 import proc from './proc'
 import { emitter } from './channel'
 import { ident } from './utils'
 
-export default function sagaMiddlewareFactory(options = {}) {
+export default function sagaMiddlewareFactory({ context = {}, ...options }) {
```
aikoven: Won't this break if we call factory without arguments?

Andarist: it would, thanks for noticing! gonna fix it with default param in a minute

AriaMinaei:
Guys, do you think this branch is going to get merged eventually?

Andarist:
@AriaMinaei yes, it sure will! I just need to get some time to merge it and some other branches, so I can release new version, sorry about the delay

neurosnap:
Is there anything I can do to help out with this PR?

Andarist:
@neurosnap, i think its basically done. Will review it again in the weekend and gonna publish

Andarist:
Released, not docs yet - but if you know about the feature and want to use it - go ahead. Let me know if everything works as expected.

PR #739, opened 2017-01-03
Fixed surfaced API from join([tasks]) to join(...tasks)

Andarist:
Aint sure which one should be surfaced, so this PR might get rejectected with no regret, but I feel I've wanted the API proposed by this PR to be exposed and made it otherwise by a mistake.


yelouafi:
probably `join(...task)` is more flexible as it can handle all iterables, not just arrays. OTOH `join(Array)` is more convenient when you have your tasks stored in a an array. But we could also get both feature by allowing a signature like `join(Iterable)`, like Promise.all for example

Andarist:
Hm, I think this is what it would take for allowing both APIs. WDYT?

Andarist:
it also seems reasonable to support `cancel(...tasks)`

yelouafi:
>it also seems reasonable to support cancel(...tasks)

I agree

src/internal/io.js
```diff
@@ -104,16 +104,19 @@ export function spawn(fn, ...args) {
 
 const isForkedTask = task => task[TASK]
 
-export function join(task) {
-  if (is.array(task)) {
-    return task.map(join)
+export function join(...tasks) {
+  if (tasks.length > 1) {
```
Andarist: @yelouafi 
what do you think about those checks, it feela a little bit clunky, but I think this is what it takes to support both forms of the APIs, but maybe we should just drop one, so the code stays cleaner, aint sure however which one - `join(...tasks)` or `join([...tasks])`

yelouafi: yeah, I think we should just keep join(...tasks)

Issue #608, opened 2016-10-30
TypeScript Cancelled Missing


JesterXL:
Still learning TypeScript casting, but I believe v0.12.1 didn't fix all the missing TypeScript types. I'm using Angular CLI 1.0.0-beta.17 with the `tsconfig.json` set to ES6 output so generators, and thus sagas, work.

<img width="1205" alt="screen shot 2016-10-30 at 12 03 23 pm" src="https://cloud.githubusercontent.com/assets/424737/19837854/8a95c1e4-9e99-11e6-91cf-886bb35dd118.png">

Looking at where error might potentially be, the `effects.d.ts` looks correct, however, and you can see the interface for `CancelledEffect` has a cancelled property, so... confused. @jhsu Do you know what it could potentially be?

JesterXL:
Temporary workaround to get it compiling:

```typescript
const myThing:any = rootSaga;
const rootSagaResult = sagaMiddleware.run(myThing);	
```

aikoven:
Can you please post the code fragment where it fails?

JesterXL:
Yup:

```typescript
export class AppModule
{
	constructor(ngRedux:NgRedux<IAppState>)
	{
		ngRedux.configureStore(rootReducer, {}, [ createLogger(), sagaMiddleware]);
		const myThing:any = rootSaga;
		const rootSagaResult = sagaMiddleware.run(myThing);	
		rootSagaResult.done.catch((err)=>
		{
			console.log("Error in Sagas", err);
		});
	}
}
```
It's mad at the `sagaMiddleware.run`.

aikoven:
Seems to be a result of some of `yield`ed value types, although it's not enough to tell where the problem is exactly.
TypeScript is still not very good at generators, but at least it can check types of values you `yield`. My current approach is to set return type of every saga to `SagaIterator`:
```ts
import {SagaIterator} from "redux-saga";

function* someSaga(): SagaIterator {
  // ...
}
```
This way you always know that you only `yield` correct values i. e. effects.

JesterXL:
Nice! I'll try that, and once I get her 100% compiling without `any`'s scattered all over the place, I'll update this thread.

JesterXL:
Another problem is you can't use Enum's for Actions:

```typescript
enum Actions
{
	PING,
	PING_ERROR,
	PINGED,
	
	LOGIN,
	LOGGED_IN,
	LOGOUT,
	LOGIN_ERROR
}
```

[...]

Issue #619, opened 2016-11-08
TypeScript CallEffect signature


tvedtorama:
Hi, 

the current typescript definition for the call effect comes with max 4 arguments, and a fallback for any argument count.  

When I am working with this, the fallback is preventing type checking on arguments - and I had to remove it for it to actually use the T1,T2, etc types of the more specific signatures.  I was otherwise not stopped from passing the wrong argument list.

Can we get rid of the fallback signature and provide a five-argument version?

From this:
```
interface CallEffectFactory<R> {
  (fn: CallEffectArg<CallFunc0>): R;
  <T1>(fn: CallEffectArg<CallFunc1<T1>>,
       arg1: T1): R;
  <T1, T2>(fn: CallEffectArg<CallFunc2<T1, T2>>,
           arg1: T1, arg2: T2): R;
  <T1, T2, T3>(fn: CallEffectArg<CallFunc3<T1, T2, T3>>,
               arg1: T1, arg2: T2, arg3: T3): R;
  <T1, T2, T3, T4>(fn: CallEffectArg<CallFunc4<T1, T2, T3, T4>>,
                   arg1: T1, arg2: T2, arg3: T3, arg4: T4): R;
  (fn: CallEffectArg<CallFuncRest>, ...args: any[]): R;
}
```

To this:
```
interface CallEffectFactory<R> {
  (fn: CallEffectArg<CallFunc0>): R;
  <T1>(fn: CallEffectArg<CallFunc1<T1>>,
       arg1: T1): R;
  <T1, T2>(fn: CallEffectArg<CallFunc2<T1, T2>>,
           arg1: T1, arg2: T2): R;
  <T1, T2, T3>(fn: CallEffectArg<CallFunc3<T1, T2, T3>>,
               arg1: T1, arg2: T2, arg3: T3): R;
  <T1, T2, T3, T4>(fn: CallEffectArg<CallFunc4<T1, T2, T3, T4>>,
                   arg1: T1, arg2: T2, arg3: T3, arg4: T4): R;
  <T1, T2, T3, T4, T5>(fn: CallEffectArg<CallFunc5<T1, T2, T3, T4, T5>>,
                   arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5): R;
}

```

?

aikoven:
What if you had to pass six arguments?
Actually the right way would be to have something like this:
```ts
(fn: CallEffectArg<CallFuncRest>, 
[...]

PR #740, opened 2017-01-03
Update TypeScript definitions

aikoven:
**BREAKING CHANGE**
This PR can potentially break builds for TS users.

* Stronger typings for `call`, `fork` etc effects. Passed argument types are now checked correctly.
* Add tests.
* Add missing and remove deprecated APIs.
* Export effect descriptor types to aid testing.
* Includes some of not yet merged APIs: ~~https://github.com/redux-saga/redux-saga/pull/733~~, https://github.com/redux-saga/redux-saga/pull/739
* Helpers (`takeEvery`, `takeLatest`, `throttle`) now accept up to 6 arguments. `...rest` arguments removed for the sake of correctly checking last `action` argument. `cps` effect creator accepts up to 4 arguments.


aikoven:
Resolves https://github.com/redux-saga/redux-saga/issues/608 and https://github.com/redux-saga/redux-saga/issues/619.

Andarist:
> remove deprecated APIs

I think they should be kept while deprecated and dropped accordingly when the features are removed from the core

> Helpers (takeEvery, takeLatest, throttle) now accept up to 6 arguments. ...rest arguments removed for the sake of correctly checking last action argument. cps effect creator accepts up to 4 arguments.

wont it be limiting? aint possible making those of completely variadic length?

aikoven:
> wont it be limiting? aint possible making those of completely variadic length?

Latest TS only supports vararg at the end of arguments list, so it's impossible to strongly annotate `(...args, action)`. I think I'll return the `...rest` version, but `action` argument won't be type-checked.

CarsonF:
@aikoven Can you add `emitter` to the `Monitor` interface?

nbostrom:
These typing doesn't actually allow the iterator to use `delay` since it returns a Promise in place of an Effect. I think that we need to update the SagaIterator definition to something like:
```
export type SagaIterator = IterableIterator<Effect|Effect[]|Promise<any>>;
```
Or update the Effect def if redux-saga counts Promises as Effecs.
I'm not very good at typescript yet, so there's probably a better way. ;)

aikoven:
@nbostrom IMO `yield`ing promises is an anti-pattern, since it means you create a side-effect in the generator itself instead of saga runtime. You can use `yield call(delay, 50)` instead, which is pure.

nbostrom:
@aikoven Oh, I see. That makes a lot of sense. Thanks!

Andarist:
@aikoven whats the status of the PR? do you know what things does it include? should I cross check against what might be missing and what potentially could be updated? or are you on master's toes? ;)

would like to proceed with this PR as im preparing a new release (no rush though!), if I can help in any way, just let me know

aikoven:
@Andarist There are a couple things that should be done:
1. [`yield cancel(...tasks)`](https://github.com/redux-saga/redux-saga/pull/800). I'm confused about whether it is in `master` already.
2. `onError` option for `runSaga`
3. Possibly a solution for https://github.com/redux-saga/redux-saga/pull/770

aikoven:
Updated the PR. Added some missing APIs and loosened restriction on generators passed to `SagaMiddleware.run` as well as iterators passed to `runSaga`. Strict type-check of `yield`ed values is still possible by annotating return types of generators with `SagaIterator`.

Andarist:
Thanks for updating the PR! Im trying to push v0.15 forward and I think I wont release anything in between. Some things ere not yet merged in, but I think their typings could be provided in this PR, so we do not have to come back to the matter latter:

- `runSaga` refactor - basically just 1. args reordering (store interface geting in there first) 2. it accepting a generator instead of an iterator instance 3. variadic args at the end to be passed into the generator internally. That means that `sagaMiddleware.run` becomes just partially applied `runSaga` - dont know if that helps you to dedup something
- self cancellation - `cancel(task)` and `cancel()` becomes valid, but not `cancel(undefined)`!
- `all` effect - just an explicit alternative for `yield [...effects]`
- context features from #735 - so `getContext`, `setContext` effects and things like `sagaMiddleware.setContext` and `{ options: { context } }` passed in upon creation - although maybe i would hold with that until the PR is more stabilized. There are some unanswered comments in the added code, maybe you could look into them?

aikoven:
Do you expect all of these to be merged for `v0.15`? Otherwise I could just make another PR.

Andarist:
Yeah, I would like to include all of them in v0.15

Andarist:
cc @aikoven #823

aikoven:
- [ ] New `runSaga` signature
- [ ] Self cancellation via `yield cancel()`
- [ ] `all` effect
- [ ] Context feature

aikoven:
- [ ] Support for passing method names to `call` etc. https://github.com/redux-saga/redux-saga/pull/826

dcolthorp:
+💯 Many of the current definitions are close to useless, as they provide no safety.

Any news as to whether this will be merged and/or released? It would be nice to at least pin to a commit in the `redux-saga` instead of a fork.

Andarist:
@dcolthorp 
Yes, this is certianly going to be merged. Cannot give exact estimate though, trying to push things forward as quick as I can.

Is this PR solving ur problem?

dcolthorp:
@Andarist Yeah, they're exactly what I was hoping for. One of the main reasons to use TypeScript to begin with is safety, and the currently released type definitions are useless there, because any invalid arguments to e.g. `call` fall back on an alternate definition that completely gives up type safety.

The currently released type definitions are problematic enough that I was questioning the viability of using sagas in my TypeScript project at all until I found this PR. I suspect getting this merged would be of high priority to anyone using sagas in typescript.

Andarist:
I'll try to review what needs to be done, so this can get merged in.

Andarist:
Merged in, preparing a release. Stay tuned. ;)

dcolthorp:
Excellent, thank you @Andarist and @aikoven for your work!

Andarist:
@aikoven needs some loud 👏 👏 here, ive just merged it in ;)

PR #741, opened 2017-01-03
Added support for easily making channels out of observables

Andarist:
The 'problem' is that simple
```javascript
const chan = observableChannel(Observable.from([1, 2, 3]))
yield take(chan)
```
will result in `END`, as `subscribe` happens first, before taking, but I guess its just the nature of observables and there is not much to do about this without buffering and it is hard to tell upfront what buffers should be used by default.

Ideas:
1. wondering if we could handle yielding observables (maybe wrapping them automatically with `observableChannel`? and then we could treat `call` results the same same when observable
2. also maybe we could implement `$$observable` on the middleware so `put` actions can be observed?

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,9 @@
     "url": "https://github.com/redux-saga/redux-saga/issues"
   },
   "homepage": "https://github.com/redux-saga/redux-saga#readme",
-  "dependencies": {},
+  "dependencies": {
+    "symbol-observable": "^1.0.4"
+  },
   "devDependencies": {
     "babel-cli": "^6.1.18",
     "babel-core": "^6.14.0",
@@ -68,6 +70,7 @@
     "redux": "^3.5.1",
     "redux-logger": "^2.6.1",
     "rimraf": "^2.4.3",
+    "rxjs": "^5.0.2",
     "tap-spec": "^4.1.1",
     "tape": "^4.2.2",
     "typescript": "^1.8.10",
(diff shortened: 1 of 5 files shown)


yelouafi:
We could delay subscription until the first take (using a custom channel instead of eventChannel). I think this makes more sense since most of the observables are lazy (cold). And for the case of eager (hot) observables I dont think it makes a difference (unless we introduce buffering, but I dont think we need to support buffering for those channels since observables have more powerful support for that)

Andarist:
Hm, delaying the subscription is certainly a neat idea, only cancelation, unsubscription etc will be a little bit more trickier, shouldnt be a problem though. Gonna attemp to implement this.

In most cases such event channels gonna be created and immediately followed by a `take`, so it in fact doesnt really matter, as `take` will have its opportunity to set up, therefore subscribe and nothing will be missed as nothing can happen in the same stack.

As to buffering - we have those primitives already so it is reasonable to allow passing them in I think.

For a brief moment I thought it would be neat to propagate the item to every registered taker, but after a thought it would only cause confusion and also initial `take` (the one causing subscription) would be problematic.

> also maybe we could implement $$observable on the middleware so put actions can be observed?

WDYT about this idea?

yelouafi:
>also maybe we could implement $$observable on the middleware so put actions can be observed?

would be useful

z0al:
> We could delay subscription until the first take (using a custom channel instead of eventChannel)

Sorry to comment on an old PR but I'm facing the same problem (also using `eventChannel`). @Andarist did you figure out something or have some tips to share?

z0al:
Umm, passing `buffers.expanding()` to `eventChannel` seems to fix the issue. Not sure if I missed anything.

PR #743, opened 2017-01-04
Helper functions should return Effect instead of SagaIterator

yenshih:
And it would be nice if we can change the version to 0.14.2 :)


Andarist:
There is already pending [PR with typings](https://github.com/redux-saga/redux-saga/pull/740). Could you confront how does your PR relate to the other one? Are they overlapping?

cc @aikoven

yenshih:
I've seen his PR, it's definitely better than mine. A great job :)

PR #748, opened 2017-01-08
docs(translate): Add Korean translation

JisuPark:
Hello all,

I've added Korean translation file `README_ko.md` and I've updated translation section too.
If I missed something, please feel free to tell me.

I'll keep working on other docs too. 

Thanks,
Jisu

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ sagaMiddleware.run(mySaga)
 - [Chinese](https://github.com/superRaytin/redux-saga-in-chinese)
 - [Chinese Traditional](https://github.com/neighborhood999/redux-saga)
 - [Japanese](https://github.com/redux-saga/redux-saga/blob/master/README_ja.md)
+- [Korean](https://github.com/redux-saga/redux-saga/blob/master/README_ko.md)
 
 # Using umd build in the browser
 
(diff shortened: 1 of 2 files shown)


Andarist:
As those translations tend to outdate themselves when we work on the core itself and are able to update only the english one, I think it would be good to add a note at the top about last update + version of the saga at the moment of the update. Exact same note should be added in other translations too, I'll try to confront them with git history and add it appropriately when i have time.

JisuPark:
@Andarist Thank you for the feedback. I agree with that. So I added NOTE section at the top. It will notes last version with exact commit history.

README.md
```diff
@@ -124,6 +124,7 @@ sagaMiddleware.run(mySaga)
 - [Chinese](https://github.com/superRaytin/redux-saga-in-chinese)
 - [Chinese Traditional](https://github.com/neighborhood999/redux-saga)
 - [Japanese](https://github.com/redux-saga/redux-saga/blob/master/README_ja.md)
+- [Korean](https://github.com/JisuPark/redux-saga/blob/master/README_ko.md)
```
Andarist: it should point to the `https://github.com/redux-saga/redux-saga/blob/master/README_ko.md`

JisuPark: Thanks. I've changed it.

PR #749, opened 2017-01-09
Fix the link to unpkg in README

clbn:


README.md
```diff
@@ -130,7 +130,7 @@ sagaMiddleware.run(mySaga)
 
 There is also a **umd** build of `redux-saga` available in the `dist/` folder. When using the umd build `redux-saga` is available as `ReduxSaga` in the window object.
 
-The umd version is useful if you don't use Webpack or Browserify. You can access it directly from [unpkg](unpkg.com).
+The umd version is useful if you don't use Webpack or Browserify. You can access it directly from [unpkg](https://unpkg.com/).
```
Andarist: maybe it should be a direct link to the - https://unpkg.com/redux-saga ?

clbn: I think, as a reader, I would expect it to be the link to unpkg service, especially when links to the specific builds are available below. But I don't know if it really was the intention of the sentence's original author. :)

Andarist: you are right, didnt spotted that by looking on the diff alone

PR #750, opened 2017-01-09
Removing significant whitespace

TehShrike:
The space between the brackets and the link are breaking the CDNJS badge when the markdown is rendered at https://redux-saga.github.io/redux-saga/


Andarist:
Wow, thanks! Spotted this yesterday but didnt have time to find the reason behind this.

PR #751, opened 2017-01-09
Updated docs with RollupJS troubleshooting section

zech:
As requested by @Andarist here https://github.com/redux-saga/redux-saga/issues/689#issuecomment-269318658 this is the PR documenting the issue. ;-)


Andarist:
Merged! Thanks :)

PR #752, opened 2017-01-09
Expose onError in runSaga

blocka:
This allows `onError` to be used when using redux-saga outside of react.

### DIFF
--- a/src/internal/runSaga.js
+++ b/src/internal/runSaga.js
@@ -8,7 +8,8 @@ export function runSaga(
     dispatch,
     getState,
     sagaMonitor,
-    logger
+    logger,
+    onError
   }
 ) {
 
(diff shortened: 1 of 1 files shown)


Andarist:
Thanks! gonna merge right away

Side note - when i look on both exposed APIs (`sagaMiddleware.run` and `runSaga`) they are 90% the same and I'm gonna prepare a PR with their unification. That however will be probably a breaking change for `runSaga` (I'll try to avoid it, but it might be impossible).

Would you consider such unification beneficial, even if its gonna be a breaking change?

blocka:
I think unification is a worthy goal. I actually had expected that and had to dig through the code to realize why `onError` was not working (I realize it's not in the docs...but it was worth a shot anyway, and off hand seemed reasonable).

Andarist:
The refactor is available [here](https://github.com/redux-saga/redux-saga/pull/755/files#diff-9d82e6f7b43f79dd646cfe01cb37a7e0). If you have time please take a look. I've managed to avoid breaking change here, only mere deprecation warning for now.

PR #755, opened 2017-01-10
Run saga refactor

Andarist:
Didnt want to run into conflicts, so this one includes also small PR #754. But the other one should be merged much quicker into the `master`, so its commit will vanish from this one.

My reasoning is that this PR deduplicates some logic and treats `runSaga` as a building block upon which `sagaMiddleware.run` is built. Where before that those 2 were really similar, but duplicated and often change in one was forgotten to be added to the other one.


Andarist:
cc @blocka @slorber @zoff @HansDP @sergej-s @DEllement @pavelkornev @DjebbZ @aft-luke 

I've tried to mention every1 I've found on the issues mentioning theit potential usage of `runSaga` API, sorry if I've missed anyone

If any of you have any concern about this, please raise it :)

sergej-s:
@Andarist Thx for notification. I don't get how api of runSaga will be changed? Will I need to change my code in some way?

Andarist:
@sergej-s 
for now you wont need to change anything, but you will get a deprecation warning when using old form of this old API

changes required to get rid of them:

old API
`runSaga(saga(arg1, arg2), {dispatch, getState, subscribe})`
new API
`runSaga(saga, {dispatch, getState, subscribe}, arg1, arg2)`
@yelouafi 
in the sense with my PR [runSagaDynamically](https://github.com/redux-saga/redux-saga/pull/755/files#diff-56ae1a53433b6c6fbff8fe83311f4a94R52) is becoming a partially applied `runSaga`, so it could be made more obvious here and `partial` utility could be written and used here to prefill `runSaga`'s 2nd parameter - store interface

slorber:
as long as the feature is not removed I'm not against an api change if you need it

Andarist:
No intention to remove it at all, if its useful for non-redux projects (and i know it is ;) ) it shouldnt be limited to those only!

yelouafi:
>in the sense with my PR runSagaDynamically is becoming a partially applied runSaga, so it could be made more obvious here and partial utility could be written and used here to prefill runSaga's 2nd parameter - store interface

Wouldn't be better to flip the 1st and 2nd param

```js
runSaga({dispatch, getState, subscribe}, saga, arg1, arg2)
```

The argument behind this is that one will likely setup options only once and run sagas more often

Andarist:
Could be done, would ease currying this for sure. I'll update this later (and merge!) with flipping args in `runSaga`, so that both arguments orderings are supported for a time being (old one being deprecated ofc).

The whole thing would also allow us to get rid of the `sagaMiddleware.run` / `runSaga` doubled tests. Probably we could drop nearly all `runSaga` tests, as it will be only a generalization of `sagaMiddleware.run`, so everything should work the same way for both of the APIs.

Andarist:
I've updated the PR, didnt test it properly yet though. Was wondering if maybe a test on a living project would be possible, if anyone has time for it? cc @slorber

Issue #414, opened 2016-06-28
Custom effects


christian-schulze:
Is it possible/recommended to create custom effects?

My team was throwing around the idea, and on looking at the existing effects in `src/internal/io.js` it looks like this wasn't really intended.

What do you think?

MattKunze:
I had similar thoughts a while back, but I haven't been able to come up with any concrete use cases for it yet. It seems like it might be useful as a dependency injection mechanism to isolate sagas from runtime instances that aren't necessarily available through the store.

But I have [yet] come up with a scenario that isn't handled through `call`, it would be interesting to hear what other's have come across.

yelouafi:
I considered this idea when starting the lib. But as @MattKunze didn't think of use cases for it.

IMO it may be also more confusing. The semantics of a few effects (call, put, ...) if well understood make the code immediately understandable; With custom effects however, this may create another level of inderiction b/c the reader has to also understand the semantics of the custom operator (which may denote a call to a function)

christian-schulze:
Thank you for your responses @MattKunze and @yelouafi.

For our specific use case, we have a function called `followLink`, which given a url, will expect a well defined JSON document in return. The JSON document has an embedded status, which `followLink` will use to retry (with exponential back-off and max retries) until either `error` or `ready` is returned.

We use `followLink` extensively in most of our Saga's hence why we were enquiring about custom effects. I guess in this scenario, `followLink` is a more specialised form of `call`.

yelouafi:
>I guess in this scenario, followLink is a more specialised form of call.

Exact. You can create an effect creator which builds the specialized call effect.

For example

```js

// Saga to handle the effect logic
function* followLinkSaga(url) {
   yield call(fetch, url)
   //... retry logic
}

// Effect creator
function followLink(url) {
  return call(followLinkSaga, url)
}

// Use it like other effects
function* saga() {
   const {url} = yield take(ACTION)
   yield followLink(url)
}
```

christian-schulze:
This is excellent advice, and a very simple refactor to existing code. Will take this idea for a spin tomorrow, but please consider this issue solved 👍

PR #759, opened 2017-01-12
Add settle effect combinator

pcgilday:
This effect combinator will run effects in parallel and settle all effects before returning.

This is something that would be very useful for us, and I believe could be useful for others as well. When calling effects to fetch data in parallel, you may not want one failure to cause the other effects to be cancelled.

I was not sure what the desired response object should look like, so I left it simple and am open to changing it if there is something that would be better or more useful. While this is similar to [promise settling](http://bluebirdjs.com/docs/api/reflect.html), using the [PromiseInspection](http://bluebirdjs.com/docs/api/promiseinspection.html) interface didn't seem appropriate.

If you agree this is a useful effect combinator, I will write the docs for it. I didn't want to spend that time until this proposal was accepted and the response object format was agreed upon.

BTW - This is a great library, thank you for all your hard work!

cc - @spaintrain


SpainTrain:
Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/redux-saga/redux-saga/759)*

Andarist:
so its basically a `parallelEffect` (yielding array of effects) but each `effect` is wrapped in a pair of `{ result, isError }`?

That should be quite easily implemented on the user side. It sounds for me as too specific for being added to the lib just like that, especially when `settle` does not tell you much about its behaviour until you read about it.

Although I think its a great ground for a discussion about:
- custom effects - would be really cool if we support those, any ideas for the API?
- error handling, cancellation - what do you feel is lacking and what could be done better? should behaviours be more configurable somehow?

pcgilday:
> so its basically a parallelEffect (yielding array of effects) but each effect is wrapped in a pair of { result, isError }?

Somewhat, the major benefit is making sure all the effects running in parallel complete, whereas with `parallelEffect`, if one effect throws an error, the rest are cancelled. For example, if I have the below call (taken from docs, but we have something similar), and one of those effects throws an error, any of the effects still pending will be cancelled, when using `parallelEffect`. When we call something like `fetchCustomers`, a complete action is dispatched after the data is fetched so that the data is reduced into state. If `fetchProducts` throws an error, then `fetchCustomers` is cancelled and the customers are never reduced into state, so instead of missing one chunk of data, we are now missing two. 

```js
const [customers, products] = yield [
    call(fetchCustomers),
    call(fetchProducts),
]
```

If I use this settle effect combinator, I can still handle an error that may occur, but the rest of the effects are able to complete first, leaving state in a better place. This is just one example, but I believe having a way to run effects in parallel without having a single error cancel the rest of the effects running would be desirable. 

I think this is a useful effect combinator to have in this lib, promise libraries like [q use something similar](https://github.com/kriskowal/q/wiki/API-Reference#promiseallsettled) (not that someone else doing it makes it right, just a note). As far as settle not telling much about the behavior, I would be open to name suggestions, but that seemed the best choice as it is a [known](http://bluebirdjs.com/docs/api/reflect.html) [term](https://www.npmjs.com/package/promise-settle) used when dealing with promises, and seems to be used for internal variables representing the same concept within redux-saga [here](https://github.com/redux-saga/redux-saga/blob/master/src/internal/proc.js#L309)

> custom effects - would be really cool if we support those, any ideas for the API?

Agree, we have used the approach mentioned [here](https://github.com/redux-saga/redux-saga/issues/414#issuecomment-229653161), but having an API for custom effects would be nice. As far as a suggestion for that, I would have to think on it some.

> error handling, cancellation - what do you feel is lacking and what could be done better? should behaviours be more configurable somehow?

I believe this was covered by my response above, but if not I can expand.

aikoven:
I think that more generic way is to have the same stuff but for single effect, something like this:
```js
function safeCall(fn, ...args) {
  return call(function* () {
    try {
      return {result: yield call(fn, ...args)};
    } catch (error) {
      return {error};
    }
  });
}
```
This way you can wrap any call that could potentially throw, without needing to `try`/`catch`. And you can achieve what you want with:
```js
yield [
  safeCall(fetchCustomers),
  safeCall(fetchProducts),
]
```

pcgilday:
That was an option @SpainTrain and I discussed before I opened this PR. We thought it would be more elegant to use an effect combinator like `settle`, instead of having to wrap each call with a new effect. With the `safeCall` approach, if you used an effect other than `call`, you would need `safePut`, `safeCps`, etc.

But I think that is the decision to be made, the `safeCall` approach or the `settle` approach, unless there is a better way to accomplish this that we're missing.

Andarist:
Like @pcgilday said the proposed `settle` works for array of any effects. Could be achieved probably with something like this:
```javascript
function* settle(effects) {
  return effects.map(effect => function* () {
  	try {
  		return { result: yield effect }
  	} catch (err) {
  		return { err }
  	}
  })
}
```
Didnt test it though, so keep that in mind.

Your promise analogy is quite compelling. However still I would like to put it on hold, so we have time to take a stand on it.

pcgilday:
Just wanted to ping and see what the current thoughts or status for this PR are. I'll still need to add documentation if we think it will be merged, so if that would help further the discussion I would be happy to get that done in the next couple days.

Andarist:
I would like to have @yelouafi's opinion on this before proceeding, cant settle (:P) my mind on this firmly.

SpainTrain:
ping a doodle doo :-)

yelouafi:
I dont mind adding this in principle. But, like @aikoven, I suspect there could be a more basic solution behind this. For example if I have some effect `catch` (or `wait` if it'd return also cancelled outcomes) that behaves like that would act like `safeCall` but for all effects. Then we could implement `settle` on top of it as a helper effect.

Andarist:
cant rly see how the proposed effect of `catch` or `wait` would work, something like this doesnt make sense to me 
```javascript
yield catch(errHandler)
yield call(someAPIcall)
```
and cannot see how would you like to structure code in other way with `catch` effect

yelouafi:
`catch` would trap errors and return an explicit object `{result, isError}`.

yelouafi:
Example, suppose this code throws some `NetworkError` object

```js
try {
  const res = yield call(fetch, data)
} catch (err) {
  ...
}
```

Using catch, instead of throwing in the generator we return an explicit result
```js
  const {result, isError} = yield call(fetch, data)
```

Andarist:
in the decorator/higher-order-effect fashion? `yield catch(call(someAPI))`? 

then in fact `settle` would become something like `yield effects.map(catch)`

yelouafi:
>then in fact settle would become something like yield effects.map(catch)

Exact. We could also make the effect trap all completion values not just errors (cancellations, channel ending)

yelouafi:
In fact, if we use the method version `yield effect(...).catch()`. We could implement it just by setting a `catch: true` flag on the effect, then each effect runner would handle the flag by decorating the result appropriately  (like setting isError to true) and resuming the generator normally

Andarist:
Hm, aint sure though if that would be helpful to people and also what I as a user would expect from it. Cause then we would burden users with the need to handle all 'completion'/errors types. 

On one hand it might be useful in some rare situations, but in the most common scenarios catching errors is all what there is needed. 

We could introduce something like `catch` for just error traping and `catch.all` for traping all completion values.

Andarist:
hah! we are brainstorming too fast and our posts are becoming overlapping ;) liking it! you need to come here more often :P

> In fact, if we use the method version yield effect(...).catch(). We could implement it just by setting a catch: true flag on the effect, then each effect runner would handle the flag by decorating the result appropriately (like setting isError to true) and resuming the generator normally

this would require all effects to expose `.catch()` method. Which ofc could be easily done [here](https://github.com/redux-saga/redux-saga/blob/c8d027e7c4dda8c6e8ca6dced71470ae1a8ac669/src/internal/io.js#L23), but we'd lose serialization of the effects then. I would argue that `catch(effect)` is the better option here, working internally just the same - setting flag on the effect

yelouafi:
>Hm, aint sure though if that would be helpful to people and also what I as a user would expect from it. 

For blocking effects, there is no need since there no way to cancel manually for example a call effect from outside. So the user is not required to check other values (eg isCancel). For take effects, it'd play the same rule as takem effect (trapping channel ends).

In effects like `join(task)`. If the task is cancelled then the join effect will propagate cancellation to the calling saga. With catch, we could avoid that.

yelouafi:
>this would require all effects to expose .catch() method. Which ofc could be easily done here, but we'd lose serialization of the effects then

How so? if we create Effects based on a custom prototype (or a simple class) I think it'll be ok

Andarist:
For example simple objects (POJOs 😆) can be transferred through `postMessage` interface, but ofc function references cant be transferred between such contexts.

yelouafi:
>For example simple objects (POJOs 😆) can be transferred through postMessage interface, but ofc function references cant be transferred between such contexts.

If I'm not mistaken, serialization doesn't include fields from base prototypes, only own fields

yelouafi:
In fact, serialization even excludes own field functions

Andarist:
Oh yeah, my thought was only about when `.catch` being a method the effect wouldnt be able to serialize 'fully'. With `cancel` being a decorator no assumption gets broken and you can still `cancel` between contexts etc (not that I would recommend using saga interpreter leaving in a worker, but would certainly open more possibilities in such way)

Andarist:
Also if possibly other decorators could be implemented in the future, its always easier to compose numerous functions imho, than chain all of them on the effect itself.

The problem with chosing `catch` as function name is that it is a reserved word and cannot be used as such 😄 but thats just the naming problem.

You have also proposed a `wait` decorator, was it just other name for `catch` or did you have something else in mind for this one?

SpainTrain:
Any updated thoughts on this one?

Andarist:
@SpainTrain 
implementing smth like this - https://github.com/redux-saga/redux-saga/pull/759#issuecomment-278421268 would be super cool, thats the path i would like to follow with this one 

and ofc on top of this proposal it would be trivial to implement your `settle` proposal

PR #770, opened 2017-01-17
Update type definitions for recursive SagaIterator

dxiao:
Previously, the following code would not work:

```const rootSaga = function* () { yield takeLatest(...) };
createSagaMiddleware.run(rootSaga);```

This was because `rootSaga` returned a `IterableIterator<SagaIterator>`, which did not match type `SagaIterator`. This change makes `SagaIterator` recursive.

Note that change from type to interface is due to https://github.com/Microsoft/TypeScript/issues/3496#issuecomment-128553540

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,7 +4,8 @@ import {Task, Buffer, Channel, Predicate} from "./types";
 
 export {Effect, Pattern, Task, Buffer, Channel, Predicate};
 
-export type SagaIterator = IterableIterator<Effect|Effect[]>;
+type SagaIteratorReturns = Effect | Effect[] | SagaIterator | SagaIterator[];
+export interface SagaIterator extends IterableIterator<SagaIteratorReturns> {}
 
 type Saga0 = () => SagaIterator;
 type Saga1<T1> = (arg1: T1) => SagaIterator;

### CODE REVIEW

**Comment by aikoven** [comment]
I think the reason why you got an error here is because you `yield`ed an iterator instead of effects. To me this sounds like an anti-pattern, just like `yield`ing promises. Typings were designed to enforce you to always `yield` effects.

However, I may be missing something. cc @Andarist 

Also, what's the reason for making `SagaIterator` an interface instead of type alias?

**Comment by dxiao** [comment]
cause if it's a type alias, the recursion doesn't work. something subtle about how the typescript caches types.

I get that you should always yield effects if possible (for testing purposes) but seems weird to limit typings based on what's suggested, instead of what's possible.

**Comment by aikoven** [comment]
It is actually possible to `yield` any value and runtime will work fine, but if we change `SagaIterator` to be `IterableIterator<any>`, the typings will become much less strict and TS will miss much more errors.

I guess we need more opinions here.

**Comment by Andarist** [comment]
Unfortunately I dont know how to help you guys here as Im not using typings myself. 

Thats however true that you can `yield` literally anything, but that special types (promises, iterators, effects etc) will get interpreted in a special way by the `redux-saga`. So in general yielding promises/iterators directly could be seen as anti-pattern in most cases, but its still possible and if it works for anyone - for me its fine. There are cases (especially with promises) where wrapping them in `call` sounds for me like a serious overhead for a simple task. But ofc its arguable and in most cases I would still recommend wrapping them in a `call`.

But thats a classic example I think why I cant convince myself for using types. In the dynamic world of JS they just seems often to be... crippling (speaking about myself!). There seems to be always that edge case they do not support gracefully, even when nothing bad is happening. Especially heard that they do not go well with currying.

As to this PR - I've always count on you @aikoven in regards of TS typings. So... what should be done here? :D

**Comment by aikoven** [comment]
I think we can make a compromise here. Currently `SagaIterator` type only allows `yield`ing Effects, but the only place where it is checked is `SagaMiddleware.run()` method. Which means that only root saga is checked, unless you annotate return type of non-root sagas like this:
```ts
function* nonRootSaga(): SagaIterator {
}
```
We could loosen the type `SagaIterator` to `IterableIterator<any>`, but introduce another type `EffectIterator = IterableIterator<Effect>`, so that anyone who wants more strict checks could use it instead of `SagaIterator` to annotate their sagas.

aikoven:
Fixed in https://github.com/redux-saga/redux-saga/pull/740

Andarist:
Closing this in favor of the other PR, thanks you @dxiao for this PR and starting a discussion!

dxiao:
Thank you!

Mateusz Burzyński <notifications@github.com>于2017年2月9日 周四23:45写道：

PR #771, opened 2017-01-18
Enhancing onError callback for createSagaMiddleware

al-zah:
Allow for `onError` callback to take an additional param with getState function.

This would allow to retrieve snapshot of the last state before error did happen and possibly send it to error tracking service such as sentry.io.


Andarist:
Is it even possible to send big states to sentry?

slorber:
Hi,

Funny as I'm also working on getSentry.

I think it accept big JSON states yes.

It would only be for conveniency because you can easily call store.getState() directly in your error listener, I'm not sure it's worth changing the signature of onError just for this usecase

al-zah:
@Andarist Thats what sentry is recommends when using redux: https://blog.sentry.io/2016/08/24/redux-middleware-error-logging.html

@slorber You right, it's better to call getState directly, going to close this PR, sorry for wasted time :sweat_smile:

Andarist:
@al-zah still its probably preferable to you to use also `onError` handler in the `redux-saga` as not all exceptions can be caught by this sentry middleware.

PR #776, opened 2017-01-20
Support of Regular Expressions directly into `take` effect

notgiorgi:
I added support of regular expressions directly into `take` effect.


src/internal/utils.js
```diff
@@ -35,10 +35,11 @@ export const is = {
   task          : t => t && t[TASK],
   observable    : ob => ob && is.func(ob.subscribe),
   buffer        : buf => buf && is.func(buf.isEmpty) && is.func(buf.take) && is.func(buf.put),
-  pattern       : pat => pat && ((typeof pat === 'string') || (typeof pat === 'symbol') || is.func(pat) || is.array(pat)),
+  pattern       : pat => pat && ((typeof pat === 'string') || (typeof pat === 'symbol') || is.func(pat) || is.array(pat) || is.regex(pat)),
   channel       : ch => ch && is.func(ch.take) && is.func(ch.close),
   helper        : it => it && it[HELPER],
-  stringableFunc: f => is.func(f) && hasOwn(f, 'toString')
+  stringableFunc: f => is.func(f) && hasOwn(f, 'toString'),
+  regex         : exp => Object.prototype.toString.call(exp) === '[object RegExp]'
```
Andarist: `regexp: exp => exp instanceof RegExp` would be rather easier

garmjs:
I dont think it need to be in library

src/internal/utils.js
notgiorgi: This is the way mentioned in the specification for getting the class of object.
http://stackoverflow.com/a/4339385/5341953

notgiorgi:
Yep, there are ways to achieve this, but you can also manually do any of the supported patterns using your own predicate.

I think regex would be convenient. Checking for suffixes is common thing

src/internal/utils.js
Andarist: But we dont need to retrieve here the class of object but rather just test it against one (known one).

Your way is a more general, but should be rather used when somebody doesnt know against what to test, or just log something etc which might be of any type.

Andarist:
hey, ive seen that you have pushed a new commit with description of - `Merge branch 'master' into master`

did you try to update the branch against `redux-saga`'s master? if so, please reset your branch by 1 HEAD, rebase it on the origin's master (instead of merging it in) and push with force. Be cautious while doing this though as you might damage your branch if you make a mistake during the process.

Still I cant guarantee that this one is going to get included in `redux-saga`, but Im considering it.

PR #778, opened 2017-01-20
Check if process is defined

forest:
* process is not available in browser context

### DIFF
--- a/src/internal/utils.js
+++ b/src/internal/utils.js
@@ -10,7 +10,7 @@ export const kFalse = konst(false)
 export const noop = () => {}
 export const ident = v => v
 
-export const isDev = process.env.NODE_ENV === 'development'
+export const isDev = typeof process !== "undefined" && process.env.NODE_ENV === 'development'
 
 export function check(value, predicate, error) {
   if(!predicate(value)) {


Andarist:
Thanks for the PR, but im wondering in what scenario this is needed?

forest:
I have an app built with Aurelia CLI + TypeScript. It uses RequireJS as the loader. I load the main file `lib/index.js` and that loads `internal\utils.js`. This file includes the reference to `process` which is not defined in the browser context.

Maybe I'm doing something wrong.

Andarist:
How is it bundling modules internally? webpack? rollup? browserify? smth other?

forest:
I believe it uses its own system. I'm guessing there is a bug in [Aurelia CLI](https://github.com/aurelia/cli) since this is not an issue for others. Let's close this for now.

Is this code excluded when bundling with webpack, for instance?

Andarist:
Not exactly. Bundling with webpack or rollup replaces those with configured env vars. Please try to find out if Aurelia CLI could do the same or what other approach do they use to branch env dependent code.

If they wont provide any alternative I can merge your PR happily

Out of curiosity - why did u chose redux-saga for ur project?

forest:
Ahh. I didn't know they were replaced. That makes much more sense. I'll find out if they support that.

I have found redux to be a really nice way to manage state in an Aurelia project. redux-thunk has been working fine but is hard to test, as you mention in the saga docs. I also like how you can create flows with saga that are pretty easy to read. I originally found saga when I was looking for a way to have one redux state change trigger another change. This isn't how redux works, but saga flow makes this possible and very clean.

forest:
So there is a babel plugin to do this, but I'm using typescript as my transpiler. I ended up just adding something to my gulp pipeline at the end to replace `process.env.NODE_ENV`.

Andarist:
But is there some other standard approach for differentiating dev and prod builds then? would be nice if we can support this out of the box, instead of need for each user to configure it on his own

Andarist:
@forest any updates on the issue?

forest:
The Aurelia project has this code to be defensive about using environment variables.

```javascript
var hasEnvVariables = typeof process !== "undefined" &&
    typeof process.env !== "undefined";

function env(key) {
    return hasEnvVariables ? process.env[key] : undefined;
}
```

Not sure there is a better solution than being defensive.

Andarist:
Could you give reference to this piece of code? Where is it used?

If we must go defensive, we'll do. Im only now wondering how aurelia users differentiate envs. :D

forest:
Sorry. This is actually bluebird code. Aurelia doesn't use `process` in any of its code since it is expected to run in the browser and not a node environment.

https://github.com/petkaantonov/bluebird/blob/ac381051cd26a6d1b735d9f8a239710f68a899c4/src/util.js#L321

Andarist:
Differentiating between envs is not only made for node vs browser, but also for differentiating your dev and production builds - both targeting browser.

Anyway - I'll look later into the PR and we gonna go full defensive, it wont hurt the project after all and it'' help you

forest:
Sounds good.

src/internal/utils.js
```diff
@@ -10,7 +10,7 @@ export const kFalse = konst(false)
 export const noop = () => {}
 export const ident = v => v
 
-export const isDev = process.env.NODE_ENV === 'development'
+export const isDev = typeof process !== "undefined" && process.env.NODE_ENV === 'development'
```
Andarist: actually it is not the only place we use `process`, there are few (like 2 or so) more

Do you want to update your PR or should I just do it? by mistake Ive already done it on my local branch, but wouldnt like to take this away from you, so let me know :)

forest: Go for it. Thanks.

Andarist:
closing in favor of #825

@forest 
Please read this - https://github.com/redux-saga/redux-saga/pull/781#issuecomment-279866887 and let me know if you can wait for the release of the new version (v0.15) which will include this fix.

forest:
@Andarist I can wait. Thanks.

Issue #246, opened 2016-04-05
takeEvery doesn't work when action type is a Symbol


Zacqary:
When dispatching actions with a `type` equal to a `Symbol`, `takeEvery` throws `TypeError: Cannot convert a Symbol value to a string`.
```javascript
const ACTION_TYPE = Symbol("ACTION_TYPE Debug String");
```
```javascript
// Inside a saga
yield* takeEvery(ACTION_TYPE, sagaToYield); // Throws TypeError
```

However, using `take` works fine.
```javascript
for (;;) {
    yield take(ACTION_TYPE);
    // Yield results...
}
```

yelouafi:
The issue seems [from here](https://github.com/yelouafi/redux-saga/blob/master/src/internal/sagaHelpers.js#L49). String interpolation doesn't work for symbols

yelouafi:
fixed by https://github.com/yelouafi/redux-saga/commit/c9c63d95a95846162a3f51739d2b9dee41bd695a

jscinoz:
Seems to still be an issue when pattern is an array of symbols; the following line throws with TypeError: Cannot convert a Symbol value to a string:

https://github.com/yelouafi/redux-saga/blob/master/src/internal/sagaHelpers.js#L64

bvalosek:
(highly related question to this thread)

When using a action creator function with a `toString` property, currently that is preventing `take` from working (very similar to this issue).

eg

```
const login = credentials => {type: 'LOGIN', payload: credentials};
login.toString = () => 'LOGIN';

...

take(login)
```

Or more specifically: https://github.com/acdlite/redux-actions

Is ensuring that `take(pattern)` calls `String(pattern)` like the above fixes something we could have as well?

yelouafi:
@bvalosek What error do you get?

bvalosek:
I realized that this can't be implemented without breaking BC with your API.

I was asking if, when passing in a function that overrides `toString` that it could be cast to a string when running the matcher... but that would break the current expectations when calling `take(fn)` (as it serves as a predicate filter in this case).

https://github.com/yelouafi/redux-saga/blob/master/src/internal/proc.js#L17

What I am doing now is that i've made a utility `const takes = p => take(String(p))` to use in all the places where i have an action creator function that returns the action type when cast to a string. If you're open to having this be a built-in Effect then I can open a PR.

thezanke:
[...]

PR #781, opened 2017-01-21
Fixed action type to pattern matching in case type is a Symbol

iMoses:
Currently the default matcher transforms all patterns into strings, which is a problem in case of a pattern that needs to be matched against a Symbol. This fixes matching of Symbol action types, which are otherwise ignored by the system due to invalid comparison.

See issue #246

### DIFF
--- a/src/internal/proc.js
+++ b/src/internal/proc.js
@@ -11,7 +11,7 @@ export const TASK_CANCEL = {toString() { return '@@redux-saga/TASK_CANCEL' }}
 
 const matchers = {
   wildcard  : () => kTrue,
-  default   : pattern => input => input.type === (typeof input.type === 'symbol' ? pattern : String(pattern)),
+  default   : pattern => input => input.type === (typeof pattern === 'symbol' ? pattern : String(pattern)),
   array     : patterns => input => patterns.some(p => matcher(p)(input)),
   predicate : predicate => input => predicate(input)
 }
--- a/src/internal/proc.js
+++ b/src/internal/proc.js
@@ -11,7 +11,7 @@ export const TASK_CANCEL = {toString() { return '@@redux-saga/TASK_CANCEL' }}
 
 const matchers = {
   wildcard  : () => kTrue,
-  default   : pattern => input => input.type === String(pattern),
+  default   : pattern => input => input.type === (typeof input.type === 'symbol' ? pattern : String(pattern)),
   array     : patterns => input => patterns.some(p => matcher(p)(input)),
   predicate : predicate => input => predicate(input)
 }
--- a/test/proc/take.js
+++ b/test/proc/take.js
@@ -6,6 +6,8 @@ import * as io from '../../src/effects'
 test('processor take from default channel', assert => {
   assert.plan(1);
 
+  const typeSymbol = Symbol('action-symbol');
+
   let actual = [];
   const input = (cb) => {
     Promise.resolve(1)
@@ -16,6 +18,7 @@ test('processor take from default channel', assert => {
       .then(() => cb({isAction: true}))
       .then(() => cb({isMixedWithPredicate: true}))
       .then(() => cb({type: 'action-3'}))
+      .then(() => cb({type: typeSymbol}))
       .then(() => cb({...END, timestamp: Date.now()})) // see #316
     return () => {}
   }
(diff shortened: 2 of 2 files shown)


Andarist:
Hey, thanks for the PR. Could you provide a test for this?

iMoses:
Sure, my mistake for not thinking about it myself :)

Andarist:
Thank you for the fix!

iMoses:
I don't fully understand your question.

If you're talking about a case in which Symbol isn't supported at the browser, or even at Node, then `typeof` would never return a `"symbol"` string and behavior will continue as it does today.

The only place which may fail due to a lack of Symbol support is the test itself, in which case I'm not sure how important it is, you tell me.

Andarist:
> If you're talking about a case in which Symbol isn't supported at the browser, or even at Node, then `typeof` would never return a `"symbol"` string and behavior will continue as it does today.

Hm, I guess so - if fallback policy is a plain string and not some object.

iMoses:
@Andarist If you're worried that `pattern` may contain a value other than a string, I could change the condition to check if `pattern` is a Symbol instead of `input.type`, that way we can be sure that it's either a Symbol or a string, and it could never be anything else.

Andarist:
That actually would be preferable I think, and would provide a shorter code as well
`default   : pattern => input => typeof pattern === 'symbol' ? pattern : String(pattern)`

iMoses:
Aren't you missing the type comparison?

It should be:
```
default : pattern => input => input.type === typeof pattern === 'symbol' ? pattern : String(pattern)
```
Anyhow, I'll change it. It certainly makes more sense to validate against the pattern rather than the input.

Andarist:
Yeah, you are right, missed that cause of the rush. Merged and thank you for your work! :)

iMoses:
@Andarist Can you please increase the package version to include this fix in the latest npm version?

Andarist:
I'll try to do this in following days, just need to check if everything that got into `master` is ready for release.

just4fun:
Thanks @iMoses , I'm working with `0.14.3` and using Symbol as action type since I don't want to declare so many variables as action names. However, `take` is still suspended even I dispatched action.

After dug into source code, I found it has been fixed by this PR.

@Andarist I found latest version is still `0.14.3`, could you please update package version? Thank you!

Andarist:
Sorry guys that I hadnt time yet to release the new version. Im preparing it though, didnt have time in the weekend like ive hoped due to illness.

iMoses:
@Andarist Any updates on this?

Andarist:
Sorry for no update, i thought ive made an update here (had to be about some similar case). 

In general I'm prepering v0.15 and with current git flow on this project I really wouldnt like to release any intermediate versions. Im doing something nearly early day to reach this goal, but you know how it is - not much of a time after a day job.

Sorry for the inconvenience as such cases should be handled asap and nothing should stop it before releasing. 

The big picture is that Im pushing towards v1 and hope to have it released not that far into the future, just few things needs to be done before it happens. Then some important issues can get more priority - such as testing story and releasing process.

Could you maybe put link to the repo and the relevant commit instead of the npm's package version in your `package.json` for a time being? I would ping your asap when I release v0.15

just4fun:
@Andarist if I put link like `"redux-saga": "git://github.com/redux-saga/redux-saga.git#5858a7c"` in package.json, I will get `Unable to resolve module redux-saga from...`.

iMoses:
@Andarist - It's being 19 days since your last response.
It's taking a long time to promote the library's version, but the change discussed here is extremely small and manageable.
Why can't it be cherry picked into the current version, pushed into a new minor version and not be a part of the major changes which are currently happening in the code?
If it'll help I'll create a new PR from the last stable version, but I think it's kind of absurd that a one line change has to wait more than a month and a half to be deployed.

Andarist:
Sorry for the delay, I really wanted to push it faster - but I had really small amount of time to do so due to other stuff happening in my life. 

Would be really helpful if you provide a version I could just publish. Wondering how it should be done though. We do not want to merge anything to master, so aint sure what PR (from what to what branch) should be done 🤔 I think you need to push out a version with this cherry-picked commit going out from latest tag. But how should I incorporate it into this repo? Maybe I can just download your version of the repo then, checkout on correct branch/commit, tag it and publish - but aint sure if thats the best flow. WDYT?

just4fun:
@Andarist I know this repo is your side project (maybe), but could you please give out an ETA? since without this PR, we should always declare action name to string in redux, which is ugly.

alexgalin:
@Andarist Any update on this? we're really stuck due to this issue since we've started using sagas.

Andarist:
I promise to release this by the end of the week.

Andarist:
@alexgalin @iMoses @just4fun released [`v0.14`](https://github.com/redux-saga/redux-saga/releases/tag/v0.14.4) with this fix, sorry for such a delay, hope it helps and confirmation from you guys would be great that everything works fine as intended now

PR #784, opened 2017-01-24
Make the code snippet consistent with repo

justsayno:
The button and the div were formatted on three lines in the repo but one line in the tutorial. Makes it harder to find the place to put the code.

### DIFF
--- a/docs/introduction/BeginnerTutorial.md
+++ b/docs/introduction/BeginnerTutorial.md
@@ -88,9 +88,13 @@ First thing's first, we'll provide an additional callback `onIncrementAsync` to 
 const Counter = ({ value, onIncrement, onDecrement, onIncrementAsync }) =>
   <div>
     {' '}
-    <button onClick={onIncrementAsync}>Increment after 1 second</button>
+    <button onClick={onIncrementAsync}>
+      Increment after 1 second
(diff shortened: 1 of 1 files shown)


Andarist:
Thanks for your PR!

justsayno:
Would you be against updating other snippets? I found this one confusing:

``` lang-javascript
function render() {
  ReactDOM.render(
    <Counter
      value={store.getState()}
      onIncrementAsync={() => action('INCREMENT_ASYNC')}
    />,
    document.getElementById('root')
  )
}
```

Because in reality my final render should look like this:

``` lang-javascript
function render() {
  ReactDOM.render(
    <Counter
      value={store.getState()}
      onIncrement={() => action('INCREMENT')}
      onDecrement={() => action('DECREMENT')} 
      onIncrementAsync={() => action('INCREMENT_ASYNC')} />,
    document.getElementById('root')
  )
}
```

I think that is small enough function that there is no need to remove the other functions... Especially since they are still props in the Counter.js snippet right above.

Andarist:
> Would you be against updating other snippets?

Not at all :) Would only be good to provide reasoning behind each change with links to the places in the project it relates to, so its easier to track the change.

PR #785, opened 2017-01-24
BeginnerTutorial.md: Include full render method 

justsayno:
In the tutorial as is the original redux action creators `onIncrement` and `onDecrement` are not shown as props on the `Counter` component. This makes it so that the other buttons will no longer work if you follow the instructions to the letter.

Since the entire render method is very small I suggest including the whole render method like this:

``` lang-javascript
function render() {
  ReactDOM.render(
    <Counter
      value={store.getState()}
      onIncrement={() => action('INCREMENT')}
      onDecrement={() => action('DECREMENT')} 
      onIncrementAsync={() => action('INCREMENT_ASYNC')} />,
    document.getElementById('root')
  )
}
```

Making it much more clear how the `main.js` file should end up looking.

This is relating to the tutorial beginner tutorial:

[https://github.com/justsayno/redux-saga/edit/master/docs/introduction/BeginnerTutorial.md](https://github.com/justsayno/redux-saga/edit/master/docs/introduction/BeginnerTutorial.md) and the instructions on adding the `onIncrementAsync={() => action('INCREMENT_ASYNC')}` to the render method in main.js in the tutorial project.


Andarist:
Thanks for contributing!

PR #790, opened 2017-01-26
Fix: Under certain conditions scheduler may crash with "Maximum call stack size exceeded" error

pbadenski:
I run into this issue when using custom emitter and dispatching large number of actions from it. Those actions were directed through "asap" mechanism and put on a scheduler queue. If the number of actions is big enough once they are ready to process the recursive "exec -> flush" mechanism will cause stack overflow.

I propose that, when possible, actions are processed in a loop rather than by using recursion to avoid stack overflow.

See test scenario and attached screenshot for additional clarification.

screen shot 2017-01-25 at 17 43 41

### DIFF
--- a/src/internal/scheduler.js
+++ b/src/internal/scheduler.js
@@ -15,12 +15,18 @@ let semaphore = 0
   state).
 **/
 function exec(task) {
-  try {
-    suspend()
-    task()
-  } finally {
-    flush()
-  }
+  do {
+    if (!task) {
+      task = queue.shift();
+    }
+    try {
+      suspend();
+      task();
+    } finally {
+      task = undefined;
+      semaphore--;
+    }
+  } while (!semaphore && queue.length);
 }
 
 /**
--- a/src/internal/scheduler.js
+++ b/src/internal/scheduler.js
@@ -14,29 +14,25 @@ let semaphore = 0
   and flushed after this task has finished (assuming the scheduler endup in a released
   state).
 **/
-function exec(task) {
-  do {
-    if (!task) {
-      task = queue.shift();
-    }
+function exec() {
+  let task;
+  while (!semaphore && (task = queue.shift()) !== undefined) {
     try {
       suspend();
       task();
     } finally {
-      task = undefined;
       semaphore--;
     }
-  } while (!semaphore && queue.length);
+  }
 }
 
 /**
   Executes or queues a task depending on the state of the scheduler (`suspended` or `released`)
 **/
 export function asap(task) {
+  queue.push(task)
   if(!semaphore) {
-    exec(task)
-  } else {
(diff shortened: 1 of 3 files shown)


test/scheduler.js
```diff
@@ -1,4 +1,7 @@
+import test from 'tape'
+
 import { take, put, race } from '../src/effects'
+import { asap, flush, suspend } from '../src/internal/scheduler'
```
Andarist: previous tests in this file were not directly dependant on `scheduler`, is it neccessary to expose this implementation details here?

src/internal/scheduler.js
```diff
@@ -14,23 +14,25 @@ let semaphore = 0
...
 **/
-function exec(task) {
-  try {
-    suspend()
-    task()
-  } finally {
-    flush()
+function exec() {
+  let task;
+  while (!semaphore && (task = queue.shift()) !== undefined) {
```
Andarist: Shouldnt this while loop be a responsibility of `flush` rather than `exec`?

marcselman:
I see this error occur in combination with Redux-Form.
I have a page with a lot of form controls and, when mounting, for every control a "@@redux-form/REGISTER_FIELD" action is dispatched. And when unmounted a "@@redux-form/UNREGISTER_FIELD" action is dispatched.

If you want I can do some tests etc.

src/internal/scheduler.js
marcselman: Actually flush and exec can be combined in a single method (processQueue for example).

marcselman:
I can confirm this change works and solves the problem.

singggum3b:
I'm also having this problem when trying to dispatch many action from a list of react component.

mistadikay:
Do you need any help to get this merged? I'm waiting for a fix as well.

Andarist:
@mistadikay yeah, that would be great, as Im rly busy at the moment with other stuff and cant get to it right now :/

basically my last comments still apply - while the fix is working i would refactor it a little bit so iterating through tasks would be done in `flush` and not in `exec`

mistadikay:
@Andarist okay, let me see if I'll have some time tomorrow to take a look at it

marcselman:
What's the status of this pull-request? @mistadikay have you had time to take a look at it?

mistadikay:
@marcselman the project I need this PR for was down-prioritized, so I'm working on other stuff at the moment

Andarist:
just today @wilsaj offered his help to finish this one

wilsaj:
Hi everybody. I opened a follow-up PR to try to finish this up: https://github.com/redux-saga/redux-saga/pull/923

Andarist:
and I've just released it under `v0.14.7` 🎉

Issue #707, opened 2016-12-18
Why is there different behaviors with END.


awaw00:
I've written a simple saga demo [here](http://www.webpackbin.com/4ykwZgJNG)

In `main.js`, I run the root saga, when root saga is resolved, will log text 'DONE'.

After this, I dispatch an action 'START', then close saga use store.close():

```javascript
// main.js
import 'babel-polyfill';
import createStore from './createStore';
import sagas from './sagas';

const store = createStore();
const d = type => store.dispatch({type});
store.runSaga(sagas).done.then(() => bin.log('DONE'));

d('START');

store.close();
```

```javascript
// sagas.js
import { take, put, call, fork, takem } from 'redux-saga/effects';
import { delay } from 'redux-saga';

function *subTask () {
  let times = 0;
  while (times++ < 6) {
    bin.log(yield take('NEXT'));
    yield call(delay, 1000);
    bin.log('next');
  }
}
export default function *() {
  while (true) {
    yield take('START');
    bin.log('start');
    yield call(delay, 1000);
    bin.log('delayed');
    yield fork(subTask);
    bin.log('end');
  }
}
```

Check the Log tab and we can see output like:

**"start"
"delayed"
"end"**

It seems like saga hasn't been resolved yet.

I think this is the correct result because in #255 , @yelouafi said:

> The only way to escape from END will be inside forked tasks and within race effects.

But, if we remove two lines in sagas.js:

```diff
yield take('START');
bin.log('start');
--- yield call(delay, 1000);
--- bin.log('delayed');
yield fork(subTask);
bin.log('end');
```

The log output is:

**"start"
"end"
"DONE"**

-- Saga has been resolved, why?

awaw00:
@Andarist thanks for your reply.

Andarist:
@awaw00 

I've wanted to let you know that I'm working on the `stdChannel` so it can become a singleton, and im pretty much done, but have 1 case to work out, before its ready. Hopefully it wont take me forever ;) It also will help on some perf issues when one have really many tasks in his app.

yelouafi:
@Andarist Perhaps we should decouple `stdChannel` totally from `channel` as they seem to be completely different things: 

- one is unbuffered, multicast (dispatch to all subscribers), event pubsub (ie with matchers)
- while the other can be buffered, unicast (dispatch to one taker using FIFO rule), and matchers just doesn't make sense (as we encountered in a previous issue IIRC).

[...]

PR #801, opened 2017-01-31
Added support for self cancellation via yield cancel()

Andarist:
- [x] docs

In some discussions on reactiflux the idea of smth like `cancel.propagate()` came up, it would propagate through the task tree 1 level up, leaving further cancellation as responsibility for parent's finally block. Didnt think it through yet, just pitching an idea, but on the first look it could have been somehow leveraged for making supervisor tasks.

### DIFF
--- a/src/internal/io.js
+++ b/src/internal/io.js
@@ -1,4 +1,4 @@
-import { sym, is, ident, check, deprecate } from './utils'
+import { sym, is, ident, check, deprecate, SELF_CANCELLATION } from './utils'
 import { takeEveryHelper, takeLatestHelper, throttleHelper } from './sagaHelpers'
 
 const IO             = sym('IO')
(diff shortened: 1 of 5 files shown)


src/internal/io.js
```diff
@@ -108,24 +108,22 @@ export function join(...tasks) {
...
-  if(!is.task(tasks[0])) {
-    throw new Error(`cancel(task): argument ${tasks[0]} is not a valid Task object ${ TEST_HINT }`)
+  const task = tasks[0]
+  if (tasks.length === 1) {
+    check(task, is.notUndef, 'cancel(task): argument task is undefined')
+    check(task, is.task, `cancel(task): argument ${task} is not a valid Task object ${ TEST_HINT }`)
   }
-
-  return effect(CANCEL, tasks[0])
+  return effect(CANCEL, task || SELF_CANCELLATION)
```
Andarist: `SELF_CANCELLATION ` could be omitted if we `return effect(CANCEL, { task })`, but in that case self cancellation would depend on creating a `cancel` effect with `{ task: undefined }` so it can be interpreted later in `runCancelEffect`, didnt chose it as it seemed too implicit and non obvious

aikoven:
Isn't it easier to just `return`?

Andarist:
Not exactly, but I thought the same at first.

Difference is when you need both self cancellation and cancellation from outside via cancel and you need to handle both the same in your finally block within if (yield cancelled()) condition. Simple return wont have impact on the cancelled effect

aikoven:
I see. So, if I understand correctly, there are three ways to end up in `finally` block: return, cancellation and exception. If so, wouldn't it be more generic to be able to distinguish between return and exception somehow?

Andarist:
Well, exception specific logic should be done in a catch block. Or did you mean cancellation by parent's exception for example?

yelouafi:
>Isn't it easier to just return?

Another difference is that `return` just terminates the *main task* (the currently executing generator) but in order for the *whole task* to finish it has to wait for all its forked child tasks to terminate. see https://redux-saga.github.io/redux-saga/docs/advanced/ForkModel.html

self cancellation actually terminate the whole task (main tasks + childs)

Andarist:
> Another difference is that return just terminates the main task (the currently executing generator) but in order for the whole task to finish it has to wait for all its forked child tasks to terminate. see https://redux-saga.github.io/redux-saga/docs/advanced/ForkModel.html

wow, didnt know that difference, so many surprises yet to come for me i guess 😄

yelouafi:
>wow, didnt know that difference, so many surprises yet to come for me i guess 😄

It's totally understandable, the behavior may seem unexpected at first, but it works nicely with server side rendering. Closing the store on the server (by dispatching END action) will cause all watcher to terminate the main task (watcher body), but they will still wait for all forked workers to finish (get fetch results, update the store state) before terminating. 

In final, we render to the client after all work is done. This way we can have *universal sagas*, the same saga (and also reducer & React) code can work both on client and server

Andarist:
Speaking about server rendering and `END`, could you give a comment [here](https://github.com/redux-saga/redux-saga/issues/707#issuecomment-277562954) ? 😈

yelouafi:
leaving from work right now. I'll look at this tonight, Promise :)

PR #809, opened 2017-02-04
Add onError as option to runSaga

xanf:
runSaga actually accepts onError option in the same way as it is done in middleware


Andarist:
Yep, you are right, forgot to update the docs. Thanks for contributing!

PR #810, opened 2017-02-06
Improved documentation wording

aryzing:


docs/introduction/BeginnerTutorial.md
```diff
@@ -198,8 +198,7 @@ test('incrementAsync Saga test', (assert) => {
 });
 ```
 
-Since `incrementAsync` is a Generator function, when we run it outside the middleware,
-Each time you invoke `next` on the generator, you get an object of the following shape
+`incrementAsync` is a generator function. When run, it returns a generator object. The generator object's `next` method returns an object with the following shape
```
Andarist: It even should be changed further - generator function returns an iterator object when run

aryzing:
New PR with suggested changes in #811.

PR #818, opened 2017-02-11
Improve "External Resources" document with new links

nktssh:
Add links for new article on `redux-saga` topic and new addon (bridge between Vuex and Redux-Saga)


Andarist:
Thanks for the update!

PR #824, opened 2017-02-14
stdChannel remade to be a singleton, therefore a multicast now. Expos…

### LINKED ISSUE #707: Why is there different behaviors with END.

Andarist:
…ed multicast channel as primitive for external usage

it should fix #707

@yelouafi 
I would like u to comment on the code, its not finished yet - will try to build `stdChannel` on top of `multicast`, need to get rid of `broadcast` so strict difference between regular channel (unicast) and the new multicast channel is possible and that both should share the same interface of `put` alone.

In general I didnt have time to refactor yet as I wanted to make tests work first. From what I can judge they are failing solely because `proc`'s API changed and once they are adjusted they will just start to work. Didnt want to do that without preliminary code review.

Also I think it would be better if internal tests didnt rely so much on such an implementation detail. After all `proc` is not an exposed API. WDYT?

EDIT://
Aint sure yet but it could fix this - http://www.webpackbin.com/4J2hZmktG - scheduling 'problem'. Wondering if it is even a problem, but I would like to think of takes as FIFO. Ive ran into this in my real app and had to debug quite a bit to solve it by a workaround for now.

I suppose it may happen because emitter dispatched the response action into multiple stdChannels (as there are multiple tasks) and for some reason they run out of order. Possibly with 1 stdChannel it would just work more like FIFO from the start.


yelouafi:
I did a quick review. While I totally agree on having a global shared StdChannel, as I commented on #707  I think it's be both better but also **safer** to have a totally separate implementation for multicast channels because their logic is quite different from unicast ones. 

The most obvious is related to buffering, I don't think multicast should buffer their input when there are no takers. because some input actions could get buffered infinitely it there are not takers interested in them. Similarly when a saga later take a specific action it shouldn't get an old buffered actionbecause it's typically interested in future actions. 

Also having a separate multicast we can get rid of the MATCHER hack inside unicast channels, because I tink similarly it doesn't make also for here ( or at least making their logic more confusing and complicated)

I think it'd be a lot cleaner to have separate abstractions/implementations for the 2 types of channels

- multicast channels: no buffering, take can get a matcher as argument
- unicast: buffering, no matchers

Andarist:
Yeah, i completely agree that `multicast` should be completely separate, tried to point this out in the description of the PR, maybe not well enough

Just did it like this temporarily when i was hacking around and had no time to refactor. Just wanted to w8 on your feedback - so I could implement more final version on top of this.

For example I didnt think of multicast having no buffer - I guess it makes sense in most cases.

> Also I think it would be better if internal tests didnt rely so much on such an implementation detail. After all proc is not an exposed API. WDYT?

would like to have ur input on this matter, that was the biggest hold up for me with this PR as i didnt want to refactor quite all tests just to find out it should be tackled in some other manner ;)

yelouafi:
>Also I think it would be better if internal tests didnt rely so much on such an implementation detail. After all proc is not an exposed API. WDYT?

I think you're right. Looking at the current signature of `proc` it seems calls to it in all tests could be replaced with `runSaga` (or `middleware.run`). extra params (like `parentEffectId`, `cont` ..) are not actually needed in tests.

as for the other points

>- proc signature changes

Not sure if I understood the question. But I think now `proc` will take a multicast channel instead of `subscribe`. The middleware will ofc build a global one that will get passed around to child proc calls. 

For `runSaga` either we keep the api as it is (and then it should also build a multicast channel itself) or we change the API and make it take a channel as argument (or support both options). The advantage of the second is that multiple `runSaga`s can be connected together using the same channel 

> actionChannel's support for multicasting - extra param?

I can't see how that would be useful. The point of `actionChannel` is to buffer actions while the saga is being blocked. If we make it multicast it means no buffering and thus it will behave the same as the global Redux channel

> blocking channel idea

actually the example you mentioned corresponds to how CSP channels behave (put is blocking until someone takes the message or it's buffered). When I first started adding channels, my intention was to make them CSP compliant. But finally settled on a non blocking model.

The main reason was that blocking channels *may* deadlock. If 2 sagas are using the same channel as a bidirectional medium there could be a situations where the 2 will be trying to put in a full channel, thus waiting for each other to take an item to free the channel. 

Also we can implement CSP blocking by using 2 channels: 1 to send and the other to acknowledge, and without deadlock risks. [see those 2 test cases](https://github.com/redux-saga/redux-saga/blob/master/test/proc/takeSync.js#L378-L466)).

> ur comment about "unicast: buffering, no matchers" - eventChannel still needs matcher I think, OTOH it could be achieved in userland within subscribe callback

That's right. We can rework channels spec : to support matchers we need to account of the case that we can have buffered channels AND queued takers in the same time.  But I wonder if this is worth it?

Andarist:
> For runSaga either we keep the api as it is (and then it should also build a multicast channel itself) or we change the API and make it take a channel as argument (or support both options). The advantage of the second is that multiple runSagas can be connected together using the same channel

we could offer both APIs, just wondering if its worth it, `channel` is simplification in this regard I think as it provide emitter like behaviour and doesnt conflict with redux `subscribe` (which has different API - its argument-less),

cc @slorber does `subscribe` solves any problem for you? would channel API suits u?

> That's right. We can rework channels spec : to support matchers we need to account of the case that we can have buffered channels AND queued takers in the same time. But I wonder if this is worth it?

not sure if I follow on this one - ideally we drop matchers, it only should be somehow internally exposed for `stdChannel`'s takes

I've refactored the code - but there are some open issues with it, I've put TODO comments in the code so its easier to point to the problematic parts. Will wait for your review - u can skip test refactors, but please dont skip comments put there.

slorber:
Hi, I can't provide much feedback @Andarist I don't have used much channels so far

Andarist:
In runSaga example it would act as a bus for passing ur actions into the redux-saga runtime, so in general should pretty much cover the same use case as subscribe option. Only in a slightly different way.

test/proc/take.js
```diff
@@ -33,23 +25,36 @@ test('processor take from default channel', assert => {
       actual.push( yield io.take(['action-3', a => a.isMixedWithPredicate]) ) // take if match any from the mixed array
       actual.push( yield io.take(typeSymbol) ) // take only actions of a Symbol type
       actual.push( yield io.take('never-happening-action') ) //  should get END
+      // TODO: never-happening-action replaced with such case is not working
```
Andarist: cc @yelouafi

test/proc/channel.js
```diff
@@ -21,54 +20,51 @@ test('proc create channel for store actions', assert => {
...
-  const spyBuffer = {
-    isEmpty: () => !buffer.length,
-    put: ({payload}) => buffer.push(payload),
-    take: () => buffer.shift()
-  }
+  const buffer = buffers.expanding()
 
   function* genFn() {
-    let chan = yield io.actionChannel('action', spyBuffer)
+    // TODO: this might mean that we do not close / flush channels when sagas ends
```
Andarist: cc @yelouafi

ms88privat:
Well done. Was waiting for this :) What about the documentation? hehe.
Thanks a lot guys!

Andarist:
@ms88privat I will update documentation before the final v1 release (which might take some time :/). I've already published very much usable [redux-saga@1.0.0-beta.0](https://github.com/redux-saga/redux-saga/releases/tag/v1.0.0-beta.0) if you are interested in using it.

PR #825, opened 2017-02-14
Going defensive against lack of process.env - this will allow usage i…

Andarist:
…n i.e. Aurelia project

cc @forest


forest:
👍

src/internal/utils.js
```diff
@@ -11,7 +11,11 @@ export const kFalse = konst(false)
 export const noop = () => {}
 export const ident = v => v
 
-export const isDev = process.env.NODE_ENV === 'development'
+const hasEnv = typeof process !== 'undefined' && typeof process.env !== 'undefined'
+
+export const env = (key) => hasEnv ? process.env[key] : null
+
+export const isDev = env('NODE_ENV') ? env('NODE_ENV') === 'development' : false
```
aikoven: There's a chance that this won't work with this Webpack configuration:
```js
new DefinePlugin({
  'process.env': JSON.stringify({NODE_ENV: 'development'}),
})
```
Because Webpack only replaces `process.env` and not `process`. This is how it's done in e.g. `create-react-app`: https://github.com/facebookincubator/create-react-app/blob/v0.9.0/packages/react-scripts/config/env.js

I may be mistaken, but I remember I've ran into such issue before.

Andarist: Thanks for the comment, I need to definitely look into this.

Andarist:
@aikoven, this thing that its in fact string replacing later is making it hard to reason about, is the updated code better? 🤔 

problem is that ive spotted now that using `process` alone is causing its shim to be added to the dist files :s

aikoven:
@Andarist Unfortunately I'm not aware of the most robust way of doing this.

Current Redux approach is just evaluating `process.env.NODE_ENV` without any checks: https://github.com/reactjs/redux/blob/v3.6.0/src/index.js

However, in the `next` branch they still do the check for `process`: https://github.com/reactjs/redux/blob/next/src/index.js

src/internal/middleware.js
```diff
@@ -31,6 +29,8 @@ export default function sagaMiddlewareFactory(options = {}) {
           const store = createStore(reducer, applyMiddleware(sagaMiddleware))
           sagaMiddleware.run(saga, ...args)
       `)
```
yelouafi: the long msg was introduced with breaking changes of 0.10. perhaps we could get rid of it now and just keep the short message

Andarist: we'll do! :)

Andarist:
@forest 
unfortunately i have no idea how to configure the tooling, so you (as aurelia user) could use this and webpack/rollup users at the same time wouldnt be affected by this

a little bit more about the matter can be read [here](https://github.com/reactjs/redux/issues/2244#issuecomment-287634636)

do u know any other package that u use in combination with aurelia and which uses `process.env` variables? maybe other packages have it configured somehow better and i could look on it, but for now it becomes an open issue here :/

Andarist:
The problem is that doing it like this is causing process polyfill to be included in the dist files. We cannot afford doing this :/ 

As @aikoven observed process.env.SOME_VAR is simply replaced (think about code as a string) in the code in webpack/rollup world. Referencing process alone on the other hand is treating it as process (node's global object) and therefore polyfilling it automatically so it wont crash in the browser

forest:
@Andarist I understand now.

Can you change this https://github.com/redux-saga/redux-saga/blob/master/webpack.config.dev.js#L10 to 
```
PRODUCTION: JSON.stringify(false)
```

and this https://github.com/redux-saga/redux-saga/blob/master/webpack.config.prod.js#L10 to 
```
PRODUCTION: JSON.stringify(true)
```

I think that will create a global constant that is just for your purposes and don't conflict with something existing.

Andarist:
Do u know any project which uses it like this?

forest:
The docs of DefinePlugin give this as an example.
https://webpack.github.io/docs/list-of-plugins.html#defineplugin

Andarist:
I think this would require u to create such globals, which is rather a brute force approach.

Is there a way to incorporate a tool such as https://github.com/hughsk/envify to you aurelia bundling process?

forest:
@Andarist yes. I could use envify. I already did something similar to work around this issue. I opened this issue because it didn't seem like my responsibility to handle that, but if there isn't a good solution then other users will need to deal with this too.

Andarist:
Im going to close this. Thought before it could be handled on our side, but now I realize that its not possible and that both webpack and rollup users actually do the same. Ive just made a PR to replace our umd build tool for rollup and I had to do the same [here](https://github.com/redux-saga/redux-saga/blob/97af91123687d22202dbb70405293b51e5059a3b/rollup.config.js#L19). Going to close this.

Sorry that it took so much time to finally establish that we cant do this here.

PR #826, opened 2017-02-15
Added support for passing names of methods in string for call etc whe…

Andarist:
…n passing context, i.e call([context, 'method'])

cc @aikoven ;)

### DIFF
--- a/src/internal/io.js
+++ b/src/internal/io.js
@@ -79,6 +79,9 @@ function getFnCallDesc(meth, fn, args) {
   } else if(fn.fn) {
     ({context, fn} = fn)
   }
+  if (context && is.string(fn) && is.func(context[fn])) {
+    fn = context[fn]
+  }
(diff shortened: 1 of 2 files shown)


src/internal/io.js
```diff
@@ -77,6 +77,9 @@ function getFnCallDesc(meth, fn, args) {
   } else if(fn.fn) {
     ({context, fn} = fn)
   }
+  if (context && is.string(fn)) {
+    fn = context[fn]
+  }
   check(fn, is.func, `${meth}: argument ${fn} is not a function`)
```
yelouafi: In case a wrong string is provided, since `fn` was mutated in the previous line, the message will print smth like `argument undefined is not a function`. perhaps better to keep the string name for more helpful messages (since it's more likely that an error could happen with string args)

Andarist: yeah, i even think i had that thought but forgot to do something about it, will add this and merge in, thanks for the review

PR #828, opened 2017-02-17
add test utilitary function cloneableGenerator

Nihaux:
We use redux-saga at work and in order to test some of them I created a util function cloneableGenerator.
It allows you to clone a generator at a certain point of it's execution.

This is usefull when you want to test a saga which can have differents endings.
ex: the user cancels/approves the order, the http response status, etc

You don't have to create a new generator and "replay" all the steps necessary to come to the branching point.
You can just "clone" the generator at a certain step and use the cloned generator to test the other case.

### DIFF
--- a/docs/advanced/Testing.md
+++ b/docs/advanced/Testing.md
@@ -1,8 +1,185 @@
-# Examples of Testing Sagas
+# Testing Sagas
 
-**WIP**
+**Effects return plain javascript objects**
 
-See Repository Examples:
+Those objects describe the effect and redux-saga is in charge to execute them.
+
+This makes testing very easy because all you have to do is compare that the object yielded by the saga describe the effect you want.
+ 
+## Basic Example
+
+```javascript
+console.log(put({ type: MY_CRAZY_ACTION }));
+
+/*
+{
+  @@redux-saga/IO': true,
+  PUT: {
+    channel: null,
+    action: {
+      type: 'MY_CRAZY_ACTION'
+    }
+  }
+}
+ */
+```
+
+Testing a saga that wait for a user action and dispatch
+
+```javascript
+const CHOOSE_COLOR = 'CHOOSE_COLOR';
+const CHANGE_UI = 'CHANGE_UI';
+
+const chooseColor = (color) => ({
+  type: CHOOSE_COLOR,
+  payload: {
+    color,
+  },
+});
+
+const changeUI = (color) => ({
+  type: CHANGE_UI,
+  payload: {
+    color,
+  },
+});
+
+
+function* changeColorSaga() {
+  const action = yield take(CHOOSE_COLOR);
+  yield put(changeUI(action.payload.color));
+}
+
+test('change color saga', assert => {
+  const gen = changeColorSaga();
+
+  assert.deepEqual(
+    gen.next().value,
+    take(CHOOSE_COLOR),
+    'it should wait for a user to choose a color'
+  );
+
+  const color = 'red';
+  assert.deepEqual(
+    gen.next(chooseColor(color)).value,
+    put(changeUI(color)),
+    'it should dispatch an action to change the ui'
+  );
+
+  assert.deepEqual(
+    gen.next().done,
+    true,
+    'it should be done'
+  );
+
+  assert.end();
+});
+```
+
+Another great benefit is that your tests are also your doc! They describe everything that should happen.
+
+## Branching Saga 
+
+Sometimes your saga will have different outcomes. To test the different branches without repeating all the steps that lead to it you can use the utility function **cloneableGenerator**
+```javascript
+const CHOOSE_NUMBER = 'CHOOSE_NUMBER';
+const CHANGE_UI = 'CHANGE_UI';
+const DO_STUFF = 'DO_STUFF';
+
+const chooseNumber = (number) => ({
+  type: CHOOSE_NUMBER,
+  payload: {
+    number,
+  },
+});
+
+const changeUI = (color) => ({
+  type: CHANGE_UI,
+  payload: {
+    color,
+  },
+});
+
+const doStuff = () => ({
+  type: DO_STUFF, 
+});
+
+
+function* doStuffThenChangeColor() {
+  yield put(doStuff());
+  yield put(doStuff());
+  const action = yield take(CHOOSE_NUMBER);
+  if (action.payload.number % 2 === 0) {
+    yield put(changeUI('red'));
+  } else {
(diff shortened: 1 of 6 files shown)


test/utils/clonableGenerator.js
```diff
@@ -0,0 +1,156 @@
+import test from 'tape';
```
Andarist: this file has wrong name - it lacks the 'e'

test/utils/clonableGenerator.js
```diff
@@ -0,0 +1,156 @@
...
+import { spy } from 'sinon';
+import { cloneableGenerator } from '../../src/internal/utils';
+
+test('it should proxy the original generator', assert => {
+  assert.plan(4);
+
+  const spyGenerator = {
```
Andarist: technically this is an iterator, not a generator

definition of the generator is a function which produced an iterator object

test/utils/clonableGenerator.js
```diff
@@ -0,0 +1,156 @@
...
+  assert.deepEqual(
+    clonedGen2.next(),
+    {
+      value: undefined,
+      done: true
+    }
+  );
+
+  assert.end();
+});
```
Andarist: tests should include an example of the logic which it simplifies - branching

src/internal/utils.js
```diff
@@ -148,3 +148,24 @@ export function wrapSagaDispatch(dispatch) {
     return dispatch(wrappedAction)
   }
 }
+
+/**
```
Andarist: i think rest of the code is not commented like this

src/internal/utils.js
```diff
@@ -148,3 +148,24 @@ export function wrapSagaDispatch(dispatch) {
...
+ * takes a generator and return a wrapped generator that you can clone.
+ */
+export const cloneableGenerator = (generatorFunction) => (...args) => {
+  const history = [];
+  const gen = generatorFunction(...args);
```
Andarist: same remark about generator vs iterator semantics, this should be `const it`

src/internal/utils.js
```diff
@@ -148,3 +148,24 @@ export function wrapSagaDispatch(dispatch) {
     return dispatch(wrappedAction)
   }
 }
+
+/**
+ * takes a generator and return a wrapped generator that you can clone.
+ */
+export const cloneableGenerator = (generatorFunction) => (...args) => {
```
Andarist: `generatorFunction` feels verbose, probably just `generator` would be ok

also thinking if this should be just called `cloneable`, but pick whichever you want for now

test/utils/clonableGenerator.js
```diff
@@ -0,0 +1,156 @@
+import test from 'tape';
+import { spy } from 'sinon';
+import { cloneableGenerator } from '../../src/internal/utils';
+
+test('it should proxy the original generator', assert => {
+  assert.plan(4);
+
+  const spyGenerator = {
+    args: undefined,
+    next: spy(),
```
Andarist: couldnt it be tested like rest of the code? I mean pushing to the `actual` within `next`, `throw`, `return` etc and in the end just comparing `actual` vs `expected`?

test/utils/clonableGenerator.js
Nihaux: According to the MDN,

> The function* declaration (function keyword followed by an asterisk) defines a generator function, which returns a Generator object.

> The Generator object is returned by a generator function and it conforms to both the iterable protocol and the iterator protocol.

So it's kind of both :)
In my original code, it was called iterator but looking at MDN, I renamed it generator.
I you think its best calling it iterator, i will, i just wanted to put it outhere

Andarist: i would keep change it to iterator, as it brings less confusion, dont u think?

Nihaux: So I did a little digging to understand what is what.
According to the ECMA 2015 spec (just started to read it from scratch, very interesting stuff :) )
The correct naming would be:

`function* helloWorld() {...}`
=> helloWorld is a generator function, it is an instance of (GeneratorFunction) constructor and it extends the (Generator) prototype.

`const helloGen = helloWorld();`
=> helloGen is an instance of the helloWorld constructor, it extends helloGen.prototype which extends (Generator).prototype which extends (IteratorPrototype).

 helloGen is a kind of Generator, which is a kind of Iterator.

So I think it would be more accurate to name it gen for generator. Even if it seems confusing at first, I think it's better to stick as close as possible to the spec.

I find that [this schema](http://www.ecma-international.org/ecma-262/6.0/#sec-generatorfunction-objects) makes it more understandable.

Let me know if you think I missed/missunderstood sth :)

src/internal/utils.js
Nihaux: I don't really understand what you mean, should I remove the comment or format it differently ?

Andarist: rather remove, i think this function is quite self-explanatory

Nihaux:
I added documentation, refactored the tests to be closer to the real use case and squashed the commits.

src/internal/utils.js
```diff
@@ -148,3 +148,21 @@ export function wrapSagaDispatch(dispatch) {
...
+export const cloneableGenerator = (generatorFunc) => (...args) => {
+  const history = [];
+  const gen = generatorFunc(...args);
+  return {
+    next: (arg) => {
+      history.push(arg);
+      return gen.next(arg);
+    },
+    clone: () => {
+      const clonedGen= cloneableGenerator(generatorFunc)(...args);
```
Andarist: missing whitespace before equals sign

docs/advanced/Testing.md
```diff
@@ -1,8 +1,172 @@
...
+    yield put(changeUI('red'));
+  } else {
+    yield put(changeUI('blue'));
+  }
+}
+
+import { put, take } from 'redux-saga/effects';
+import { cloneableGenerator } from 'redux-saga/utils';
+
+describe('doStuffThenChangeColor', () => {
```
Andarist: other parts of the docs when describing the testing story are using `tape`, consistency would be appreciated by the reading users i think

Andarist:
Besides the parts with my comments everything else looks good and is ready to merge. 

Btw thanks for squashing <3

Nihaux:
switched testing doc to tape style and fixed typo.

Nihaux:
hello @Andarist anything blocking this PR to be merged ?

Andarist:
Merged in!

Nihaux:
@Andarist thx !

Issue #835, opened 2017-02-22
disable logger in production builds


anthonychung14:
production builds have logging still happening from the utils file

https://github.com/redux-saga/redux-saga/issues/343

there is no documentation (that I could find) indicating how to accomplish this other than stating that there is a  "logger" option in the createSagaMiddleware

mainly all the console.infos that inform me when a task is cancelled

Andarist:
According to the [source code](https://github.com/redux-saga/redux-saga/blob/3b90d33d27cb1a5aa99e8672b47cf8e496d686f6/src/internal/proc.js#L277-L279) this log should happen only in dev builds. Are you sure you have correctly configured your build?

anthonychung14:
You are correct. I did not read the source code/issue correctly in the above issue

PR #836, opened 2017-02-23
disable logger in production builds

restrry:
https://github.com/redux-saga/redux-saga/issues/835


Andarist:
Hey, thanks for the PR, but im not sure if we should just drop all logs. Most of the logs are restricted for dev build in the source code + those which are not are only in cases of thrown error. If we drop it it might swallow some info which is easing the post mortem debugging experience.

PR #853, opened 2017-03-05
Here is the logo in 3 different flavors, two resolutions, with README and LICENSE

karland:


Andarist:
Wow, great! Thank you very much

yelouafi:
Awesome! thanks @karland

PR #863, opened 2017-03-13
feat: add generator function resolve

ipluser:
I think somebody will write below code:

```js
function* preTask2() {
  // todo
}

function* task1() {
  // ...
  yield takeEvery('*', (action) => {
    // todo
  })
}

function* task2() {
  // ...
  yield preTask2
}

function* rootSaga() {
  // ...
  yield [task1, task2]
  // ...
}

sagaMiddleware.run(rootSaga)
```

but it doesn't work, sometime it's not easy to find this wrong.
I know that we should invoke the generator function to generator iterator or use `call` effect method:

```js
// ...
yield [task1(), task2()]
// or
yield [call(task1), call(task1)]

//...
yield preTask2()
// or
yield call(preTask2)
```

so I hope add **generator function resolve** feature for the `runEffect`.

### DIFF
--- a/src/internal/proc.js
+++ b/src/internal/proc.js
@@ -372,6 +372,7 @@ export default function proc(
         is.promise(effect)                                   ? resolvePromise(effect, currCb)
       : is.helper(effect)                                    ? runForkEffect(wrapHelper(effect), effectId, currCb)
       : is.iterator(effect)                                  ? resolveIterator(effect, effectId, name, currCb)
+      : is.generator(effect)                                 ? resolveIterator(effect(), effectId, name, currCb)
 
       // declarative effects
       : is.array(effect)                                     ? runParallelEffect(effect, effectId, currCb)
--- a/src/internal/utils.js
+++ b/src/internal/utils.js
@@ -33,6 +33,10 @@ export const is = {
   array          : Array.isArray,
   promise        : p => p && is.func(p.then),
   iterator       : it => it && is.func(it.next) && is.func(it.throw),
+  generator      : it => it && is.func(it) && (
+                      'GeneratorFunction' === it.constructor.name ||
+                      'GeneratorFunction' === it.constructor.displayName ||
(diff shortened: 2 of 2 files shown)


Andarist:
Sorry, but I dont feel it should be handled on our side, with such APIs ppl tend to overuse them and I think it would be quite harmful in a long run. Would like to prefer explicit APIs than implicit ones, even `yield [...effects]` is going to be deprecated soon in favor of `yield all([...effects])`, just so it can be more explicit in the code about what it does.

Andarist:
I think though that when [this](https://github.com/redux-saga/redux-saga/pull/729) gets merged in you could implement it easily by yourself on your side, if you feel thats a good thing for your project

ipluser:
@Andarist  @yelouafi 

I'm happy to see the discuss [Feat/run saga middleware](https://github.com/redux-saga/redux-saga/pull/729).
It's easy to meet my requirement, the **saga middleware** can do more (e.g. resolve others custom effects) that like **redux middleware**.

But, in this issue, I have different thoughts:

e.g.

```js
// node.js, like koa.js

function* generator() {
  // todo
}

// ....

yield generator
// or 
yield generator()
```

In the above code, it's work, cuz it integrate [co](https://github.com/tj/co).

The **redux-saga** that has built-in **generator** based flow-control too. 
But, now you tell me that the **redux-saga** cann't resolve **generator function**, must be an **generator iterator**.

e.g.

```js
function* generator() {
  // todo
}

// ...
yield generator  // terrible

yield generator()  // good, must be to an iterator
```

I think the **redux-saga** run **generator** effect that is a basic feature, not just **iterator** effect.

If you think that my point is not good, I hope we can add some description to warn developer.

e.g.

If you invoke some function, we recommend you use **call** effect.

or others notifaction

ipluser:
@Andarist 

I prefer to use`yield all([...effects])` that sounds great.

Issue #625, opened 2016-11-11
real-world example: import declarations Syntax Error


ci5er:
When running 'npm start', loading http://localhost:3000/, in either FF 49.0.2 (and Chrome 54.0.xxxx...), in the debug console (Firebug on FF), I am seeing two variations of the error:
> SyntaxError: import declarations may only appear at top level of a module
> import middleware from './internal/middleware';

Both the redux-devtools panel and the input field (and submit button) underneath the prompt "Type a username or repo full name and hit 'Go':" are also unresponsive. By which I mean: text can be entered, but the "Go" button does nothing.

Ubuntu 16.04, 64-bit, if that makes any difference.

Andarist:
I got it working, but gonna publish it later, cause made it kinda in a clumsy way, will try to do it better. For now you could paste in this webpack config - https://jsfiddle.net/nggwhr5g/

ci5er:
Thanks. There's too much going on at the bottom of the file you changed for me to follow, but I can confirm that it works.

carlos-peru:
Hi @Andarist,

I'm trying to understand the change in your fiddle because it made it worked (Thanks!). Webpack is a new thing to me, and I would highly appreciate a short clarification.

In the top module exports change, did you just set a loader instead of an array of loaders and explicitly set the presets as query parameters as opposed to waiting for babel to pull the presets from .babelrc? Why?

In the bottom module exports change, did you include the redux saga and saga monitor modules dynamically in the babel loader instead of just adding them to the arrays of loaders (that was replaced by a single loader at the top)? Why?

Thanks again!

PR #871, opened 2017-03-17
Fix webpack config for realworld example https://goo.gl/3R7bt7

jaimeagudo:
My 0,02$ back @Andarist, I see you are quite busy so I just pasted your solution. Thanks
Closes https://github.com/redux-saga/redux-saga/issues/625


Andarist:
thanks!

PR #877, opened 2017-03-20
Fixes #875

### LINKED ISSUE #875: eventChannel emit END, unsubscribe not called
In the advanced docs for [using channels](https://github.com/redux-saga/redux-saga/blob/master/docs/advanced/Channels.md#using-the-eventchannel-factory-to-connect-to-external-events), I don't understand why the `clearInterval` is written in 2 places.

Here's the code for reference:
```
import { eventChannel, END } from 'redux-saga'

function countdown(secs) {
  return eventChannel(emitter => {
      const iv = setInterval(() => {
        secs -= 1
        if (secs > 0) {
          emitter(secs)
        } else {
          // this causes the channel to close
          emitter(END)
          clearInterval(iv)
        }
      }, 1000);
      // The subscriber must return an unsubscribe function
      return () => {
        clearInterval(iv)
      }
    }
  )
}
```

As I understand it, calling `emitter(END)` causes the channel to close. And according to the docs, the `unsubscribe` function is "`used by the channel to unsubscribe before the event source complete.`"

So, in the case of the above code, I would expect calling `emitter(END)` to cause the `unsubscribe` function to be executed since we just closed the channel by emitting the END event. However, this is not the case, as `clearInterval` is coded in both places, i.e. emitting END event does NOT call the unsubscribe function.

Is there something I'm not understanding about emitting an END event? Does the unsubscribe function only get called when the consumer saga calls `channel.close`?

baldwmic:
### What the PR does
- extracts `close` function into variable to ensure `unsubscribe` function gets called when `END` event is emitted.
- updates advanced docs for channels to show that unsubscribe function will be called 

#### Note:
- this is my first code PR for redux-saga, I'm not sure the best way to test this? I see the appropriate [test code here](https://github.com/redux-saga/redux-saga/blob/e237a20e6973585c44b5a098a688a4a3c102c033/test/channel.js#L195) but I'm used to using [Sinon](http://sinonjs.org/) to test that a function got called and I don't see that in this repo.
- thanks to @Andarist for the help! 😄

### DIFF
--- a/src/internal/channel.js
+++ b/src/internal/channel.js
@@ -124,9 +124,15 @@ export function eventChannel(subscribe, buffer = buffers.none(), matcher) {
   }
 
   const chan = channel(buffer)
+  const close = () => {
+    if(!chan.__closed__) {
+      chan.close()
+      unsubscribe()
+    }
+  }
   const unsubscribe = subscribe(input => {
     if(isEnd(input)) {
-      chan.close()
+      close()
       return
     }
     if(matcher && !matcher(input)) {
@@ -142,12 +148,7 @@ export function eventChannel(subscribe, buffer = buffers.none(), matcher) {
   return {
     take: chan.take,
     flush: chan.flush,
-    close: () => {
-      if(!chan.__closed__) {
-        chan.close()
-        unsubscribe()
-      }
-    }
+    close,
   }
 }
 
--- a/src/internal/channel.js
+++ b/src/internal/channel.js
@@ -127,7 +127,9 @@ export function eventChannel(subscribe, buffer = buffers.none(), matcher) {
   const close = () => {
     if(!chan.__closed__) {
       chan.close()
-      unsubscribe()
+      if (unsubscribe) {
+        unsubscribe()
+      }
     }
   }
   const unsubscribe = subscribe(input => {
@@ -140,6 +142,9 @@ export function eventChannel(subscribe, buffer = buffers.none(), matcher) {
     }
     chan.put(input)
   })
+  if (chan.__closed__) {
+    unsubscribe()
+  }
 
   if(!is.func(unsubscribe)) {
     throw new Error('in eventChannel: subscribe should return a function to unsubscribe')
@@ -148,7 +153,7 @@ export function eventChannel(subscribe, buffer = buffers.none(), matcher) {
   return {
     take: chan.take,
     flush: chan.flush,
-    close,
+    close
   }
 }
 
--- a/src/internal/channel.js
+++ b/src/internal/channel.js
@@ -126,10 +126,10 @@ export function eventChannel(subscribe, buffer = buffers.none(), matcher) {
   const chan = channel(buffer)
   const close = () => {
     if(!chan.__closed__) {
-      chan.close()
       if (unsubscribe) {
         unsubscribe()
       }
+      chan.close()
     }
   }
   const unsubscribe = subscribe(input => {
--- a/test/channel.js
+++ b/test/channel.js
@@ -250,10 +250,8 @@ test('unsubscribe event channel', assert => {
     }
   });
   chan.take(input => {
-    // console.log(input, unsubscribed);
     assert.equal(input, END, 'should emit END event');
-    // assert.ok(unsubscribed, 'eventChannel should call unsubscribe when END event is emitted asynchronously');
-    setTimeout(() => assert.ok(unsubscribed, 'eventChannel should call unsubscribe when END event is emitted asynchronously'), 0);
+    assert.ok(unsubscribed, 'eventChannel should call unsubscribe when END event is emitted asynchronously');
     assert.end();
   });
 });
--- a/test/channel.js
+++ b/test/channel.js
@@ -202,42 +202,8 @@ test('event channel', assert => {
 
   assert.ok(unsubscribeErr, 'eventChannel should throw if subscriber does not return a function to unsubscribe')
 
-  let unsubscribeWasCalled = false;
-  let chan = eventChannel(() => () => {
-    unsubscribeWasCalled = true;
-  });
-  chan.close();
-  assert.equal(unsubscribeWasCalled, true, 'eventChannel should call unsubscribe when channel is closed')
-
-  unsubscribeWasCalled = false;
-  chan = eventChannel((emitter) => {
-    emitter(END);
-    return () => {
-      unsubscribeWasCalled = true;
-    }
-  });
-  assert.equal(unsubscribeWasCalled, true, 'eventChannel should call unsubscribe when END event is emitted');
-
-  unsubscribeWasCalled = false;
-  let milliseconds = 2;
-  chan = eventChannel((emitter) => {
-    const interval = setInterval(() => {
-      milliseconds -= 1;
-      if (milliseconds > 0) {
-        emitter(milliseconds);
-      } else {
-        emitter(END);
-      }
-    }, 1);
-    return () => {
-      clearInterval(interval);
-      unsubscribeWasCalled = true;
-    };
-  });
-  setTimeout(() => assert.equal(unsubscribeWasCalled, true, 'complex eventChannel should call unsubscribe when END event is emitted'), 5);
-
   const em = emitter()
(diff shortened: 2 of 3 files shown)


Andarist:
We do not have sinon dependency, and probably not worth including it now.

I would recommend to just make some let variable in the closure, so it can be set in `unsubscribe`. Create eventChannel, emit `END` there and check if that variable has changed.

Great PR and the finding!

baldwmic:
I added 2 tests:
- one to test that unsubscribe function gets called when channel is closed
- another test to check that unsubscribe function gets called when END event is emitted

However, I'm having a problem with [this test](https://github.com/redux-saga/redux-saga/pull/877/files#diff-ec2106b885fcd06fe9f0a0ff69c86d41R219).
```
TypeError: unsubscribe is not a function
```
I think it's because [this code](https://github.com/redux-saga/redux-saga/pull/877/files#diff-013b38ba49b148310bcd0a4a810880a1R133) is when the function is defined, and it gets called [here](https://github.com/redux-saga/redux-saga/pull/877/files#diff-013b38ba49b148310bcd0a4a810880a1R130) which is before the function is defined.

Looking at the code, the subscribe function returns the unsubscribe function. And the unsubscribe function calls the close function, which tries to call the unsubscribe function. This doesn't seem possible as far as I understand about JavaScript. Perhaps we need to revamp this.

Andarist:
Interesting case. Probably if you do `emitter(END)` in next tick (i.e. `setTimeout(() => emitter(END), 0)` it should work as `unsubscribe` will be defined by then. Could u try it?

But also its quite interesting case especially in terms of something like `BehaviourSubject` from `rxjs`.
```javascript
eventChannel((emitter) => {
	const unsubscribe = new Rx.BehaviorSubject('smth').subscribe(emitter); 
	emitter(END);
	return unsubscribe;
});
```
this `emitter` in `subscribe` will get called sync for the first time as `BehaviourSubject` is holding its last value (`'smth'` in this case). Its quite a circular dependency but could probably be tackled a little bit in the lib.

Just make sure that when closing via `END` `unsubscribe` is called only if its already defined + check after returning `unsubcribe` if channel is closed and if it is then call immediately `unsubscribe`.

baldwmic:
I tried the suggestion and changed [this line](https://github.com/redux-saga/redux-saga/pull/877/files#diff-ec2106b885fcd06fe9f0a0ff69c86d41R214) to `setTimeout(() => emitter(END), 0);`, which meant the `unsubscribe` function was defined by that time. But this means the test itself needs to be wrapped in a setTimeout. Is that ok?

baldwmic:
Ok @Andarist I've made the suggested changes. I even added a [test case here](https://github.com/redux-saga/redux-saga/pull/877/files#diff-ec2106b885fcd06fe9f0a0ff69c86d41R237) for the eventChannel defined in the docs [here](https://github.com/redux-saga/redux-saga/blob/37f519ca1ab4041e7b571d7e172705ff307c1029/docs/advanced/Channels.md#using-the-eventchannel-factory-to-connect-to-external-events).

The one odd part is that I'm using a setTimeout in the test since the eventChannel uses a setInterval. Is there a better way to do this in mocha?

test/channel.js
```diff
@@ -202,8 +202,42 @@ test('event channel', assert => {
 
   assert.ok(unsubscribeErr, 'eventChannel should throw if subscriber does not return a function to unsubscribe')
 
+  let unsubscribeWasCalled = false;
```
Andarist: maybe a better name would be a shorter `unsubscribed`?

test/channel.js
```diff
@@ -202,8 +202,42 @@ test('event channel', assert => {
...
+      } else {
+        emitter(END);
+      }
+    }, 1);
+    return () => {
+      clearInterval(interval);
+      unsubscribeWasCalled = true;
+    };
+  });
+  setTimeout(() => assert.equal(unsubscribeWasCalled, true, 'complex eventChannel should call unsubscribe when END event is emitted'), 5);
```
Andarist: probably this wont even be checked as this test is synchronously going further till `assert.end()`, I would recomment splitting this whole test in some smaller ones and then in more scoped test call `assert.equal` (like you did) + `assert.end()`

test/channel.js
```diff
@@ -202,8 +202,42 @@ test('event channel', assert => {
...
+    return () => {
+      unsubscribeWasCalled = true;
+    }
+  });
+  assert.equal(unsubscribeWasCalled, true, 'eventChannel should call unsubscribe when END event is emitted');
+
+  unsubscribeWasCalled = false;
+  let milliseconds = 2;
+  chan = eventChannel((emitter) => {
+    const interval = setInterval(() => {
```
Andarist: this could be rather simplified, all of this logic (complexity) boils down to the fact that `END` is emitted asynchronously, so simple `setTimeout(() => emitter(END), 0)` should be enough

+ ofc propert `assert.end()` should accompany this

Andarist:
As to more complex time dependant scenarios I've used before such test utility - https://github.com/baldwmic/redux-saga/blob/6516de097870805aaed0d6e75c64063a69d961f5/test/sagaHelpers/throttle.js#L9 to control time passing.

However here I think we could make this test independant of others and use something like
```javascript
chan.take(input => {
  assert.equal(...) // should be END here
  assert.end()
})
```

that way we do not have to hack the global clock and still the test should run till the completion just right

test/channel.js
baldwmic: - [ ] change name to `unsubscribed`

Andarist:
@baldwmic is the PR finished on ur side? should i review it again and possibly merge in? 

thanks for great work!

test/channel.js
baldwmic: - [ ] split up tests

test/channel.js
baldwmic: - [ ] simplify asynchronous test

baldwmic:
@Andarist i still need to finish updating the tests according to your last review. i expect to finish by the end of today. thank you for all the help in explaining things! 😄

test/channel.js
```diff
@@ -225,6 +225,39 @@ test('event channel', assert => {
...
+  chan = eventChannel((emitter) => {
+    setTimeout(() => emitter(END), 0);
+    return () => {
+      unsubscribed = true;
+    }
+  });
+  chan.take(input => {
+    // console.log(input, unsubscribed);
+    assert.equal(input, END, 'should emit END event');
+    // assert.ok(unsubscribed, 'eventChannel should call unsubscribe when END event is emitted asynchronously');
```
baldwmic: i can't get this test to pass without using setTimeout, any thoughts @Andarist ?

Andarist: Fixed it on your branch, take a look at [this commit](https://github.com/baldwmic/redux-saga/commit/5a9108066c36728090c23364acc14a7e7b9f8137). 

I'll review the whole thing once more later and gonna merge it in. Thanks for the great work and noticing this.

src/internal/channel.js
```diff
@@ -124,16 +124,27 @@ export function eventChannel(subscribe, buffer = buffers.none(), matcher) {
   }
 
   const chan = channel(buffer)
+  const close = () => {
+    if(!chan.__closed__) {
+      if (unsubscribe) {
+        unsubscribe()
+      }
+      chan.close()
```
baldwmic: ahh, so we typically unsubscribe before closing the channel. thank you for the fix @Andarist 👍

Andarist: ive debugged it to the moment ive realised that this was caused by this ordering `unsubscribing` vs `chan.close` and I think its reasonable anyway to have this fixed ordering of things, it seems... better, dont you think? i couldnt find any argument against this

Andarist:
Looks great, when I get back home I'm gonna add one more test (code mentioned [here](https://github.com/redux-saga/redux-saga/pull/877#issuecomment-290326332)) and merge it in. Seems reasonable to have real-life like example in the tests.

src/internal/channel.js
baldwmic: yeah, that makes sense to me. i think generally the argument for doing this would be you want to unsubscribe and allow the developer to clean up everything before officially closing the channel.

PR #889, opened 2017-03-30
Replace incorrect function call in tutorial

martensonbj:
At this point in the tutorial we have updated the `sagas.js` file to fire up both sagas at once in the `rootSaga()` generator, but the yield statement still contained a default `helloSaga()` function from one of the earlier examples. This PR updates the code snippet replacing `helloSaga()` with `incrementAsync()`.


Andarist:
Thanks!

PR #890, opened 2017-04-01
Custom Effects - Spike

eyn:
I've been using redux-saga outside of a normal react/redux app - mainly connecting it to zeromq sockets inside node. I've found myself wanting to create custom declarative effects and supply different effect runners to them depending on the environment. For example I'd like to create a get effect which fetches some data from a database. Once you have released the get/setContext methods this could be achieved using those but for my use case I think custom declarative effects would make the code more readable and testable.

I'm sure the code is full of mistakes etc as I've not hacked around in redux-saga before but hopefully can start as a discussion as to how this could work and if its appropriate to go into redux-saga.


eyn:
Reflected on original approach and have simplified and tidied up a bit.

Have exported new API:
- effect(type, payload) which returns an effect (effect[IO] is true)
- payload(type, effect) which returns the effect's payload
- customEffectRunner added as an option to runSaga/middleware - function should return an effect to run

Code in test case feels much tidier from original spike.

Andarist:
Hey, thanks for the contribution and starting this discussion. I need to analyze the submitted code more before commenting too much, but from what I see you have gone kinda in runtime-agnostic direction. Would be cool if custom effects could use some internal (for now) so they could have more control about the timing of effect resolution.

Could you post example usage of something more complicated than the example in tests? Would like to see the proposed API in action

cc @yelouafi

yelouafi:
I did a quick review (so I may be missing something). from what I understand custom effect runners can only transform from a custom effect to a 'standard' one. But couldn't this be already achieved with a more direct method?

```js
const customEffect = (paylaod) => put({type: 'CUSTOM', payload})

function* saga() {
  yield customEffect(payload)
}
```

Andarist:
@yelouafi what about the idea to expose resolver (`currCb` or similar) to the custom effect runner and possibly some more runtime (`proc`'s) stuff to it? it kinda overlaps with https://github.com/redux-saga/redux-saga/pull/729 though, only the approach is different (middleware chain in the other PR)

yelouafi:
> what about the idea to expose resolver (currCb or similar) to the custom effect runner and possibly some more runtime (proc's) stuff to it?

that might be interesting, although I can't actually think of a use case besides integrated testing

Andarist:
@yelouafi well, most of the 'custom effects' are rather easily implemented as higher order sagas, but there are rare occasions when its not possible - i.e. context effects, maybe some mix of spawn/fork behaviours, self-cancellation, even maybe an effect instructing middleware to send websocket message (without need for creating a watcher saga and dispatching an action).

Aint sure if those examples could be solved with such a resolver though. Also exposing the resolver might be just a small piece allowing more sophisticated use cases. In general I thought about exposing more runtime data/structures, aint sure about use cases and API though.

Andarist:
This is somewhat covered with `effectMiddlewares` option now, so I'm going to close this.

Issue #894, opened 2017-04-04
Webpack bundles both redux-saga/es and redux-saga/lib


Ephys:
The current way of accessing redux-saga's effects is by importing `redux-saga/effects.js`, but that file uses `redux-saga/lib`. This causes webpack to bundle both `redux-saga/lib` and `redux-saga/es`.

I didn't see anything in the documentation mentioning this, so I'm assuming it's a bug?

My current workaround is to add the following to my webpack configuration, but it feels unstable

```javascript
{
  resolve: {
    alias: {
      'redux-saga/effects.js': 'redux-saga/es/effects.js'
      'redux-saga/util.js': 'redux-saga/es/util.js'
    }
  }
}
```

Thank you for your help :)

Andarist:
Whats the undesired effect of the current situation? Duplicated dependencies in the dist file?

From what I understand your alias fix is allowing webpack2 to perform tree shaking and all by using es modules, right?

Ephys:
If by current situation you mean without the alias fix, the undesired effect is having two versions of redux-saga in the same bundle (the contents of `/lib` and `/es`).

Here's the contents of my bundle without the alias:
![screen shot 2017-04-05 at 14 06 00](https://cloud.githubusercontent.com/assets/1280915/24704705/110e11b4-1a09-11e7-8231-090f00dc6326.png)

And wWith the alias:
![screen shot 2017-04-05 at 14 07 37](https://cloud.githubusercontent.com/assets/1280915/24704761/403aef16-1a09-11e7-8edb-2ea644602064.png)

So while the fix does indeed allow webpack2 to perform tree shaking, the bigger benefit is not having two different redux-sagas.

Andarist:
Would you like to collaborate on straightening things? Im not too proficient with bundler configurations, would be cool to work on this with somebody more experienced on the matter. We could schedule some gitter/reactiflux/other_channel time

Ephys:
Actually if you want I can just make a PR with a fix that should make rollup/webpack2 always bundle /es and node/browserify/others always use /lib :)

It's a really minor change https://github.com/Ephys/redux-saga/commit/1ab5bb537279bfb3f1c4e3ecdd065f770d21bccc

Don't worry it doesn't break anything :)

<img width="491" alt="screen shot 2017-04-05 at 17 11 20" src="https://cloud.githubusercontent.com/assets/1280915/24712366/f7b9ee6c-1a22-11e7-80ed-8918fac2c9c7.png">

[...]

PR #895, opened 2017-04-05
Make redux-saga/utils and redux-saga/effects use the right builds

Ephys:
Fixes issue https://github.com/redux-saga/redux-saga/issues/894

It replaces effects.js and utils.js with package.json files that tell node and bundlers where their respective version of the library resides.

It also adds the "module" field in package.json as an alternative to "jsnext:main" (see this for why https://github.com/rollup/rollup/wiki/pkg.module)

### DIFF
--- a/effects.js
+++ b/effects.js
@@ -1,1 +1,0 @@
-module.exports = require('./lib/effects')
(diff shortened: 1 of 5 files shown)


Andarist:
It also should fix #689 . Gonna check it out soon and merge afterwards. Thank you very much

PR #896, opened 2017-04-05
Build umd with rollup

Andarist:
```
redux-saga.js 72K -> 56K
redux-saga.min.js 31K -> 23K
```


yelouafi:
didnt work before with Rollup, but hear about its flat bundles. and that's nice size gain.

Andarist:
what problem did u have with bundling saga with rollup? i didnt test it thoroughly, just that it bundles the lib just fine, also kinda trusted that some other 'big' oss projects switched to it recently and the gain in the file size is obvious

PR #909, opened 2017-04-15
clarified docs concurrency pattern examples (related to https://githu…

epikhighs:
…b.com/redux-saga/redux-saga/issues/883)

The docs and the sample code for concurrency didn't work for my use case b/c of the usage of fork instead of spawn. Thought I'd clarify for others.

Feel free to reword it however you want.

### DIFF
--- a/docs/advanced/Concurrency.md
+++ b/docs/advanced/Concurrency.md
@@ -4,11 +4,14 @@ In the basics section, we saw how to use the helper effects `takeEvery` and `tak
 
 In this section we'll see how those helpers could be implemented using the low-level Effects.
 
+These helpers return a forked task, which creates an attached task that will block execution of the saga that invokes the helper. If you need the helpers to create a detached task, which is what the official redux-saga's implementation of `takeLatest` and `takeEvery` does, then use `spawn` instead. 
+
 ## `takeEvery`
 
 ```javascript
 function* takeEvery(pattern, saga, ...args) {
-  const task = yield fork(function* () {
+    // can optionally use spawn here instead for detached task 
+    const task = yield fork(function* () {
     while (true) {
       const action = yield take(pattern)
       yield fork(saga, ...args.concat(action))
@@ -24,6 +27,7 @@ function* takeEvery(pattern, saga, ...args) {
 
 ```javascript
 function* takeLatest(pattern, saga, ...args) {
+  // can optionally use spawn here instead for detached task
   const task = yield fork(function* () {
     let lastTask
     while (true) {


docs/advanced/Concurrency.md
```diff
@@ -4,11 +4,14 @@ In the basics section, we saw how to use the helper effects `takeEvery` and `tak
 
 In this section we'll see how those helpers could be implemented using the low-level Effects.
 
+These helpers return a forked task, which creates an attached task that will block execution of the saga that invokes the helper. If you need the helpers to create a detached task, which is what the official redux-saga's implementation of `takeLatest` and `takeEvery` does, then use `spawn` instead. 
+
```
Andarist: Thanks for the PR. Could u specify ur original problem? `takeEvery` etc are using `fork` internally, so they are very much attached. This paragraph implies they are not, so maybe there was some other issue u were facing.

epikhighs: in my root saga, if I use the implemntation of takeLatest or takeEvery that uses fork, then it blocks the other yield statements.

```
export default function* rootSaga () {
    yield takeLatest();
    yield takeLatest();// this line will not execute unless spawn() is used
}
```

Andarist: could u specify from which package do you import `takeLatest` and which version of the `redux-saga` do u use?

epikhighs: my import statement is:

```
import {takeEvery, takeLatest} from 'redux-saga';
```

I've had this problem on 
    "redux-saga": "0.14.6",
and
    "redux-saga": "0.14.2",

Andarist:
If u import them from redux-saga/effects the problem will be gone.
Wondering - didnt u have deprecations warnings in the console about the
form u have tried?

docs/advanced/Concurrency.md
epikhighs: sorry, I think I may have confused you.  the takeLatest() used in the sample code I put above was referencing the takeLatest implementation from the documentation found here (https://redux-saga.github.io/redux-saga/docs/advanced/Concurrency.html).  If I use the official redux-saga's takeLatest() or takeEvery(), then it works fine.

epikhighs:
I remember a while ago, that if I tried to import from redux-saga/effects that I could not get the takeLatest/takeEvery import to work.  But now that I try again using the latest redux-saga, it seems to import from redux-saga/effects just fine.  So I"ll use that from now on.  Thanks.

docs/advanced/Concurrency.md
Andarist: I think ive made a mistake there and the correct version should be this
```
function takeLatest(pattern, saga, ...args) {
  return fork(function* () {
    let lastTask
    while (true) {
      const action = yield take(pattern)
      if (lastTask)
        yield cancel(lastTask) // cancel is no-op if the task has already terminated

      lastTask = yield fork(saga, ...args.concat(action))
    }
  })
}
```

Would u like to fix this in the docs? Probably there are like 3 places to correct this when describing `takeEvery`, `takeLatest` and `throttle`

epikhighs: ohhh you took out the yield... i guess that makes sense why it blocked haha.

I'll close this PR and let you handle all the cases where the sample codes need to be adjusted.

Thanks for the help.

Issue #832, opened 2017-02-21
Cannot build gitbook documentation


tmerse:
# Problem:
Trying to build the gitbook documentation does not work for me.

# Steps 

```
git clone https://github.com/redux-saga/redux-saga
cd redux-saga
redux-saga git:(master) gitbook install
Error loading version latest: Error: Cannot find module 'internal/fs'
    at Function.Module._resolveFilename (module.js:470:15)
    at Function.Module._load (module.js:418:25)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at evalmachine.<anonymous>:18:20
    at Object.<anonymous> (/home/tobias/.gitbook/versions/2.5.2/node_modules/graceful-fs/fs.js:11:1)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)

TypeError: Cannot read property 'commands' of null
```

# Installed gitbook version

``` 
redux-saga git:(master) gitbook --version
CLI version: 2.3.0
GitBook version: 2.5.2

redux-saga git:(master) node --version
v7.5.0
redux-saga git:(master) npm --version
4.1.2
```

I just build the redux gitbook and it worked without problems.
I manually bumped the gitbook version in `book.json` to a higher version and repeated the process.
That got me rid of the above error but only to get a new error later on :).

Would be glad for any help, so I can read the docs on my ereader.

Andarist:
Thanks for reporting the problem. I'll try to check whats going on.

Andarist:
Got hopefully fixed by @neighborhood999  . Much thanks to him!

baldwmic:
I'm seeing this problem on another machine of mine, but it appears to be a problem with node: https://github.com/nodejs/node/issues/9377

https://github.com/GitbookIO/gitbook/issues/1774

Andarist:
hm, ive had the same problem, but after incorporating @neighborhood999 fix it was gone. @baldwmic did u test on repo with that fix?

baldwmic:
@Andarist I rebased my fork to be up to date with latest fix and it worked! All good on my end :)

neighborhood999:
@Andarist @baldwmic I'm happy we can solved this issue! 😄

Andarist:
as its resolved, im gonna close this

PR #918, opened 2017-04-19
Fixed gitbook build issue

neighborhood999:
As issue #832 , I had same problem after translate and publish on gh-pages,
So I reference [gitbook document](https://toolchain.gitbook.com/) and try to fix this problem.

Currently works for me, my node and npm version is:

```
node: v7.7.4
npm:  v4.1.2
``` 

Please check this PR, Hopeful we can fix this problem.


SUMMARY.md
```diff
@@ -0,0 +1,28 @@
+* [Read Me](/README.md)
```
Andarist: was adding this neccessary?

Andarist:
Thanks for the fix! working for me too

PR #919, opened 2017-04-19
update createMockTask docs

### LINKED ISSUE #495: CreateMockTask should be documented
I had to look in the source code in order to find that it is possible to mock a task and how to do it.
It will be really helpful if it can be added to documention

baldwmic:
should fix #495 

- adds `createMockTask` to API section
- references TaskCancellation by @henrika 
- fixes import catch by @cellog


Andarist:
Wow, this night we got so many good PRs! Thanks.

PR #920, opened 2017-04-19
Create CNAME

### LINKED ISSUE #731: Put the website at redux-saga.js.org.
Just like [redux.js.org](http://redux.js.org). I think js.org postfix is awsome.  😺

baldwmic:
Fixes #731 

After [reading this](https://dns.js.org/), seems like this is all we need to do. [Redux](https://github.com/reactjs/redux/blob/gh-pages/CNAME) does it.

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+  redux-saga.js.org


Andarist:
`redux` has also tweaked their [docs task](https://github.com/reactjs/redux/commit/37e00083d0c57778337141b8e096d4142223f16e). But for now let's roll with this one and see what will happen ;p

baldwmic:
Are you trying it now?

screen shot 2017-04-19 at 7 21 51 am

I missed a [final step here](https://github.com/js-org/dns.js.org#step-4). Going to submit a PR to them, assuming so DNS will work.

Andarist:
Thanks for making a PR there, was just about to do it too :) also the mentioned copying `CNAME` to `_book`  is needed I think and Ive added it on the master

Issue #888, opened 2017-03-30
`Max call stack exceeded` / `Out of stack space` when using lots of sagas


wilsaj:
We have an app that is making heavy use of sagas. With versions `0.13.0` and up we sometimes run into max call stack errors due to the recursive scheduler and having too many things on the queue. The exact error depends on the browser and circumstance (for example: "Maximum call stack size exceeded" or "Out of stack space") . There is a PR to fix this and a lot of relevant discussion in that PR: https://github.com/redux-saga/redux-saga/pull/790

Our workaround so far has been to revert and pin to `redux-saga@0.12.1`

Opening this as a separate issue here because:
- It will be easier to find when searching for the error. I happened to stumble upon the PR via google search after not finding any GH issues about it.
- This issue can be a place for discussion about things like workarounds that are unrelated to that PR itself
- If for some reason that PR doesn't work out then the issue is still valid

Andarist:
[This](https://github.com/redux-saga/redux-saga/pull/790#issuecomment-287302932) is PR attempting to solve this issue, however author abandoned it due to lack of time. Would you be willing to give it a shot? Not much there to finish the work. Would be happy to merge it in asap when its finished.

wilsaj:
Hi @Andarist - I've got some time to work on this now. Logistically, how should that work - open a new PR?

Andarist:
@wilsaj I think the best way is to:
```sh
git remote add pbadenski https://github.com/pbadenski/redux-saga.git
git fetch pbadenski
git checkout pbadenski/master
git checkout -b BRANCH_NAME
```
and start from there

hendywijaya:
Hi Guys,

I'm still getting this issue at `0.15.6`

When utilizing a lot of Sagas, in high traffic / concurrency, my app will break:

```
redux-saga error: uncaught at fetchCategoryAPI
RangeError: Maximum call stack size exceeded
    at /var/ruparupa/frontend/node_modules/redux-saga/lib/internal/middleware.js:68:23
    at /var/ruparupa/frontend/node_modules/redux-saga/lib/internal/middleware.js:72:22
```

It doesn't always break at the same Sagas 

Should I increase my OS limit on the stack size ? I doubt that will answer the issue.

it seems to point to this part of the redux-saga code

```
return function (next) {
      return function (action) {
        if (sagaMonitor && sagaMonitor.actionDispatched) {
          sagaMonitor.actionDispatched(action);
        }
[...]

PR #923, opened 2017-04-19
Fix: Under certain conditions scheduler may crash with "Maximum call stack size exceeded" error (pt. 2)

wilsaj:
This is a followup/continuation of https://github.com/redux-saga/redux-saga/pull/790 to address https://github.com/redux-saga/redux-saga/issues/888

There is a good description of the issue and discussion in the links above, but the short version is that `exec()` and `flush()` have a recursive relationship as a way of iterating through the scheduler's queue, and that errors out at a certain recursion depth. So the fix is to change things so we loop through the queue in a non-recursive way.

### DIFF
--- a/src/internal/scheduler.js
+++ b/src/internal/scheduler.js
@@ -15,12 +15,18 @@ let semaphore = 0
   state).
 **/
 function exec(task) {
-  try {
-    suspend()
-    task()
-  } finally {
-    flush()
-  }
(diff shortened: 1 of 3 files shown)


Andarist:
Seems good, tests are passing 🎉 Thank you for the contribution, great work! We'll release a new `redux-saga` version in a moment

PR #929, opened 2017-04-21
Fix type definitions for greater type safety.

dcolthorp:
Hello!

I'm new to redux-saga and trying it for the first time with TypeScript. I hope this PR is helpful, but please let me know if anything should be changed.

It looks like there are a few issues with the type definitions that keep typescript from making good use of the types to detect user errors (or I'm just missing something).

In particular:

* `SagaRest` is compatible with e.g. `Saga1`, such that typescript will fall back on it if the argument to a `Saga1` to `run` does not match, thereby not flagging an error.
* The same goes for `call` – the `...args` case is not mutually exclusive with the prior cases, so typescript will not flag argument type errors.
* Same thing with apply.

This PR addresses the first two cases, but I couldn't figure out a way to resolve `apply`.

Side note about apply: I suspect it can't be fixed. If that is the case, it's a little surprising to see a static type for `apply` in a correct case, without any hint that this will not catch errors statically due to the fallback to the `any[]` case. Perhaps apply should not pretend to type its args and just use any?

## Example

I discovered the issue with my very first exploration of redux saga in Typescript:

```ts
function foo(x:number) : string {
  return null as any;
}

export function* rootSaga() : SagaIterator {
  const x : string = yield call(foo, 3)
}
```

Inspecting the type of `call` to see which type overload was selected, VSCode shows me

```ts
(alias) call<number>(fn: ((arg1: number) => any) | [any, (arg1: number) => any] | {
    context: any;
    fn: (arg1: number) => any;
}, arg1: number): CallEffect (+5 overloads)
import call
```

Great! Just as hoped, it knows that my function takes an argument that my argument is a number. 

I then purposefully introduced a type error, by changing argument from a number, to a string:

```ts
  const x : string = yield call(foo, 'bar')
```

Much to my surprise, this did not introduce a type error! The type inferred for `call` in this case was:

```ts
(alias) call(fn: ((...args: any[]) => any) | [any, (...args: any[]) => any] | {
    context: any;
    fn: (...args: any[]) => any;
}, ...args: any[]): CallEffect (+5 overloads)
import call
```

With the updated definitions in this PR, I get the following type error, as desired:

```
The type argument for type parameter 'T1' cannot be inferred from the usage. Consider specifying the type arguments explicitly.
  Type argument candidate 'number' is not a valid type argument because it is not a supertype of candidate '"3"'.
(alias) call<T1>(fn: ((arg1: T1) => any) | [any, (arg1: T1) => any] | {
    context: any;
```

### DIFF
--- a/effects.d.ts
+++ b/effects.d.ts
@@ -60,7 +60,9 @@ type CallFunc4<T1, T2, T3, T4> = (arg1: T1, arg2: T2, arg3: T3,
                                   arg4: T4) => any;
 type CallFunc5<T1, T2, T3, T4, T5> = (arg1: T1, arg2: T2, arg3: T3,
                                       arg4: T4, arg5: T5) => any;
-type CallFuncRest = (...args: any[]) => any;
+type CallFuncRest<T1,T2,T3,T4,T5> = (arg1: T1, arg2: T2, arg3: T3,
+                                     arg4: T4, arg5: T5,
+                                     ...args: any[]) => any;
 
 type CallEffectArg<F> = F | [any, F] | {context: any, fn: F};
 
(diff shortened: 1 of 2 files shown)


aikoven:
Please have a look at https://github.com/redux-saga/redux-saga/pull/740

dcolthorp:
Oh, thanks! Looks like my searching was too narrowly focused on `call`.

PR #930, opened 2017-04-21
Begginer Tutorial: Fix wrong saga name in snippet

ravicious:
The code executes `incrementAsync` and `watchIncrementAsync`, but the docs say:

> This Saga yields an array with the results of calling our two sagas,
> `helloSaga` and `watchIncrementAsync`.


Andarist:
Thanks!

PR #935, opened 2017-04-26
Add a link on ES6 generators and 2 on redux-saga to ExternalResources

goshakkk:
First off, thanks for all the effort that went into redux-saga! It certainly did make my day as a React developer more manageable.

I just realized the three particular posts that I've authored some time ago might be relevant to mention on the External Resources page... which I hope does not come off as blatant self-promotion. I'm not in it for the page views, just want to spread the relevant knowledge, which would have been definitely helpful to me when I was just starting out with sagas.

The post on understanding ES6 generators is my third most visited blog post in the past year. It describes the way I think of ES6 generators as "interpreter—program", and redux-saga fits perfectly in this model.

Then there's one comparing the *approach* redux-saga to other alternatives.

The last one demonstrates how sagas make possible a common UI pattern of lazy registration.

If some of the links — or all of them — don't fit this page, feel free to reject the PR.


Andarist:
Great articles, thanks for the input. Noone will blame you here for self promotion ;) Content of those is rock solid

PR #936, opened 2017-04-26
Enable runSaga to subscribe to a channel

Zacqary:
## What this PR does
This allows `runSaga` to be passed a Channel instead of a function as its `subscribe` parameter.

## Motivation
We're using something like this at Datadog where we have our frontend telemetry process running as a separate Saga process apart from our Redux-connected Sagas (both for code separation and ease of reuse in non-Redux code). The channel interface makes it a lot easier to send messages to this process from both other sagas and non-saga code, but we had to write a hack like this to get it to work:
```javascript
let subscribers = new Set();
const subscribe = (callback) => {
        subscribers.add(callback);
        return () => subscribers.delete(callback);
};
const telemetryChannel = channel();
const watchChannel = () => telemetryChannel.take(payload => {
        subscribers.forEach(subscriber => subscriber(payload));
        watchChannel();
})();
runSaga(telemetrySaga(), { subscribe })
```

With this PR, we'd be able to just:
```javascript
const telemetryChannel = channel();
runSaga(telemetrySaga(), { subscribe: telemetryChannel });
```

I elected not to modify the `dispatch` property in this PR, because it's easy to just pass `{ dispatch: myChannel.put }`.

## Alternatives
Is overloading `subscribe` a good idea for this behavior? Or, should `runSaga` take a `channel` property instead? 

My concern with that approach would be for a case where you want to subscribe to one channel but dispatch to something else — passing `runSaga(mySaga(), { channel: myChannel })` seems to imply both `take`ing and `put`ting to that channel.

### DIFF
--- a/src/internal/channel.js
+++ b/src/internal/channel.js
@@ -157,14 +157,16 @@ export function eventChannel(subscribe, buffer = buffers.none(), matcher) {
   }
 }
 
-export function stdChannel(subscribe) {
-  const chan = eventChannel(cb => subscribe(input => {
-    if (input[SAGA_ACTION]) {
-      cb(input)
-      return
-    }
-    asap(() => cb(input))
-  }))
+export function stdChannel(subscribeFnOrChannel) {
+  const chan =
+      is.channel(subscribeFnOrChannel) ? subscribeFnOrChannel
+    : eventChannel(cb => subscribeFnOrChannel(input => {
+      if (input[SAGA_ACTION]) {
+        cb(input)
+        return
+      }
+      asap(() => cb(input))
+    }))
 
   return {
     ...chan,
--- a/test/runSaga.js
+++ b/test/runSaga.js
@@ -2,7 +2,7 @@ import test from 'tape'
 
 import { runSaga } from '../src'
 import { fork, take, put, select, all } from '../src/effects'
-import { emitter } from '../src/internal/channel'
+import { emitter, channel } from '../src/internal/channel'
 import { runSyncDispatchTest } from './scheduler'
 
 function storeLike(reducer, state) {
@@ -70,4 +70,37 @@ test('put causing sync dispatch response in store-like subscriber', assert => {
   const store = storeLike(reducer, {})
 
   runSyncDispatchTest(assert, store, (saga) => runSaga(saga(), store))
+});
+
+test('runSaga subscribing  to a channel', assert => {
+     assert.plan(1)
+
+    let actual = []
+    const testChannel = channel()
+    const task = runSaga(root(), { subscribe: testChannel })
+
+    testChannel.put({type: 'ACTION-1'})
+    testChannel.put({type: 'ACTION-2'})
+
+    function* root() {
+      yield [fork(fnA)]
+    }
+
+    function* fnA() {
+      actual.push( yield take('ACTION-1') )
+      actual.push( yield take('ACTION-2') )
+    }
+
+    const expected = [
+      {type: 'ACTION-1'},
+      {type: 'ACTION-2'}
+    ]
+
+    task.done.then(() =>
+      assert.deepEqual(actual, expected,
+        'runSaga must connect the provided iterator to the channel, and run it'
+      )
+    )
+
+    task.done.catch(err => assert.fail(err))
 });


Andarist:
hey @Zacqary , you have found a nice use case here and personally I love what you guys do at Datadog, great product. 

From what I see you basically need a multicast channel and as we do not expose such (yet) you need to make this hackery with `subscribers`. Am I reasoning correctly?

If yes then I think such API `runSaga(mySaga(), { channel: multicastChannel })` would fit perfectly into your use case and would ofc be more explicit about what it actually does.

Actually I have 2 PRs open and one of them could potentially impact your proposal. 

First is - https://github.com/redux-saga/redux-saga/pull/755 and it switches the arguments order for `runSaga` API. Will merge this one today (I think), so rebasing your work later would be appreciated to reflect the changes.

More interesting one is this - https://github.com/redux-saga/redux-saga/pull/824 . Its not finished as I couldnt find enough time lately on refactoring this but the idea behind it is exactly what you need - multicast channel (I think). Would you like to colaborate on this one maybe? If thats something your goal aligns to. Beside making proposed API possible its goals are:
- perf boost for `sagaMiddleware.run` API (one shareable `stdChannel` for internal usage, now each task spawns its own `stdChannel`)
- exposing `multicastChannel` as primitive for people's usage

Zacqary:
Rebased my branch for #755 

I'm not sure I understand the multicast channel PR enough to know how it's relevant to my use case. My goal here was just to be able to run a Saga that `take`s directly from a channel instead of a Redux store, because I like the ability to use `take` and `put` everywhere. 

The point of the `subscribers` hack isn't that the channel is broadcasting to multiple subscribers, it's that it's broadcasting to a `runSaga`-compatible `subscribe` function. I'm not actually sure if there needs to be a whole `Set` of subscribers or if I only need one; I just wrote the hack this way to be safe.

So, admittedly, this is all coming from a place of ignorance about how channels actually work under the hood. If you say `multicastChannel` is the solution here then I'm inclined to believe you. But I'm a little worried reading the discussions on channels vs multicastChannels that it could become confusing. I guess the difference is a `channel` can buffer but a `multicastChannel` doesn't? That sounds to me as if `channel` is now the more "advanced" concept (I still haven't actually used buffering in a project yet), but the longer, scarier name `multicastChannel` represents something simpler.

Andarist:
> Rebased my branch for #755

Actually this got into master and is already released in `v0.15` so probably should be rebased against master

> So, admittedly, this is all coming from a place of ignorance about how channels actually work under the hood.

You can look into the source code - they are really easy, just exposing put/take/close interface which is leveraged by a runtime when you yield effects + they can buffer stuff, so your items wont get lost while you process other one.

> If you say multicastChannel is the solution here then I'm inclined to believe you. But I'm a little worried reading the discussions on channels vs multicastChannels that it could become confusing. I guess the difference is a channel can buffer but a multicastChannel doesn't? That sounds to me as if channel is now the more "advanced" concept (I still haven't actually used buffering in a project yet), but the longer, scarier name multicastChannel represents something simpler.

I *think* it should cover ur use case + it would provide a nice integration point for others too, but im not 100% sure, so correct me if Im wrong.

Idea is that a channel is now a unicast - that means that if some item shows up in it only 1 taker can consumer it, even if there are multiple competing takers. As its the only type of channel now `stdChannel` is implemented as one. Its just an abstraction channel which subscribes to the store's dispatches and proxy them to your saga, internally `take('ACTION')` is subscribing to this `stdChannel`. Problem is that each saga needs to create its own `stdChannel` (because of unicast behaviour), so its kinda inefficient because we can look at dispatches being the stream of actions which every saga is interested in.

Because of this Ive came with idea to implement a second type of channel - multicast. With same interface ofc, but the behaviour would differ - all pending takers would be able to consume from it at the same time (hence multi). That way we could create a single `stdChannel` on which each saga could subscribe to. Because of the multicast behaviour Im not sure if it makes sense to have buffering option on such a channel yet, it sounds weird and can lead to confusion. Im not against the idea yet, just unsure.

I think the same concept could be applied to ur idea. You just need to pass a channel to the `runSaga` as something which makes integration easier for u and because I would prefer to have more explicit API for this (`{ channel: someChannel }`) and its kinda what normal usage (`sagaMiddleware.run`) also do but in slightly other way (hidden from the user) I think it makes sense to have those unified.

Andarist:
I've [merged](https://github.com/redux-saga/redux-saga/commit/d062f3f4cc85af0f4ab618fe2664f1efeb14c75d) the other PR which adds support for this requested feature. 

The usage will be:
```
import { stdChannel, runSaga } from 'redux-saga'

const channel = stdChannel()
const task = runSaga({
  channel,
  // ...
})

// ... 
channel.put({ type: 'SOME_ACTION' })
```

Sorry it took so long. This feature is not yet released though, I'm working on releasing v1, progress can be check [here](https://github.com/redux-saga/redux-saga/milestone/1). Hopefully you will be able to enjoy this soon. If you are using npm5 you can try it out by specifying redux-saga's version as `"redux-saga/redux-saga#COMMIT_HASH"` - we have `prepare` script in our `package.json`, this means npm5 will be able to download redux-saga's repository contents, install our deps and build it.

PR #937, opened 2017-04-27
docs(readme): Add logo, backer, sponser and update links

JisuPark:
Hello

I've updated [README_ko.md](https://github.com/redux-saga/redux-saga/blob/master/README_ko.md) file based on the [d0aa83d](https://github.com/redux-saga/redux-saga/tree/d0aa83d782b241799264879e139db4d03438ae28).

So the outdated links are updated. And the other missing things are now added.

Thanks,
Jisu Park


Andarist:
Thanks!

Issue #722, opened 2016-12-27
Who uses Redux-Saga?


Andarist:
If your company is using Redux-Saga in production, please let us know who are you so we can mention you in the README. Linking logos would be appreciated too.

anthonychung14:
We are hoping to use Redux-Saga in production. I've got a PR pending with our company, but still some final bugs to squash!

![Brandcast](http://2016.sfdesignweek.org/wp-content/uploads/2016/05/Brandcast_logo.png)

gabrielhpugliese:
We are using in production: https://www.shore.com/de/ (not this page, only on the real webapp)

![https://cdn2.hubspot.net/hubfs/498216/02_Icons/shore_logo.svg?t=1481792698323](https://cdn2.hubspot.net/hubfs/498216/02_Icons/shore_logo.svg?t=1481792698323)

slorber:
stample.co https://stample.co/assets/homepage/img/stample_atom.png

eloytoro:
We do at [mozio](http://www.mozio.com/)


![logo](https://cloud.githubusercontent.com/assets/3936773/21569352/b1fdc098-ce81-11e6-83ac-8682bc7aa4e7.png)

kuy:
We're heavily using Redux Saga in [entrepedia](https://entrepedia.jp/pages/venture_maps).

![screen shot 0028-12-31 at 2 32 30](https://cloud.githubusercontent.com/assets/151614/21569588/752ebe08-cf01-11e6-99c1-28e6e1667f4b.png)

cheersjosh:
We use Redux Saga throughout our entire application at [Pozible](https://pozible.com) - Australia's largest crowdfunding platform.

![logo](https://cloud.githubusercontent.com/assets/7821982/21599332/12cb2268-d1bf-11e6-8b18-13a378cf5820.png)

eliseumds:
We do at [ProductReview.com.au](https://www.productreview.com.au), Australia's first and most comprehensive consumer opinion site. We'll be releasing the new version of our public-facing app in a few months.

Some of our use cases:

* Analytics
* Authentication
* A very flexible form builder to create specifications for our listings (ex: product model, color, weight, engine size). Great work from @danialk
* User activity tracking
* Files uploader ([see GIF](https://cloud.githubusercontent.com/assets/640840/22722980/cdafea26-ee0e-11e6-93c7-8405efb8c4e5.gif))
[...]

PR #944, opened 2017-04-29
Who Uses Redux-Saga

baldwmic:
- starts #722 
- you can see how it looks [here](https://github.com/baldwmic/redux-saga/tree/whoUsesReduxSaga)
- open to design feedback

Suggestions:
- [ ] logos are all same height and width
- [ ] multiple users (logo and name) on a row

screen shot 2017-04-29 at 11 39 54 am


Andarist:
Thanks, this looks good. They should be in rows and centered (possibly), but you already know that :)

baldwmic:
gonna close this since i'm not actively working on it. anyone is free to pick this up

Issue #614, opened 2016-11-04
Omit "saga has been cancelled" printout in browser console


williamgranli:
Is there any way to disable this? I'm triggering it manually with 'yield cancel(saga)' so I don't see the point of it being printed by default... 

Thanks for any help!

Andarist:
It's `console.info`, so you can easily filter it out, also its `dev` only

aikoven:
IMO it's still not very convenient. Cancellation is a regular situation, nothing like uncaught exceptions, so I don't think it should be logged. Also you can actually see it in Saga Monitor if you need.

williamgranli:
@Andarist I understand that, but I think there should be an option to not log cancellations, since it's a common practice as @aikoven suggests. Especially since the biggest(?) react boilerplate project (https://github.com/mxstbr/react-boilerplate) uses saga injections which means that sagas have to be cancelled on every route change. 

Wouldn't it be better to f.e. send in an extra param to the cancel like so:
> yield cancel(sagaName, {log: true});

slorber:
I agree that in my project it creates some useless logging statements.

However they are still useful when something is going wrong and you want to debug sagas without plugging the monitor :)

yelouafi:
should we make the logging opt-in?

aikoven:
It would be nice if this could also be configured per generator. 

Also, not sure if it was proposed already, it'd be great if we could supply a custom `displayName` for generators, different from `generator.name`, to be shown in these log messages as well as saga monitor. Would be useful for higher order generators. Setting `Function.name` doesn't work on some browsers.

Andarist:
Sorry that it took so long to act on this in any way.

What I would propose for now is - console.log only non-cancel cancellations (what a mumbo jumbo). What I mean is that we can distinguish why the cancelation has happened - is it user induced via `cancel` effect (no logging should take place) or is it cancel caused by cancel propagation when something throws (logging should take place in dev mode).

@aikoven 
I like your `displayName` idea, will try to explore this. Would be also useful for overriding display name for helper effects like `takeEvery` as its implemented as mere `fork(takeEveryHelper)`, would be cool to omit `fork` part out of this.

aikoven:
> or is it cancel caused by cancel propagation when something throws

+ for this, we could also group these messages to more clearly indicate where the error originated and to which cancellations it lead.

> Would be also useful for overriding display name for helper effects like takeEvery

[...]

PR #945, opened 2017-04-29
Removed Log for Manual Cancel

baldwmic:
- Starts #614 

Spec:
- if it's user induced via cancel effect -> no logging should take place
- is it cancel caused by cancel propagation when something throws -> logging should take place in dev mode

Question:
- what I'm not sure is if moving the log statement to the `else` statement where `isErr` is true accomplishes the second bullet point for the "Spec". thoughts @Andarist ?

### DIFF
--- a/src/internal/proc.js
+++ b/src/internal/proc.js
@@ -279,12 +279,12 @@ export default function proc(
     iterator._isRunning = false
     stdChannel.close()
     if(!isErr) {
-      if(process.env.NODE_ENV === 'development' && result === TASK_CANCEL) {
-        log('info', `${name} has been cancelled`, '')
-      }
       iterator._result = result
       iterator._deferredEnd && iterator._deferredEnd.resolve(result)
     } else {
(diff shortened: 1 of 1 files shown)


Andarist:
Truth to be told - I have no idea. Verifying various scenarios should be quite simple and probably the best way to have certainty how this works. You can write tests first according to the spec you have listed and then see whats failing.

bradennapier:
Any update on this? I get tons of these whenever im hot reloading the sagas it fills the window.

bradennapier:
I don't think logs should ever be forced upon a user in any situation unless they are opted into.  Instead there should simple be a setting when configuring the middleware to opt-into debug information.

Andarist:
@bradennapier would u like to work on a PR for that? unfortunately I dont have much time to do this atm

axelson:
Can this at least be changed to a verbose-level while waiting for a fuller fix? That would make it easier to filter out.

baldwmic:
closing in favor of #1054

Issue #931, opened 2017-04-24
Can i make a repository for translated docs?


mskims:
Hello guys, I'm studying redux libraries now and I'm very interested in `redux-saga`.

As a native Korean speaker, I've tried to find saga docs written in Korean, but what I have found is just one readme paper translation at [this](https://github.com/redux-saga/redux-saga/blob/master/README_ko.md)

if you guys don't mind, I want to write a Korean translation docs like [Chinese did!](https://github.com/superRaytin/redux-saga-in-chinese)

Thank you for contributing your awesome repository!

baldwmic:
@mskims that would be amazing! I'm sure other Korean speakers would be most grateful to you if you did. I'm not on the redux-saga team, but I'm sure @Andarist would approve as there's already some other translations

Andarist:
Yeah, sure - I can merge in. However not a big fan of those as they tend to become outdated quite fast.

PR #954, opened 2017-05-03
Add link for Korean translation docs 

mskims:
#931 

Even though translation repository can become outdated quite fast, we're gonna keep tracking and translating.

Thank you for translators and contributors 👍


Andarist:
Thanks!

PR #955, opened 2017-05-03
fixing issue on transpilation (more than one array symbol in proc.js)

sithmel:
Importing module from es/index.js I get this error:
```
ERROR in ./~/redux-saga/es/internal/proc.js
Module parse failed: /home/sithmel/src/supply/app-timesheet/node_modules/redux-saga/es/internal/proc.js Identifier 'array' has already been declared (35:18)
You may need an appropriate loader to handle this file type.
|     };
|   },
|   array: function array(patterns) {
|     return function (input) {
|       return patterns.some(function (p) {
 @ ./~/redux-saga/es/utils.js 3:0-46
 @ ./~/redux-saga/es/index.js
 @ ./src/ui/components/index.jsx
 @ ./src/ui/index.js
``` 
When I opened the file I noticed the transpiler added a function named array to the file:
```
var matchers = {
  ...
  array: function array(patterns) {
```
and it conflicts with the import:
```
import { noop, ...  array ...  } from './utils';
```

### DIFF
--- a/src/internal/proc.js
+++ b/src/internal/proc.js
@@ -1,4 +1,4 @@
-import { noop, kTrue, is, log as _log, check, deferred, uid as nextEffectId, array, remove, object, TASK, CANCEL, SELF_CANCELLATION, makeIterator, createSetContextWarning, deprecate, updateIncentive } from './utils'
+import { noop, kTrue, is, log as _log, check, deferred, uid as nextEffectId, array as arrayLib, remove, object, TASK, CANCEL, SELF_CANCELLATION, makeIterator, createSetContextWarning, deprecate, updateIncentive } from './utils'
 import { asap, suspend, flush } from './scheduler'
 import { asEffect } from './io'
 import { stdChannel as _stdChannel, eventChannel, isEnd } from './channel'
@@ -556,7 +556,7 @@ export default function proc(
     function checkEffectEnd() {
       if(completedCount === keys.length) {
         completed = true
-        cb(is.array(effects) ? array.from({ ...results, length: keys.length }) : results)
+        cb(is.array(effects) ? arrayLib.from({ ...results, length: keys.length }) : results)
       }
     }
 


Andarist:
What's your build setup?

sithmel:
I am using webpack. This is the babel config:
```
      {
        test: /\.jsx?$/,
        exclude: /node_modules/,
        loader: 'babel-loader',
        query: {
          presets: ['es2015', 'es2017', 'react'],
          plugins: ['transform-runtime', 'transform-decorators-legacy', 'transform-class-properties'],
        },
      },
```

sithmel:
But I think the issue is how babel (your babel ;-) ) transpiles this line https://github.com/redux-saga/redux-saga/blob/master/src/internal/proc.js#L15

Andarist:
Could u create a small repo with reproduced issue? Have tried but everything works fine.

sithmel:
I have tried shortly to reproduce the problem in isolation but I couldn't.

But if you open the folder node-modules/redux-saga/es/internal/proc.js you can notice very easily that there are 2 "array" in the same namespace. One in the import statement and the other added by babel in the name of the function matcher.array.

Also the change is pretty straightforward ;-)

Andarist:
It may be obsolete and for now I'm thinking it actually is. The second identifier should easily cover the one from the upper scope. Are you sure that this error is not caused by your eslint config or smth?

sithmel:
The error was happening running webpack, but I could not reproduce it with a simpler config.
For now we reverted to redux-saga 0.10 and everything is working fine.

Andarist:
Could u fork ur repo, drop its history and erase all files except simplified entry point, config etc? What version of webpack are u using?

sithmel:
are you sure it is necessary ? you can clearly see that the transpiled script is incorrect.

I'll be happy to try reproducing the issue as soon as I have time.

Andarist:
Cant see anything wrong with the transpiled code. I've literally just have checked in Safari (it shipped native module support) this code
```javascript
import { array } from './module.js'

var matchers = {
  array: function array(patterns) {
    return function (input) {
      return patterns.some(function (p) {
        return matcher(p)(input);
      });
    };
  },
}

```
and those identifiers **don't** collide. I believe there is some bug out there, just dont think its on our side. Im happy to help with identifying it, filing issues where it is neccessary and even patching this on our side, but only once we find the root cause, otherwise the issue will be forgotten and overlooked.

sithmel:
That's weird I would have bet it was not valid.
Thanks for double checking it!

Apologies I didn't have time to go in depth yet, but I'll have a look as soon as I can

Andarist:
this `array` identifier in `function array(...` should be visible only from INSIDE this function (so its possible to call it recursively, its invisible from the outside.

Try this
```javascript
var a = { b: function c() { console.log(c) } }
a.b() // function c() { console.log(c) }
console.log(c) // Uncaught ReferenceError: c is not defined
```

gunters63:
Hi,

We have the same problem, but only in the production build.

Could it be a webpack config problem? The only significant difference between development and production config of webpack is this:

      new webpack.optimize.DedupePlugin(),
      new webpack.optimize.UglifyJsPlugin({
         compress: {
            warnings: false
         }
      }),

Babel config is the same for development and production.

gunters63:
No, dedupePlugin is obsolete now, and the uglifier was also not responsible for this.

At the moment it seems our problem is some discrepancy in the local and the CI server packaging process.

Andarist:
Building production build locally is OK? Only when done on the CI server?

sithmel:
I have noticed that it works if I instruct babel to transpile inside node_modules:
```
      {
        test: /\.jsx?$/,
        // exclude: /node_modules/,
        loader: 'babel-loader',
        query: {
          presets: [['es2015', { modules: false }], 'es2017', 'react'],
          plugins: ['transform-runtime', 'transform-decorators-legacy', 'transform-class-properties'],
        },
      },
```
That in my project, I have tried to reproduce in isolation but it does not make any difference.
I am still very confused ...

Andarist:
U could try 'bisecting' (kindof) ur project to produce the minimal project setting which causes the issue.

gunters63:
I think I found a solution. A weird one however.

A co-worker had the same problem. We compared our outputs of npm list and found this:

I had:

+-- webpack@2.5.1
| +-- acorn@5.0.3
| +-- acorn-dynamic-import@2.0.2
| | `-- acorn@4.0.11

He hard:

+-- webpack@2.5.1
| +-- acorn@5.0.3
| +-- acorn-dynamic-import@2.0.2
| | `-- acorn@4.0.12

Funny, there is no known version 4.0.12 of acorn
acorn-dynamic-import@2.0.2 has a dependency of acorn@^4.0.3

When my coworker pinned his acorn version like this in his package.json:

acorn: "4.0.11"

it suddenly worked!

Weird, there is some dependency confusion here, or npm has corrupt modules cached.

sithmel:
I pinned to acorn 4.0.11 and it worked. Thanks @gunters63

gunters63:
Good to hear!

Btw, my coworker and me both have exactly the same version of npm and node (and OS), and we both deleted node_modules and did a fresh npm install.

Andarist:
https://github.com/ternjs/acorn/issues/527

sithmel:
https://github.com/ternjs/acorn/issues/527#issuecomment-291266872

sithmel:
we can close I think. Thanks a lot @gunters63 and @Andarist

PR #956, opened 2017-05-03
Fixed bug in `Retrying XHR calls` recipe

samijaber:
The if statement has a tiny bug that makes it redundant.


Andarist:
Thanks!

PR #957, opened 2017-05-03
Docs update

VictorQueiroz:


Andarist:
Thanks! `context` still needs some more love, but this is a good starting point

PR #958, opened 2017-05-03
fix: always deliver “context” object to runSaga() function

VictorQueiroz:


Andarist:
Thanks, had to miss this during merging.

PR #968, opened 2017-05-05
Docs: fix typo

beelee:
Found a typo in sagas.js:
- all is not a function but call is
- parentheses of the saga-functions in the array are removed

### DIFF
--- a/docs/introduction/BeginnerTutorial.md
+++ b/docs/introduction/BeginnerTutorial.md
@@ -160,9 +160,9 @@ Now we have 2 Sagas, and we need to start them both at once. To do that, we'll a
 ```javascript
 // single entry point to start all Sagas at once
 export default function* rootSaga() {
-  yield all([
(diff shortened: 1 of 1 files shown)


Andarist:
What version of the redux-saga do u use?

beelee:
0.14.8


LG
Bianca

Andarist:
[`all`](https://redux-saga.js.org/docs/api/index.html#alleffects---parallel-effects) was introduced in [v0.15.0](https://github.com/redux-saga/redux-saga/releases/tag/v0.15.0). 

It's basically the same thing as 'parallel' effect - `yield [eff, eff, eff]`. However `all` is preferred way now to handle this as its more explicit and corresponds in naming to the [`Promise.all`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) nicely.

PR #972, opened 2017-05-07
Fix beginner tutorial for "all" effect.

ftripier:
The beginner tutorial makes you sue the "all" effect without importing it.

The actual beginner tutorial repository also needs to be updated to include the "all" effect but I see that there's already a PR for it:

https://github.com/redux-saga/redux-saga-beginner-tutorial/pull/21


Andarist:
Thanks for the PR and mentioning the tutorial. Totally forgot about this one.

Issue #973, opened 2017-05-07
Determine origin of Error from a race


Lucretiel:
Currently I have something like this:

    function* mySaga(work1, work2) {
        let result1, result2
        try {
            { result1, result2 } = yield race({
                result1: call(work1),
                result2: call(work2),
            })
        } catch(err) {
            // ???
            return
        }
        if result1 {
            // ...
        } else {
            // ...
        }
    }
My question is about the `catch` block, with the question marks. Is there any way to determine which of my two workers threw the error? Would it be worth adding an extension option to `race` to make it possible to do so? For instance, a throw/reject error could be wrapped in a single-key object, just like the return value from the race is.

Andarist:
Interesting, we could extend `err` somehow, wouldnt like to wrap it in `{ result1: err }` though as people expect `err` to be an instance of `Error`.

Would u like to propose a semantics? The best I can think off now is to add a property to the error by throwing it in with something like
```javascript
err.originEffect = 'key'
```

Lucretiel:
Ah, yes. My original idea was to replace the error with `{[key]: error}` or `{origin: key, error: error}`, but I can see why people would want to have the error be an `Error` type, especially if exceptions bubble out.

I like the idea of attaching a property to the error. I'll see if I can write up a proposed impl.

I wrote a simple wrapper that implements the same idea, but I'd *much* rather it be an extension to `race`, because of how heavy the wrapper implementation is. It requires wrapping all of your race Effects in a whole `call`, with a whole wrapping generator for each one:

```javascript
export const raceErrOrigin = racers => race(
	mapValues(racers, (racer, key) => call(function*() {
		try { return (yield racer) }
		catch (e) { e.raceOrigin = key; throw e }
	}))
)
```

PR #975, opened 2017-05-07
Added error wrapping options to `race`

Lucretiel:
Added features described in #973

@Andarist

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -37,7 +37,7 @@
   * [`setContext(props)`](#setcontextprops)
   * [`getContext(prop)`](#getcontextprop)
 * [`Effect combinators`](#effect-combinators)
-  * [`race(effects)`](#raceeffects)
+  * [`race(effects, options)`](#raceeffects)
   * [`all([...effects]) (aka parallel effects)`](#alleffects---parallel-effects)
   * [`all(effects)`](#alleffects)
 * [`Interfaces`](#interfaces)
@@ -761,12 +761,15 @@ Creates an effect that instructs the middleware to return a specific property of
 
 ## Effect combinators
 
-### `race(effects)`
+### `race(effects, options)`
 
 Creates an Effect description that instructs the middleware to run a *Race* between
-multiple Effects (this is similar to how [`Promise.race([...])`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise/race) behaves).
+multiple Effects (this is similar to how [`Promise.race([...])`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise/race) behaves). When one effect completes or throws, that result is returned (or thrown) by the race, and all the other effects are cancelled. If the result is not an error, it wrapped in an object of the form `{[key]: result}`; this allows the caller to determine which Effect returned the result
 
-`effects: Object` - a dictionary Object of the form {label: effect, ...}
+- `effects: Object` - a dictionary Object of the form {label: effect, ...}
+- `options: Object` - an optional Object containing options for the race. Currently supported options are:
+  - `wrapErr: boolean` - if true, any error thrown by the racing Effects will be wrapped in an object of the form `{origin: <key>, error: <error>}`. This allows the caller of the race to determine which Effect caused the error.
+  - `annotateErr: boolean` - if true, any error thrown by the racing Effects will have the property `err.raceOrigin = <key>` attached to it. This allows the caller of the race to determine which Effect caused the error, while still ensuring the thrown object is an `Error`.
 
 #### Example
 
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -764,12 +764,11 @@ Creates an effect that instructs the middleware to return a specific property of
 ### `race(effects, options)`
 
 Creates an Effect description that instructs the middleware to run a *Race* between
-multiple Effects (this is similar to how [`Promise.race([...])`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise/race) behaves). When one effect completes or throws, that result is returned (or thrown) by the race, and all the other effects are cancelled. If the result is not an error, it wrapped in an object of the form `{[key]: result}`; this allows the caller to determine which Effect returned the result
+multiple Effects (this is similar to how [`Promise.race([...])`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise/race) behaves). When one effect completes or throws, that result is returned (or thrown) by the race, and all the other effects are cancelled. If the result is not an error, it wrapped in an object of the form `{[key]: result}`; this allows the caller to determine which Effect returned the result. If the effect throws an `Error`, the `raceOrigin` property is attached to the thrown error, allowing the caller to determine which Effect threw the error.
 
 - `effects: Object` - a dictionary Object of the form {label: effect, ...}
 - `options: Object` - an optional Object containing options for the race. Currently supported options are:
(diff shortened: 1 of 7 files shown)


src/internal/proc.js
```diff
@@ -600,9 +600,15 @@ export default function proc(
         }
 
         if(isErr) {
+          // TODO: pre-calculate these switches
+          // TODO: allow users to customize origin / raceOrigin key?
+          // TODO: is this behavior correct if typeof res !== Error?
+          if(annotateErr) { res.raceOrigin = key }
```
Andarist: Thinking about the name for this - any other ideas than `raceOrigin`, `effectOrigin` ?

Andarist: also seting this property should be done with `defineProperty` I think and it should be excluded from enumerability

Andarist:
I was thinking more about adding the second option (`annotateErr`) as default and for everyone. Dont think there is actually need for `options` (at least not yet).

This would allow keeping the code simpler.

Lucretiel:
I wasn't sure if modifying the `Error` counted as a breaking change; I also wasn't sure about the case where an Effect throws something that's not an `Error`

src/internal/proc.js
Lucretiel: I'll update the diff.

> any other ideas than `raceOrigin`, `effectOrigin`

Nothing comes to mind... any reason?

Andarist:
good point about throwing something else than `Error`, we could check `instanceof` before assigning this prop

as to modifying the `Error` - should be no biggy I think

src/internal/proc.js
Andarist: not yet, will try to think about the most appropriate name

Lucretiel:
How will users determine the origin of non-`Error` errors? For instance, jqXHR just raises itself in the event of an error. Can we just expect users just wrap these in `Error` types?

Andarist:
`jqXHR` does not implement Promise/A+ specification, thus we won't support its non-standard behaviours.

If you want to benefit from this change while using `jqXHR` I would advise you to write a wrapper for making requests which should use this to pass the `Error` to the saga:
```javascript
function request(opts) {
	var promise = $.ajax(opts)
		.then(
		data => data,
		(jqXHR, textStatus, errorThrown) => {
			// use those ^ to construct the Error
			return new Error()
		})
	
	return promise
}
```
with that approach `redux-saga` will be able to handle both cases by chaining `.then` to the returned `promise`

src/internal/proc.js
Lucretiel: Do you mean that you want it to be more unique? Like `CANCEL`?

Lucretiel:
How would you feel about retaining `wrapErr`, in that case? That way, users have a catch-all solution if they want to catch thrown objects more generically and immutably.

Andarist:
Its not really immutable way of wrapping the error - it mutates the original one, but I dont see anything wrong with that in this particular use case.

Would like to keep this change minimal and adding `options` to the race is quite a big change, no other effect is configurable like that.

Lucretiel:
It seems to me that there are effects taking more than one argument, but I take your meaning about wanting to keep things simple. Will fix.

src/internal/proc.js
Lucretiel: How would you feel about a new symbol, then? Users would do `err[RACE_ORIGIN]` instead of `err.raceOrigin`

Lucretiel:
Also, I'm trying to write a test for this functionality, but I confess I cannot for the life of me understand what the existing tests are actually doing.

Andarist:
> It seems to me that there are effects taking more than one argument, but I take your meaning about wanting to keep things simple. Will fix.

Sure there are, I ment that there is no 'configurable' effect in such a way now. My comment was not about arguments count per se, just presence of `options` argument

> Also, I'm trying to write a test for this functionality, but I confess I cannot for the life of me understand what the existing tests are actually doing.

Could u give a link to the lines you have trouble with? I'm happy to explain everything

Lucretiel:
I added a test, so you can tell me if it makes sense or not. Mostly I'm not clear about what the Promise.resolve is doing at the beginning of the test.

docs/api/README.md
```diff
@@ -37,7 +37,7 @@
   * [`setContext(props)`](#setcontextprops)
   * [`getContext(prop)`](#getcontextprop)
 * [`Effect combinators`](#effect-combinators)
-  * [`race(effects)`](#raceeffects)
+  * [`race(effects, options)`](#raceeffects)
```
Andarist: this should be removed

docs/api/README.md
```diff
@@ -761,12 +761,14 @@ Creates an effect that instructs the middleware to return a specific property of
 
 ## Effect combinators
 
-### `race(effects)`
+### `race(effects, options)`
```
Andarist: and this too

docs/api/README.md
```diff
@@ -761,12 +761,14 @@ Creates an effect that instructs the middleware to return a specific property of
...
 
 Creates an Effect description that instructs the middleware to run a *Race* between
-multiple Effects (this is similar to how [`Promise.race([...])`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise/race) behaves).
+multiple Effects (this is similar to how [`Promise.race([...])`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise/race) behaves). When one effect completes or throws, that result is returned (or thrown) by the race, and all the other effects are cancelled. If the result is not an error, it wrapped in an object of the form `{[key]: result}`; this allows the caller to determine which Effect returned the result. If the effect throws an `Error`, the `raceOrigin` property is attached to the thrown error, allowing the caller to determine which Effect threw the error.
```
Andarist: this should be revamped and replaced with `RACE_ORIGIN` mention and example

src/internal/proc.js
```diff
@@ -1,4 +1,8 @@
-import { noop, kTrue, is, log as _log, check, deferred, uid as nextEffectId, array, remove, object, TASK, CANCEL, SELF_CANCELLATION, makeIterator, createSetContextWarning, deprecate, updateIncentive } from './utils'
+import {
```
Andarist: hm, thats becoming quite a long list, but newline shouldnt be inserted here yet, gonna think how those utils could be splitted later

src/internal/proc.js
```diff
@@ -600,6 +604,13 @@ export default function proc(
         }
 
         if(isErr) {
+          // TODO: allow users to customize origin / raceOrigin key?
```
Andarist: those comments are really not neccessary

test/proc/race.js
```diff
@@ -110,3 +110,22 @@ test('processor race between sync effects', assert => {
     assert.end();
   })
 });
+
+test('processor race error has origin', assert => {
```
Andarist: most tests do not put assert calls inside a generator, but rather have `actual` and `expected` arrays which are compared at the end, so basically that would look something like
```
test('processor race error has origin', assert => {
  assert.plan(1);
  const actual = []
  const expected = ['willErr']

  function* genFn() {
    try {
      yield io.race({
        wontResolve: new Promise((resolve, reject) => {}),
        willErr: Promise.reject(new Error("FAIL")),
      })
    } catch(err) {
      actual.push(err[RACE_ORIGIN])
    }
  }

  proc(genFn()).done.then(() => {
    assert.deepEqual(actual, expected, 'thrown error should have RACE_ORIGIN set on itself')
  })
})
```

sometimes smth like `Promise.resolve(1)` is used in test to cause asynchrony, no need for that here though

this should also be accompanied by a test which do not set `RACE_ORIGIN`, when something else than an Error is thrown

Andarist:
Sorry for a long delay with this PR, but I've decided not to proceed with this feature right now. If you want to know error's origin you can track it on your own with a simple helper.

PR #976, opened 2017-05-08
Now automatically handle jqXHR

Lucretiel:
Jquery is popular and standard enough that it's probably worth automatically supporting cancellation on `jqXHR` instances, which are the Promise type used for most jq async operations. This patch simply makes it so that, in addition to checking for `promise[CANCEL]`, the promise handler also checks for `promise.abort`, and sets up a cancellation handler if that function exists.

### DIFF
--- a/src/internal/proc.js
+++ b/src/internal/proc.js
@@ -403,6 +403,13 @@ export default function proc(
     const cancelPromise = promise[CANCEL]
     if(typeof cancelPromise === 'function') {
       cb.cancel = cancelPromise
+    } else {
+      // Support jqXHR
+      if(typeof promise.abort === 'function') {
+        cb.cancel = () => promise.abort()
+      }
+      // TODO: add support for the fetch API, whenever they get around to
+      // adding cancel support
     }
     promise.then(
       cb,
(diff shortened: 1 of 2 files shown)


src/internal/proc.js
```diff
@@ -403,6 +403,13 @@ export default function proc(
     const cancelPromise = promise[CANCEL]
     if(typeof cancelPromise === 'function') {
       cb.cancel = cancelPromise
+    } else {
```
Andarist: should be `else if`, not a nested `if`

Lucretiel: I did it this way cause there wasn't really a good scope to put the local abort variable.

Andarist: hm, could u be more precise? i dont see any local variable here named `abort`, the only one is `promise`

Lucretiel: Oh, I apparently changed it. Originally I was storing `promise.abort` as a local variable. Will change to else if.

Andarist:
Thanks!

src/internal/proc.js
```diff
@@ -403,6 +403,13 @@ export default function proc(
...
+      // Support jqXHR
+      if(typeof promise.abort === 'function') {
```
yelouafi: offtopic but could probably use `is.func` util here and the 1st for the test

Andarist: oh yeah, thanks for noticing, gonna change this in a moment

PR #981, opened 2017-05-09
[Resolve #978] Update documentation with ES5 syntax for creating middleware

### LINKED ISSUE #978: Creating Saga middleware without using import syntax
I'm currently reviving an old project of mine. I'd like to migrate it to React+Redux, and to use Saga. Problem is, currently the project is not set up to use ES6 module syntax: no Webpack, no Babel, no System.js, nothing. Just a bunch of `<script />` tags to load dependencies. So I can't do this:

    import createSagaMiddleware from 'redux-saga'

I'm most likely going to work on my build system reasonably soon, but would really appreciate if I could get started with Saga before that. I looked at the ReduxSaga object that's exported to `window`, but can't find anything that would help me create middleware.

Anything I can do to create middleware with my current setup? Or a workaround?

septagram:
Documentation for UMD build explains how to load `redux-saga`, but doesn't explain how to create Saga middleware, which is not obvious because `createSagaMiddleware()` function cannot be found anywhere on the `ReduxSaga` object.

This should clarify.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,11 @@ sagaMiddleware.run(mySaga)
 
 # Using umd build in the browser
 
-There is also a **umd** build of `redux-saga` available in the `dist/` folder. When using the umd build `redux-saga` is available as `ReduxSaga` in the window object.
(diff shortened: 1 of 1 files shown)


Andarist:
Thanks!

PR #991, opened 2017-05-16
feat: use "debug" to log debug information

gajus:
As it stands, the `./redux-saga/es/internal/utils.js` is flooding the `console.log` with events without a simple way to filter out those events.

screen shot 2017-05-16 at 11 41 18

This PR adds [`debug` module](https://github.com/visionmedia/debug), which enables to filter out these messages in Node.js/ [Browser environment](https://github.com/visionmedia/debug#browser-support), e.g.

```js
localStorage.debug = 'redux-saga:*'

```

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,9 @@
     "url": "https://github.com/redux-saga/redux-saga/issues"
   },
   "homepage": "https://redux-saga.js.org/",
-  "dependencies": {},
+  "dependencies": {
+    "debug": "^2.6.6"
+  },
   "devDependencies": {
     "babel-cli": "^6.1.18",
     "babel-core": "^6.21.0",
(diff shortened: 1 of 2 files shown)


gajus:
Is there a way to filter out these logs at all?

I had to go to the source code and just delete the `console.log` logic.

gajus:
> Is there a way to filter out these logs at all? (e.g. in Chrome dev tools)

`/^((?!has been cancelled).)*$/` as a filter.

Andarist:
Hi @gajus , thanks for the PR. i feel though that this should be fixed in other way than this. proposed fix is here - https://github.com/redux-saga/redux-saga/pull/945 and the the issue is here https://github.com/redux-saga/redux-saga/issues/614 . 

Generally speaking this log should be omitted when saga is canceled via `cancel` effect, but it should stay when a saga gets cancelled due to error propagation.

gajus:
@Andarist Understood. I will comment on the respective issue/ PRs.

Issue #1012, opened 2017-06-05
Rejected Promise does not throw inside generators if rejected from another generator


jimmoy:
In this code, I am seeing `thisSagaShouldFail()` succeeding, and then reporting the promise rejection in my Chrome console. I think it should catch an error caused by the rejected promise:

```javascript
function *promiseFn(flag) {
    if (flag) return new Promise(resolve => resolve());
    else      return new Promise((r, reject) => reject({ error: 'some error' }));
}

function* thisSagaShouldSucceed() {
    try {
        yield call(promiseFn, true);
        console.log('thisSagaShouldSucceed: success!');
    } catch (error) {
        console.log('thisSagaShouldSucceed: fail!');
    }
}

function* thisSagaShouldFail() {
    try {
        yield call(promiseFn, false);
        console.log('thisSagaShouldFail: success!');
    } catch (error) {
        console.log('thisSagaShouldFail: fail!');
    }
}
```
Chrome console:
![2017-06-05 17-02-08](https://cloud.githubusercontent.com/assets/19914319/26806948/be25f84e-4a10-11e7-97d1-df63f5e6f973.png)

The test case is the same as in #283 (thank you @luisherranz) except `promiseFn()` is a generator instead of a normal function. I'm also including an object in the rejection to make it clear I don't think this is necessarily related.

I found this change in behavior while upgrading from 0.9.5 to 0.15.4, and then narrowed it down to version 0.10.0 being the version that introduced it.

Andarist:
This is because actually if u return a promise it wont get resolved automatically for u, so what u get back into ur saga is just a promise. 

Previously (pre-v0.10) internals were scheduled using Promises, but unfortunately it didnt give us enough power to steer them, introduced some troublesome bugs regarding the reace conditions and so on and such scheduling had to be dropped. 

How u have used the `call` effect with combination of the generator returning a promise worked probably quite by accident, because of the how internal promises were chained and that ur returned promise got into this chain.

The preffered way to resolve ur promises inside a saga is `call` a function returning a promise. Its because it encapsulates the promise, making it kinda lazy - wont be evaluated until saga will interpret the `call` effect. Problem with promises is exactly this - that they are eager, so they execute themselves immediately and u probably dont want that in testing environment etc. I would advise u to wrap those promises in a `call`.

That being said - I will consider supporting such use case as generator is a special function and we allow functions returning promises. Need to think about it though, so no promises.

jimmoy:
[...]

PR #1017, opened 2017-06-06
Update call() per #1012

jimmoy:
Clarification on doc for `call()`


Andarist:
Thanks for the PR! I will let it hang here for a moment though, as I'm thinking if the behaviour described by you in the issue should be restored.

jimmoy:
Yup, whatever direction you see appropriate, I'll be able to move with.

Andarist:
I know it took quite some time to merge this - but thanks for your input!

PR #1031, opened 2017-06-21
Make the code examples in the beginner's tutorial a bit clearer

asfktz:
Hi

I made a few changes to the beginner's tutorial, little things that as a first time reader wished they 
where there.

like a few more `...`  when appropriate, and adding a button to the counter instead of replacing one (that why keeping the original onIncrement callback makes more sense)

That's about it (:
Asaf


Andarist:
Thanks!

PR #1032, opened 2017-06-21
Make the test pass in "Basic Concepts: Dispatching actions "

asfktz:
Hi again (:

I'm new to saga so I could be wrong,
but I believe that the test at the end of [Basic Concepts: Dispatching actions ](https://redux-saga.js.org/docs/basics/DispatchingActions.html)
fails because the fake response needs to be `undefined` instead of `{}`

```js
// ...

// create a fake response
const products = {} // should be undefined ?

// expects a dispatch instruction
assert.deepEqual(
  iterator.next(products).value,
  put({ type: 'PRODUCTS_RECEIVED', products }),
  "fetchProducts should yield an Effect put({ type: 'PRODUCTS_RECEIVED', products })"
)
```

Am I right?

### DIFF
--- a/docs/basics/DispatchingActions.md
+++ b/docs/basics/DispatchingActions.md
@@ -54,7 +54,7 @@ assert.deepEqual(
 )
 
 // create a fake response
-const products = {}
+const products = undefined
 
 // expects a dispatch instruction
 assert.deepEqual(


Andarist:
Is it failing though? it can be whatever u really want (`undefined` included, although I would suggest to use some actual value) as long as it can be equaled deeply

by calling `.next()` with argument (`products` in this case) you effectively inject it into the saga and it resumes its execution with injected value

asfktz:
Oh, I see now!
I didn't realized that `iterator.next` needs to be called with `products`.
It makes a lot of sense.

Thanks, @Andarist

Andarist:
Well, it doesnt really need to be called with it. But if you want to have something assigned as `yield`'s result then yes - it needs to be. I.e.

PR #1035, opened 2017-06-26
Typings: add optional cancel property on CpsCallback

dannsam:
It is possible to write a cancellable CpsEffect by setting the `cancel` property to a function you want to get executed on cancel. Let's update the typings to match the logic in the code.

### DIFF
--- a/effects.d.ts
+++ b/effects.d.ts
@@ -238,7 +238,10 @@ export interface CpsEffect {
   CPS: CallEffectDescriptor;
 }
 
-type CpsCallback = (error: any, result: any) => void;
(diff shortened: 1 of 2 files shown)


Andarist:
Cc @aikoven

aikoven:
We should add a test for this.

Also, how about this:
```ts
type CpsCallback = {
  (error: any, result: any): void;
  cancel?(): void;
};
```

dannsam:
@aikoven thanks, I like it! The PR is updated.

Andarist:
Merging in, thanks!

Issue #1029, opened 2017-06-19
Typescript support requires "es2015.iterable" lib


letharion:
If you end up here from google needing help, check out the [`--lib`](https://www.typescriptlang.org/docs/handbook/compiler-options.html) setting and make sure you have `es2015.iterable` enabled.

I'm guessing that `es2015.iterable` is included in the default configuration, and this is a somewhat rare issue.

In my case however, it was not a part of the included libs, and so compilation failed for me as follows:

```
error TS2345: Argument of type '() => IterableIterator<ForkEffect>' is not assignable to parameter of type 'Saga0'.
  Type 'IterableIterator<ForkEffect>' is not assignable to type 'Iterator<any>'.
    Property 'next' is missing in type 'IterableIterator<ForkEffect>'
```

I struggled for a while with digging through the `d.ts` but not getting anywhere. Turns out there was a simple solution, so I wanted to post it here.

letharion:
Leaving the issue open so a maintainer can decide if anything needs to be done, but as far as I'm concerned, it can be closed as is. :)

Andarist:
@aikoven should it get documented or closed?

aikoven:
I think it can be closed now because typings work fine with default TS config. We could consider adding docs if there's a demand.

letharion:
@aikoven Looking at the documention again it says

> The default library injected is: 
> ► For --target ES5: DOM,ES5,ScriptHost
> ► For --target ES6: DOM,ES6,DOM.Iterable,ScriptHost

So if you're targetting ES6, it's on by default, but if you're targetting ES5, then it's not.

Andarist:
I would prefer to just close it now, people should be able to find this issue to solve their issues in the future

@letharion ofc I would gladly accept a PR with an explanation if you can provide one :)

rplotkin:
This is NOT easy to find. Can it not at least be mentioned in the typings file?

Andarist:
info about it is here https://github.com/redux-saga/redux-saga#typescript

PR #1036, opened 2017-06-26
Document Iterable TypeScript dependency.

letharion:
Following https://github.com/redux-saga/redux-saga/issues/1029, I made a search on github to see how many users would be affected if they started using Saga. Since `filename:tsconfig.json "ES5"` has several times more matches than ES6 does, I figured it might actually be worthwhile documenting better than just posting an issue.

@Andarist Since you suggested you'd accept a PR, here is one. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,10 @@ $ npm run real-world
 # sorry, no tests yet
 ```
 
+### TypeScript
+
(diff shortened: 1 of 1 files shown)


Andarist:
Thanks!

Issue #1044, opened 2017-07-01
Is BeginnerTutorial with yield all correct? 


mauron85:
I'm using redux-saga with typescript and trying very basic example from
https://redux-saga.js.org/docs/introduction/BeginnerTutorial.html

This will not compile with typescript

```
import { SagaIterator } from 'redux-saga';
import { all, takeEvery, call } from 'redux-saga/effects';

// Our watcher Saga: spawn a new incrementAsync task on each INCREMENT_ASYNC
export function* watchIncrementAsync(): SagaIterator {
  yield takeEvery('INCREMENT_ASYNC', incrementAsync)
}

// single entry point to start all Sagas at once
export default function* rootSaga() {
  yield all([
    helloSaga(),
    watchIncrementAsync()
  ])
}
```

It produces typescript error:
```
(16,13): error TS2345: Argument of type 'IterableIterator<AllEffect | RootEffect | TakeEffect | ChannelTakeEffect<any> | PutEffect<any> | ...' is not assignable to parameter of type '{ [key: string]: Effect; }'.
  Index signature is missing in type 'IterableIterator<AllEffect | RootEffect | TakeEffect | ChannelTakeEffect<any> | PutEffect<any> | ...'.
```

However adding `call` doesn't produce any type error.
```
  yield all([ call(saga) ]);
```

I'm wondering if there is problem with type definitions or what I'm doing wrong.

Andarist:
cc @aikoven

aikoven:
Current typings were designed to protect users from yielding function calls directly, but use effects instead. This way we catch situations where you unintentionally introduce side-effect in your saga forgetting to wrap function call with effect.

IMO the example should be updated to use more idiomatic way:
```js
// single entry point to start all Sagas at once
export default function* rootSaga() {
  yield all([
    call(helloSaga),
    call(watchIncrementAsync)
  ])
}
```

Otherwise, we could try to solve it like in https://github.com/redux-saga/redux-saga/pull/770, i. e. to make the strictness opt-in by declaring the return type of generator to be `SagaIterator`.

Andarist:
If similar issue got fixed like mentioned, we should prefer consistency and do the same for this case too.

aikoven:
Fixed in https://github.com/redux-saga/redux-saga/pull/1047

PR #1047, opened 2017-07-03
Support non-strict effect combinators in TypeScript

aikoven:
Allows including any value to effect combinators `all` and `race`, e. g.:

```ts
function* saga() {
  yield all([
    functionReturningPromise(),  // previously error, now allowed
    // ...
  ])
}
```

To enable strict type-checks, where every combined value must be Effect, annotate generator with `SagaIterator` return type:

```ts
function* saga(): SagaIterator {
  yield all([
    functionReturningPromise(),  // error
    // ...
  ])

  yield all([
    call(functionReturningPromise),  // ok
    // ...
  ])
}
```

Fixes https://github.com/redux-saga/redux-saga/issues/1044.

### DIFF
--- a/effects.d.ts
+++ b/effects.d.ts
@@ -84,7 +84,13 @@ export const put: {
 };
 
 
-export type AllEffectDescriptor = Effect[] | {[key: string]: Effect};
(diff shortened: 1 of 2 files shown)


Andarist:
Thanks!

PR #1048, opened 2017-07-04
Fixed typo

mohamed-ismat:
Some might debate that both forms are right, but "First things first" is the correct form of the idiom :)
https://english.stackexchange.com/questions/7419/which-is-the-correct-idiom-first-things-first-or-first-things-first
https://www.englishforums.com/English/FirstThingsFirstFirstThings-First/xjjrg/post.htm


Andarist:
Thanks!

PR #1058, opened 2017-07-08
minor edit, made a section of the docs more pluralistic

erikvold:


Andarist:
Why the 2 -> two change?

erikvold:
> Why the 2 -> two change?

It's easier to read imho, also that appears to be the style used on must pages (not all though).

https://github.com/redux-saga/redux-saga/blob/master/docs/api/README.md

Andarist:
Thanks!

PR #1061, opened 2017-07-08
adding missing imports to examples in recipes docs

erikvold:


Andarist:
Thanks! :) you are on 🔥 !

Issue #1064, opened 2017-07-09
Add more channel tests


erikvold:
It's not clear to me how I should test the sagas that I write which use channels, like `actionChannel`.  I don't see any examples, did I miss any?  Could someone please explain how to test examples like those on https://redux-saga.js.org/docs/advanced/Channels.html ?

Andarist:
Unfortunately it's not documented well enough, but you can do something like
```js
let eff = it.next() // actionChannel effect
const mockChannel = channel()
it.next(mockChannel) // inject mockChannel back (it will become left value of the assignment in ur saga)
```
It also depends on what test exactly you would like to write, so more details would be appreciated.

erikvold:
This issue is more about adding examples to the documentation so that people can figure out how to test channels for themselves.

erikvold:
@Andarist how would one test the `throttle` example on https://redux-saga.js.org/docs/api/index.html ?


```javascript
const mySaga = function*() {
  const chan = yield actionChannel(pattern)

  while (true) {
    const action = yield take(chan)
    yield fork(task, ...args, action)
    yield call(delay, ms)
  }
}
```

When I test this I did something like you suggested first, then the next step fails:

```javascript
    let chan = actionChannel(pattern);
    assert.deepEqual(
      gen.next().value,
      chan
    );

    assert.deepEqual(
      gen.next().value,
      {} // bit confused what to test against here
    );
```

And the second test gives me:

Error: take(patternOrChannel): patternOrChannel is undefined
      at check (node_modules/redux-saga/lib/internal/utils.js:45:11)
      at take (node_modules/redux-saga/lib/internal/io.js:59:22)

and it's failing on this line ` const action = yield take(chan)`

Andarist:
This should go like this:
```js
    let chan = actionChannel(pattern);
    assert.deepEqual(
      gen.next().value,
      chan
    );

    const mockChannel = channel()

    assert.deepEqual(
      gen.next(mockChannel).value,
      take(mockChannel)
    );
```

erikvold:
@Andarist where does `channel` get exported from?

Andarist:
@erikvold its exported from the main entry point
```js
import { channel } from 'redux-saga'
```

Andarist:
Do you feel that there is still something lacking from the docs? Please specify, so we can work on it :) Your input in past few days was really great! Appreciate it a lot.

andresmijares:
[...]

PR #1066, opened 2017-07-09
Adding actionChannel generator test

erikvold:
Adding the code mentioned in https://github.com/redux-saga/redux-saga/issues/1064#issuecomment-313930462 to make sure that it works, and because I didn't see similar code.  Even though it probably doesn't add any code coverage I think it is a good example to include in the tests, as people using `redux-saga` will and should be able to look at the tests for `redux-saga` in order to learn how they can test their own code.


Andarist:
Thanks!

Issue #1067, opened 2017-07-09
Setup (travis) ci


erikvold:
It would be nice if travis-ci was setup so that tests could be run for every pull request automatically.

Andarist:
This would be great, although Im personally more familiar with CircleCI. Solution ofc doesnt matter, its the continous checks that matter.

Hadnt time to setup this, but for sure will try to do this. I want to release `v1` some in the near (*cough*) future, as we've reached stability long time ago already. This would pair really nicely with that. 

Also incorporating [bundlesize check](https://github.com/siddharthkp/bundlesize) would be cool.

camflan:
I'm open to working on this

camflan:
I have it working on my fork. Can someone else confirm that redux-saga currently has 1 failing test: `TypeScript files compile against definitions`

Andarist:
@camflan yeah, this one is not working, it only indicates that we really need the CI :P 

Thanks for the PR!

camflan:
@Andarist you got it -- I'm trying to debug the `bundlesize` part of this. I have it running locally using the travis-ci docker file, but it fails on travis-ci.org itself. I'll submit another pull once I have this working as well 👍

camflan:
nvm, you got it already :)

erikvold:
@camflan thanks for setting up travis!  /cc @Andarist 

I fixed the tests that were failing in #1079 and I think we should upgrade the version of node used in travis so I created #1078 as well.

erikvold:
after landing #1079 we can close this issue I think!

Andarist:
Thanks you guys for the great work. 👏

camflan:
👍

PR #1074, opened 2017-07-14
Adding Travis.ci

camflan:
This repo needs signed up for travis, but the config is now here. Will also need to change the build badge URLs for redux-saga once redux-saga/redux-saga has been added to travis-ci

this addresses #1067


Andarist:
Thanks for the PR, great work :)

Issue #1063, opened 2017-07-09
Using takeLatest on a channel doesn't work


erikvold:
Code:

```javascript
const channel = yield actionChannel(TYPE);
yield takeLatest(channel);
```

Expected:

works

Actual:

I got this error:

```
index.js: uncaught at mainSaga 
 at watchForSubmit
 at takeLatest 
 TypeError: Cannot read property 'name' of undefined
```

Andarist:
Could u share a reproduced issue on something like https://codesandbox.io/ ?

camflan:
Is `takeLatest` supposed to work with channels? I've haven't seen this before and there are no tests or documentation written about this. If it's supposed to be supported, we should add tests and documentation too.

Andarist:
Should work just fine as `takeLatest` (and `takeEvery`) are implemented using regular `take` which ofc accepts a channel as argument. 

**docs + tests**
Noted. Would love to get some help on this *wink*

Andarist:
@erikvold 

`yield takeLatest(channel);` - this probably is not working because u have not passed in any generator, u need to provide a saga handler for this `takeLatest`

erikvold:
@Andarist I updated my example, how should it be change based on your comment?

erikvold:
ah it looks like `takeLatest` is not meant to do what I want it to do, it's a non-blocking call effect, I was expected it to be a blocking effect when used with a channel.

erikvold:
seems like `takeLatest` should be called `forkLatest` to me.

Andarist:
well, it probably be even `takeAndForkLatest` or smth :D 

> I was expected it to be a blocking effect when used with a channel.

How did you want to use it (code-wise)?

> @Andarist I updated my example, how should it be changed based on your comment?

I think you might find this useful:
```js
const channel = yield actionChannel(TYPE, buffers.sliding(1));
// ...
const action = yield take(channel);
```

`throttle` helper is implemented under the hood using those. 

If you share what outcome are you reaching for (requirements) I might be able to advise more :)

Amareis:
How about takeEvery? Is it works as expected with channels?

Andarist:
@Amareis both - `takeEvery` and `takeLatest` - work fine with channel. @erikvold had concerns about the naming of those, because from the names he has expected different behaviour.

[...]

PR #1081, opened 2017-07-17
Issue #1063 - add test for using takeLatest on an actionChannel

erikvold:


Andarist:
Oh, I see what u meant! I think `buffers.sliding(1)` is exactly what you need. :)

erikvold:
> Oh, I see what u meant! I think buffers.sliding(1) is exactly what you need. :)

after looking in to it some more that, `buffers.dropping(1)` looks like what I want (I just need to know if there was a new action buffered), but now when I use it, my tests fail when I compare:

    assert.deepEqual(gen.next().value, actionChannel(TYPE, buffers.dropping(1)));

When I log the two values I get this:

```
{ '@@redux-saga/IO': true,
  ACTION_CHANNEL: 
   { pattern: 'TYPE',
     buffer: 
      { isEmpty: [Function: isEmpty],
        put: [Function: put],
        take: [Function: take],
        flush: [Function: flush] } } }
{ '@@redux-saga/IO': true,
  ACTION_CHANNEL: 
   { pattern: 'TYPE',
     buffer: 
      { isEmpty: [Function: isEmpty],
        put: [Function: put],
        take: [Function: take],
        flush: [Function: flush] } } }
```

So I guess that the functions are being compared and they are not equal.

erikvold:
@Andarist I recreated my test in #1085 to show what is failing.

Andarist:
Going to close this one as it seems not relevant anymore, please ping me here if you feel it needs reopening though.

Issue #1084, opened 2017-07-18
Add tests for using actionChannels with buffers to channel-recipes tests


erikvold:
I don't see any tests that use actionChannels with buffers.

PR #1085, opened 2017-07-18
Issue #1084 - Add tests for using actionChannels with buffers

erikvold:


erikvold:
Seems like the only way to fix this would be to rewrite the buffers implementation to avoid using functions, maybe so that it is just an array and a type, but also we should support the old buffers implementation still...

Andarist:
Yeah, that's certainly the problem which bite some people in the past as like you have noticed - checking equality wont do us much here.

Packaging it as some kind of 'effect' - like your proposed array+type - would obscure the API in my feeling.

In most cases effects are effects because its the redux-saga interpreter that knows how to interpret and execute them. It also allows effects being pure - no action will be executed (no async calls etc) until it gets interpreted and executed.

With buffers the story is simple - they are primitive objects which does not create any side effects and return synchronously, so packaging them as effect would be quite odd I think.

That being said - I recognize the problem, just aint sure whats the best way to deal with it.

**As workaround** you can do something like this:
```js
// saga.js
const chan = yield actionChannel('ACTION', yield buffers.dropping(1))

// test.js
import { is } from 'redux-saga/utils'

let gen = saga()
assert.ok(is.buffer(gen.next().value))
const mockBuffer = buffers.dropping(1)
let chan = actionChannel('ACTION', mockBuffer)
assert.deepEqual(gen.next(mockBuffer).value, chan)
```

erikvold:
> With buffers the story is simple - they are primitive objects which does not create any side effects and return synchronously, so packaging them as effect would be quite odd I think.

I'm still not understanding why we can't or shouldn't implement buffers as effects.  Why would this  be considered odd?  

It seemed odd to me the way that it is currently implemented, seeing as how I can't test the recommended way to test.

Andarist:
The problem is much broader here as you wont be able to test anything that contain local references to functions. This would only help with testing the buffers. 

Also you can actually test those as effects by changing to `yield call(buffers.dropping, 1)`. No extra effect required.

Andarist:
@erikvold also it should be possible to reimplement `buffers` to classes (bummer ;P) so their methods could be shared on the prototype and therefore make it possible to deep equal them

erikvold:
> The problem is much broader here as you wont be able to test anything that contain local references to functions. This would only help with testing the buffers.

That's a fair point.

Andarist:
Thanks! Gonna merge in but if you have any more ideas how the overall issue could be improved please open an issue for discussion.

PR #1089, opened 2017-07-18
Make single quotes vs double quotes consistent in the first example

Vpr99:
My excellent coworker noticed that all of the examples below the first code block use double quotes, but the first dispatch uses single quotes. Was this intentional?


Andarist:
It was certainly not intentional, however our source code uses single quotes, so the docs should be changed to use those where inconsistent. Would you like to provide an apposite of this PR? :)

PR #1092, opened 2017-07-20
fixed: npm test now runs linter and example tests

erikvold:
Note I also removed the `npm run check` script because it is now redundant.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,8 @@ language: node_js
 node_js:
   - "node"
 
+script:
+  - npm run check
+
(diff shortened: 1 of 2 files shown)


Andarist:
Thanks for the PR, but actually I would prefer to keep `check` as lint+test. I prefer explicit naming and it has always felt weird to put linting inside a test task as those are separate activities in my opinion.

erikvold:
@Andarist ok I reverted the removal of the `check` script now.

Andarist:
Thanks :)

Issue #1096, opened 2017-07-21
Missing TypeScript definition for cloneableGenerator


stevenmusumeche:
I am importing this method in a TypeScript file like this:

    import { cloneableGenerator } from 'redux-saga/utils';

The actual code works great, however, I'm getting a TypeScript error.

```
file: '..... src/sagas/bootstrap.test.ts'
severity: 'Error'
message: 'Module '.... node_modules/redux-saga/utils"' has no exported member 'cloneableGenerator'.'
at: '7,10'
source: 'ts'
```

Andarist:
@stevenmusumeche included typings are community-driven, so they always need some time to catch up after releasing features etc.

However this one was already merged in some time ago, gonna release new version with it right now. :)

stevenmusumeche:
Thank you! I saw them in the code but couldn't find them in the release :)

Andarist:
should already be released, please check if everything is working fine

stevenmusumeche:
@Andarist I actually get the same error with 0.15.5. I can silence it like this, but it's not ideal.

```
declare module 'redux-saga/utils' {
    var cloneableGenerator: (foo: any) => any;
}
```

Andarist:
@stevenmusumeche could u check why is it happening? does your `./node_modules/redux-saga/utils.d.ts` contain `cloneableGenerator` at the bottom of the file?

Rokt33r:
@andarist I also confirmed, It doesn't exists there.

Andarist:
@Rokt33r could u check why it might be broken? When I install `redux-saga` and inspect `utils.d.ts` contents I can see it has `cloneableGenerator` definition

stevenmusumeche:
Actually, it is there and is working for me now.

Rokt33r:
@Andarist Yea, it exists. I tried to find it with typo. :dizzy_face::dizzy_face:

Andarist:
Closing this then :)

PR #1098, opened 2017-07-21
Make take and put effects redeclarable

### LINKED ISSUE #1094: More typing extensibility for effects
I'm developing a redux enhancer for dispatch batched actions. 

As you see below, my actions are batched into another single action.

```ts
      yield put(batchAction([
        {
          type: 'SayBye'
        },
        {
          type: 'SayBye'
        },
        {
          type: 'SayBye'
        },
      ]))
```
> https://github.com/Revisolution/typed-redux-kit/blob/batched-actions/src/specs/batchEnhancer.spec.ts

This is not that bad, but I think it would be much nicer if I could `put` actions just as an array like the below example:
```ts
      yield put([
        {
          type: 'SayBye'
        },
        {
          type: 'SayBye'
        },
        {
          type: 'SayBye'
        }
      ])
```
> https://github.com/Revisolution/typed-redux-kit/blob/batched-actions-exp/src/specs/batchEnhancer.spec.ts#L42-L52

To make it possible, I have to extend dispatch and put functions just like redux-thunk does.

Here comes the problem. The typing for `put` of redux-saga is declared as a const.
It means there is no way to overload this function.

So, to make the second example possible, I have to modify the typing.

Original : 
```ts
export const put: {
  <A extends Action>(action: A): PutEffect<A>;
  <T>(channel: Channel<T>, action: T | END): ChannelPutEffect<T | END>;
  ...
};
```

Modified : 
```ts
export interface Put {
  <A extends Action>(action: A): PutEffect<A>;
  <T>(channel: Channel<T>, action: T | END): ChannelPutEffect<T | END>;
  ...
}
export const put: Put
```

Then, I could overload `dispatch` and `put`.
```ts
declare module 'redux' {
  export interface Dispatch<S> {
    <A extends Redux.Action>(asyncAction: A | A[]): A
  }
}

declare module 'redux-saga/effects' {
  export interface Put {
    <A extends Redux.Action>(action: A[]): ReduxSagaEffects.PutEffect<A>
    resolve<A extends Redux.Action>(action: A[]): ReduxSagaEffects.PutEffect<A>
  }
}
```
https://github.com/Revisolution/typed-redux-kit/blob/batched-actions-exp/src/lib/batched-actions.ts#L5-L16

Does it make sense? If so, I'll send a Pull request to fix this.

Rokt33r:
This will fix #1094.

I modified only `take` and `put` effects. Other effects doesn't seem to have this problem.

### DIFF
--- a/effects.d.ts
+++ b/effects.d.ts
@@ -36,13 +36,15 @@ export interface ChannelTakeEffect<T> {
   TAKE: ChannelTakeEffectDescriptor<T>;
 }
 
-export const take: {
+export interface Take {
   <A extends Action>(pattern?: Pattern): TakeEffect;
   <T>(channel: Channel<T>): ChannelTakeEffect<T>;
 
   maybe<A extends Action>(pattern?: Pattern): TakeEffect;
   maybe<T>(channel: Channel<T>): ChannelTakeEffect<T>;
-};
+}
+
+export const take: Take;
 
 /**
  * @deprecated
@@ -70,7 +72,7 @@ export interface ChannelPutEffect<T> {
   PUT: ChannelPutEffectDescriptor<T>;
 }
 
-export const put: {
+export interface Put {
   <A extends Action>(action: A): PutEffect<A>;
   <T>(channel: Channel<T>, action: T | END): ChannelPutEffect<T | END>;
 
@@ -80,8 +82,11 @@ export const put: {
   /**
    * @deprecated
    */
-  sync: typeof put.resolve;
-};
+  sync<A extends Action>(action: A): PutEffect<A>;
+  sync<T>(channel: Channel<T>, action: T | END): ChannelPutEffect<T | END>;
+}
+
+export const put: Put;
 
 
 export type GenericAllEffectDescriptor<T> = T[] | {[key: string]: T};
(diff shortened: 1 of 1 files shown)


effects.d.ts
```diff
@@ -80,8 +82,11 @@ export const put: {
   /**
    * @deprecated
    */
-  sync: typeof put.resolve;
-};
+  sync<A extends Action>(action: A): PutEffect<A>;
+  sync<T>(channel: Channel<T>, action: T | END): ChannelPutEffect<T | END>;
```
aikoven: Let's mark second `sync` signature as deprecated as well

aikoven:
Aside from the minor comment, 👍

effects.d.ts
Rokt33r: Like this?
```ts
export interface Put {
  <A extends Action>(action: A): PutEffect<A>;
  <T>(channel: Channel<T>, action: T | END): ChannelPutEffect<T | END>;

  resolve<A extends Action>(action: A): PutEffect<A>;
  resolve<T>(channel: Channel<T>, action: T | END): ChannelPutEffect<T | END>;

  /**
   * @deprecated
   */
  sync<A extends Action>(action: A): PutEffect<A>;
  /**
   * @deprecated
   */
  sync<T>(channel: Channel<T>, action: T | END): ChannelPutEffect<T | END>;
}
```

Andarist:
Thank you guys, ive added the deprecation warning for the overloaded `put.sync` and merged this in :)

Rokt33r:
@Andarist Could you tell me ETA? I really need this! 😈 😈

Andarist:
@Rokt33r could u check (and possibly fix) whats happening with this https://github.com/redux-saga/redux-saga/issues/1096 ? would love to release both at once

Andarist:
@Rokt33r gonna prepare a release later today

Rokt33r:
@Andarist I really appreciate it. You saved me! 😉

Andarist:
@Rokt33r published :)

but also do not be afraid of using ur own fork for a while, you **can** include other version of the library in your package.json instead of the version published to npm

PR #1113, opened 2017-07-27
Ignore logo for npm package

danners:
I don't think it is needed to ship the logos with the npm package, ignore them.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -8,4 +8,5 @@ examples
 
 # doc folders
 docs
+logo
 _book


Andarist:
makes sense :) maybe we should consider using `files` entry in `package.json` instead of `.npmignore`?

Andarist:
@danners Ive thought about this a little bit and made a poll on twitter about .npmignore vs files and according to the results (and my personal thoughts on the subject too) we should switch to using files entry in `package.json`. Would u like to prepare a PR for this or should I just push this issue forward on my own?

Cheers and thanks for bringing this up!

PR #1131, opened 2017-08-16
Add missing all combinator in cheatsheet

ca057:
In version 0.15.0 the `all`-effect was added to replace yielding of arrays with effects, which became deprecated. The documenation still includes the array in its cheatsheet at the end of the API-section. This PR exchanges the array with the `all`-effect. As a consequence of the longer text, the whitespace inside the table was adapted to support a visually pleasing formatting.

The attached screenshot shows the changed parts of the cheatsheet:
redux-saga-pr

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -1193,27 +1193,27 @@ For testing purposes only.
 
 ### Blocking / Non-blocking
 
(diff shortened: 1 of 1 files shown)


Andarist:
Thanks! I've missed that :)

PR #1138, opened 2017-08-21
Add a link to "analytics on easy mode"

goshakkk:
I've recently blogged on how the reactive approach of redux-saga can make analytics tracking cleaner.  ⚡️ 

I hope there's no problem with me proposing yet another article of mine.


Andarist:
No problem with adding another article, they are always welcome. I would argue however that instead of calling `window.ga` directly (`window.ga('track', 'CreateTodo');`) it would be more preferable to wrap it in a `call` effect.

Also im not sure why `analytics-sagas.js` calls `sagaMiddleware.run` twice and exports sagas in the same time. I think those might be some artifacts and it would be good to clean them up.

goshakkk:
Oh right, the duplicate calls were a leftover that I thought I've removed.

I agree that wrapping in yield call makes sense. I fixed both of these now

goshakkk:
<3

PR #1139, opened 2017-08-23
Add typings for `detach`; fix typings tests reporting

aikoven:
* Add typings for `detach` effect transformer (https://github.com/redux-saga/redux-saga/pull/1121)
* Fix typings test failures not being displayed in test output
* Remove failing typings test


Andarist:
Thanks! :)

PR #1147, opened 2017-08-28
Fix anchor link to blocking / non-blocking

taehwanno:
This pull request fixes `blocking / non-blocking` anchor link in API Reference


Andarist:
Thanks!

PR #1158, opened 2017-09-04
Fix typo in rootSaga

IgorSapo:


Andarist:
`helloSaga` is what should be there, it's mentioned above. I agree that it might not be obvious why its there and this could be improved. But running `incrementAsync` instead of it doesnt make much sense, `incrementAsync` is supposed to be triggered by `watchIncrementAsync`. 

If you have any questions, please ask.

PR #1160, opened 2017-09-06
Fixed test in the Task cancellation subchapter

jankalfus:
There needs to be `true` passed to the generator's next function, otherwise the while loop in the `main `generator gets terminated and the following assertions fail.

### DIFF
--- a/docs/advanced/TaskCancellation.md
+++ b/docs/advanced/TaskCancellation.md
@@ -93,7 +93,7 @@ describe('main', () => {
 
   it('forks the service', () => {
     const expectedYield = fork(bgSync);
-    expect(generator.next().value).to.deep.equal(expectedYield);
+    expect(generator.next(true).value).to.deep.equal(expectedYield);
   });
 
   it('waits for stop action and then cancels the service', () => {
--- a/docs/advanced/TaskCancellation.md
+++ b/docs/advanced/TaskCancellation.md
@@ -93,7 +93,8 @@ describe('main', () => {
 
   it('forks the service', () => {
     const expectedYield = fork(bgSync);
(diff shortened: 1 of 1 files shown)


Andarist:
Actually it would be better to inject a mocked action there as you are injecting result value of the `take('START_BACKGROUND_SYNC')`. So maybe something like:
```js
  it('forks the service', () => {
    const expectedYield = fork(bgSync);
    const mockedAction = { type: 'START_BACKGROUND_SYNC' } 
    expect(generator.next(mockedAction).value).to.deep.equal(expectedYield);
  });
```

jankalfus:
Ah, right :) I'm not sure how the internals work (it's not really important for the test), but it totally makes sense. Haven't looked at it that way. So I guess we can go with your proposal :)

Andarist:
Well, in those regards internals are pretty straightforward. In general `redux-saga` takes ur generators (sagas) and calls `.next()` on the created iterators - this is how it receives effects (what you `yield`) and it knows what to do by inspecting type of the received effect. When it decides the effect has resolved (in case of `take` it is ofc when it receives action of certain type in the middleware) it simply calls `.next(result)`. 

In tests you do not handle ur generators to the `redux-saga` but instead you run them on ur own (by calling `.next` in the same way it would call it) - so it makes more sense to 'resolve' `take` effect with a 'mocked action'.

jankalfus:
I was just unsure what does the middleware pass as a value to the generator. I know that in case of the `fork` effect, it's a `Task` object. But I don't think I've seen in the docs what gets passed after yielding the `take` effect. Thanks so much for explaining! It's all clear to me now.

PR #1167, opened 2017-09-12
Added race([...effects]) to docs api reference

klis87:


docs/api/README.md
```diff
@@ -792,6 +793,35 @@ will be a single keyed object `{cancel: action}`, where action is the dispatched
 
 When resolving a `race`, the middleware automatically cancels all the losing Effects.
 
+### `race([...effects]) (with Array)`
+
+The same as [`race(effects)`](#raceeffects) but let's you to pass in an array of effects.
```
Andarist: let's -> let

docs/api/README.md
```diff
@@ -792,6 +793,35 @@ will be a single keyed object `{cancel: action}`, where action is the dispatched
...
+function* fetchUsersSaga {
+  const result = yield race([
+    call(fetchUsers),
+    take(CANCEL_FETCH)
+  ])
+}
+```
+
+If `call(fetchUsers)` resolves (or rejects) first, the result of `race` will be an object
+with a single keyed object `{'0': result}` where `result` is the resolved result of `fetchUsers`.
```
Andarist: Hm, this has realized me that even though `Object.keys` indeed allows us to pass in an array, the race's outcome is not an array in such case, but an object - so i.e. it cannot be destructured. If we want to put it into the docs, we should convert the result to an array.

Would you like to add this to the PR and adjust the proposed docs? Conversion can be made simply by adding a length property to the object and calling `[].slice.call(result)`. The whole work would needs to be done in [this function](https://github.com/redux-saga/redux-saga/blob/f609e805a7a877fa6b8c14934ad207a05df7aad2/src/internal/proc.js#L619) and should be really straightforward to implement. If you want to do it, please do not forget about adding tests for this :)

klis87: No problem, indeed returning Array would be more consistent :) I could try to fix it myself, but I find it hard to understand how to return Array not Object in this function, this is not a pure function and no object is returned here, so I guess I need to follow this project's convention, but I need a small hint.

Btw, wouldn't `Array.from({ ...response, length: keys.length }) `be more legible than `[].slice.call(result)`?

Andarist: Result is returned through cb. Array.from would be better but its a part of es6 standard and we cant require everyone to polyfill it

klis87: Ok thx. I did it but I have a small problem. In below test
```javascript
test('processor race between array of effects handling', assert => {
  assert.plan(1)

  let actual = []
  const timeout = deferred()
  const input = cb => {
    Promise.resolve(1).then(() => timeout.resolve(1)).then(() => cb({ type: 'action' }))
    return () => {}
  }

  function* genFn() {
    actual.push(
      yield io.race([
        io.take('action'),
        timeout.promise,
      ]),
    )
  }

  proc(genFn(), input).done.catch(err => assert.fail(err))

  const expected = [[, 1]]

  setTimeout(() => {
    assert.deepEqual(actual, expected, 'processor must fullfill race between array of effects')
    assert.end()
  })
})
```
I have problem with `const expected = [[, 1]]`. It is not the same as `[[undefined, 1]` and it breaks `eslint` rule: `error  Unexpected comma in middle of array  no-sparse-arrays`. Could we block this rule for this test or you have some alternative?

Andarist: Thanks for taking the time to work on it. Disabling the rule for this line is an option, you can do it with `// eslint-disable-next-line no-sparse-arrays`

Once you push it out im gonna checkout ur branch and why `assert.deepEqual` fails when passing `[[undefined, 1]]`

klis87:
Ups, I guess I cancelled our comments by accident. Anyway, I pushed the changes and regarding sparse Array the difference between `[undefined, 1]` and `[, 1]` is the same as between `{ x: undefined }` and `{}`, it is explicit vs implicit `undefined` :)

Btw, those `Promise.resolve` reformats weren't done by me, I guess you use some precommit tool tool :)

And, for your info, I noticed usage of `Array.from` in `proc.js` in `runAllEffect`, so I guess you would like it to be refactored at some point.

klis87:
@Andarist Regarding sparse arrays, what we could do for example is:
```javascript
const x = [, 1];
const y = [...x]; // y is [undefined, 1]
```

docs/api/README.md
klis87: @Andarist Any news on this? Btw, we can solve our our issue with `[...[, 1]]`, which will be converted to `[[undefined, 1]]`

Andarist: @klis87 Sorry for the long wait, I was really busy lately. I'm going atm through the issues and PRs, expect this to be handled soon.

Andarist:
Ok, so one need to be careful when dealing with sparse arrays - empty indices are non-enumerable and `deepEqual` uses `Object.keys` under the hood. Therefore we cannot simply compare explicit undefined with the implicit one in sparse array.

Personally I think that just suppressing eslint warning is enough and no more trickery should be added to that test.

It has also an interesting side-effect for our use case - returned array (if not destructured) remains sparse, so it potentially could surprise someone, but until somebody reports this with some valid use case why he/she would like to have it non-sparse, I dont consider this much of a problem.

Thanks for patience and the PR, gonna merge this in a moment.

klis87:
No problem, and thank you for you help and time with this PR. Regarding sparse array potential problem, I also don't think this would cause any issue in the future :)

Issue #570, opened 2016-10-05
[RFC] Supervision strategy for forked tasks (like actor systems)


slorber:
Hi,

**Edit**: if you are looking for a way to start root sagas, take a look at [this](https://github.com/redux-saga/redux-saga/issues/760#issuecomment-273737022)

@yelouafi we already discussed how close redux-saga looks to me like an actor system. I'm not an expert in actor systems, only know a bit of Akka, so I may say something wrong.
- Actors are organized in trees
- A spawned saga, (or root saga), is like a root actor
- A forked saga, is like a child actor.
- Redux channel is like the actor system event bus

A notable difference between redux-sagas and actor systems is that in actor systems it's currently more easy to handle adhoc communication. A created actor has a reference that can be passed around, and the parent generally hold those references. It's easy for example to send an event:
- Only to the parent (ie the forker saga)
- Only to the childs (ie the forked sagas) (or a subset)
- By "replying" to an incoming event (ie you don't have to know who emitted the event to reply to it) 

I don't know if it's worth implementing this part in Redux-saga yet, until someone has a real usecase for it, but there's also another aspect that Redux-saga does not have that actually would be really helpful: fault tolerance and supervision strategy.

When creating a child actor in Akka (ie forking a saga), you usually define a supervision strategy. The default one is `Escalate` which is somehow similar to the behavior of Redux-saga where the error bubbles-up, and will kill all the actor tree and the childs.

This behavior is fine in simple case, but according to issues and my own experience, it's pretty easy to make your whole app unusable when relying to fork, just because a single request failed. 

I think it would make sense, when forking a task, to define the behavior it should adopt regarding failure.

For example, if Redux-saga had the supervision feature of Akka, on a fork failure, one can choose:
- to propagate the error to the parent saga, which also has its own supervision strategy
- to resume the failing saga (not sure how it's applicable for Redux-saga)
- to restart the failing saga (maybe with an exponential backoff strategy)
- to stop the failing saga, without making the parent and brother sagas know about it

It's also possible to select the appropriate strategy according to the error.

Some Akka doc may be helpful:
- http://doc.akka.io/docs/akka/2.4.1/general/supervision.html
- http://doc.akka.io/docs/akka/snapshot/scala/fault-tolerance.html

[...]

PR #1205, opened 2017-10-11
Add retry helper

nuc:
Hey there,

As a follow-up on [that old discussion](https://github.com/redux-saga/redux-saga/pull/424), I tried to move the recipe inside a helper.

The only problem I'm facing is that I can't really catch the exception thrown inside the test, so a warning is logged when running it.

Any feedback would be really much appreciated.

Cheers,
Georgios


src/effects.js
```diff
@@ -20,4 +20,5 @@ export {
   takeEvery,
   takeLatest,
   throttle,
+  retry,
```
Andarist: im not yet sure if this should be treated as effect, but that's not important right now, I can easily decide about it after merging your PR in and just move the public export appropriately

src/internal/sagaHelpers/index.js
```diff
@@ -12,5 +13,6 @@ Therefore yield ${helperName} will return task descriptor to your saga and execu
 const takeEvery = /*#__PURE__*/ deprecate(takeEveryHelper, /*#__PURE__*/ deprecationWarning('takeEvery'))
 const takeLatest = /*#__PURE__*/ deprecate(takeLatestHelper, /*#__PURE__*/ deprecationWarning('takeLatest'))
 const throttle = /*#__PURE__*/ deprecate(throttleHelper, /*#__PURE__*/ deprecationWarning('throttle'))
+const retry = /*#__PURE__*/ deprecate(retryHelper, /*#__PURE__*/ deprecationWarning('retry'))
```
Andarist: `retry` would be a new thing, we don't need to provide old-style export for it with a deprecation warning

test/sagaHelpers/retry.js
```diff
@@ -0,0 +1,39 @@
+import test from 'tape'
+import lolex from 'lolex'
+import sagaMiddleware from '../../src'
+import { createStore, applyMiddleware } from 'redux'
+import { take, retry } from '../../src/effects'
+
+test('retry fails', assert => {
+  const clock = lolex.install()
```
Andarist: would prefer some simpler test case, not based so heavily on the clock

you can just put a counter variable here, increment with each worker call and just decide in the worker if based on the counter it should fail or succeed, and ofc based on it test retry behaviour

the clock needs to be ticked ofc, but only to move the test-timeline forward

src/internal/sagaHelpers/retry.js
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable no-constant-condition */
+import { call } from '../io'
+import { delay } from '../utils'
+
+export default function* retry(maxTries, delayMs, worker, ...args) {
```
Andarist: the problem with this implementation is that it is generator based and we cannot distribute generators yet in the source code, as old browsers/nodes dont support it, library should be pretty much ready for usage in any es5-compatible environment when downloaded from npm

we also cannot transpile generators ourselves as it would require opinionated runtime to be included in out bundle (most popular being `regenerator-runtime`) which tend to be quite big

for those reasons other helpers are created with `fsmIterator` internal helper - it's quite easy to use, once you know what it does - I have once described it [here](https://github.com/redux-saga/redux-saga/issues/570#issuecomment-256717377)

It doesn't yet support errors though, so it would have need to be adjusted so it could support `retry` helper

src/effects.js
nuc: Yeah, I also don't think it classifies as an effect.

test/sagaHelpers/retry.js
nuc: I used the same setup as in `throttle` but sure I can try to simplify it.

src/internal/sagaHelpers/retry.js
nuc: Gotcha. I thought the final output was transpiled. Sure, I can use the that iterator.

nuc:
@Andarist Thanks for the feedback. 

Sure, I'd like to continue working on it. I'll ping you once I have more.

Andarist:
If you have anything you would like to ask about, don't hesitate. You can always ping me here or on gitter.

ValentinBlokhin:
Any progress on this?

Havret:
Bump up, any progress?

Andarist:
I dont think @nuc is still working on this, anyone is free to pick this work up and get it to the finish line.

restrry:
merged in https://github.com/redux-saga/redux-saga/pull/1430

PR #1220, opened 2017-10-23
Add a correctly spelled arrayOfDeferred

sputt:
Fix a misspelling and provide update path


Andarist:
Tests are using `arrayOfDeferred` so it should be fixed there too.

sputt:
Great, I've renamed the function everywhere it's used

PR #1246, opened 2017-11-02
Effect middleware

Andarist:


Andarist:
@eloytoro im merging this in a minute, if you want you can compare ur PR to what I have done to it. Changes are rather minor and cosmetic, thanks again for your work and I'm sorry it took so long to merge it in.

Notable changes:
- Ive written separate tests for the feature
- if we use `redux`'s `compose` it needs to be a dependency and not a devDep, its easy to miss such thing especially that it had to be "upgraded" in a sense and during development it was available
- to simplify things, I've renamed it to `effectMiddlewares` and accept only an array of middlewares
- `compose` is used only once in the `runSaga` and not during each effect's execution, so its result is cached
- `createSagaMiddleware` accepts now that option too and pass it trough to the underlaying `runSaga` API

Issue #1075, opened 2017-07-14
Typescript Typings for result of `call` effect


Rokt33r:
To provide typings for result of call effect, I have to do like this.

```ts
  const repository: Repository = yield call<{slug: string, name: string}>(Repository.add as () => Promise<Repository>, {
    slug,
    name
  })
```

If we change the generic a little like:

```ts
// type Func1<T1> = (arg1: T1) => any;
type Func1<RESULT, T1> = (arg1: T1) => RESULT | Promise<RESULT>;

// <T1>(fn: CallEffectFn<Func1<T1>>, arg1: T1): R;
<RESULT, T1>(fn: CallEffectFn<Func1<RESULT, T1>>, arg1: T1): R;
```

Then, we will have much better format.

```ts
  const repository: Repository = yield call<Repository, {slug: string, name: string}>(Repository.add, {
    slug,
    name
  })
```

aikoven:
Why doesn't `Repository.add()` declare a correct return type in the first place?

Also, what's the benefit from this type casting? The types are lost anyway after the `yield`.

Rokt33r:
Yes, types will break after `yield`. But, we know they are not `any` and can be specific typed value because we are using `redux-saga`.

I've seen lots of mistakes like this in my company. That's the reason why I came here.

Rokt33r:
What I expected is: 

```
const {
  someValueOfRepo,
  anotherValueOfRepo
}: Repository = yield call<Repository>(Repository.add, name)
```

1. If `Repository.add` doesn't return an instance of `Repository` anymore, it will blame.
2. If `anotherValueOfRepo` is no removed from `Repository`, It will blame too.

aikoven:
I see, and this makes sense. But you would also have to declare types for all of the arguments. And yes, this is a breaking change. We could check back on this before v1.0

grovesNL:
When is v1.0? This would be useful.

Andarist:
@aikoven I'm preparing (slowly 😉) v1. Could you revisit this matter? Unfortunately I'm not able to make any decision regarding this and I trust you with TS related stuff here :)

Rokt33r:
@Andarist How can I track the changes for v1?

Andarist:
I've created [a milestone](https://github.com/redux-saga/redux-saga/milestone/1) where I gather all issues I would like to have resolved before releasing v1.

[...]

PR #1255, opened 2017-11-08
Update TypeScript typings for v1.0

aikoven:
- [x] Remove deprecated APIs
- [x] Update Channels, add `multicastChannel` and `stdChannel`
- [x] Fix `arrayOfDeferred` name
- [x] Update `runSaga` API
- [x] Support Effect Middleware
- [x] Update `Task` interface
- [x] Add generic parameters for return types of functions passed to `call`, `apply`, `cps` etc. (fixes https://github.com/redux-saga/redux-saga/issues/1075)


index.d.ts
```diff
@@ -63,67 +68,59 @@ export type Logger = (level: string, ...args: any[]) => void;
...
 ): SagaMiddleware<C>;
 
-
-type Unsubscribe = () => void;
-type Subscribe<T> = (cb: (input: T | END) => void) => Unsubscribe;
-
 export interface RunSagaOptions<A, S> {
-  context?: object;
-  subscribe?: Subscribe<A>;
+  channel?: PredicateTakeableChannel<A>;
```
Andarist: While technically this is correct, because saga will allow for it - the best option is to pass `stdChannel` here, maybe typings could enforce it? Dunno what best practices are in such cases.

Andarist:
Also I was wondering yesterday - should we maybe just move `delay` into `redux-saga/effects` and make it:
```js
export const delay = call.bind(null, _delay)
```
We've kinda made this for helpers in 0.14 and Im wondering if we should do the same for this one.

index.d.ts
aikoven: We can use `MulticastChannel` type here, though that wouldn't enforce exactly `stdChannel`.

aikoven:
> should we maybe just move `delay` into `redux-saga/effects`

👍 for me, in our projects, we only use `delay` inside `call`.

index.d.ts
Andarist: Hm, probably not worth it then. I guess that somebody might find it useful to pass in a unicast channel anyway, so it's better to not strangle one's creativity.

Andarist:
Gonna update the code + tests later with the change, but I guess if you have time you can already include it in this PR

Andarist:
[delay as effect](https://github.com/redux-saga/redux-saga/commit/eef6f89b4c59356ff99bbeee33cfd615b38039aa) 🎉

aikoven:
Updated typings.

Issue #1260, opened 2017-11-13
Feature request: inspection utils


mikesol:
Hey all,

A couple weeks ago I published [this article](https://hackernoon.com/telling-a-redux-saga-5fbfa26c81e7) on Hacker Noon and have since been asked if it is possible to make a github repo out of this idea.  Obviously the article is way too much of a hack, but I would like to try to make the work presentable. However, I would need to do it in a way that does not hack redux saga's internal API.

My idea is to create util classes in redux-saga that can determine what objects are (i.e. `isPut`, `isCall`) and get values associated with them (i.e. `getFn`, `getArgs`).  Then, I'd create a separate `redux-saga-teller` project that takes it from htere.  Does this seem like a reasonable addition to the code base?  It is minimally invasive and would be pretty easy to update as the API changes. If so, I can make a PR.

Andarist:
Isnt [`asEffect`](https://github.com/redux-saga/redux-saga/blob/658372c9cfbaf7435600e1372739e9c78c32b202/src/utils.js#L4) what you are looking for mostly?

mikesol:
Yup! The only thing then that we'd need is a public API function to get actions for `PUT`, functions and arguments for `CALL` and `FORK`, etc.. Also, `asEffect` is currently not documented in the API section - is it supposed to be public?

neurosnap:
Closing due to inactivity.  If this is still an issue then please reply and I will reopen and try to help.  Thanks!

Issue #1213, opened 2017-10-16
Monorepo - extract some modular packages


Andarist:
i.e. `buffers`, `channel`, `is` could be easily extracted and published for better modularity

restrry:
extractable modules:
- `delay` (moved from utils) - depends on `CANCEL symbol from redux-saga `
- `channel` depends on `utils, symbols, scheduler, matchers`
- `is` depends on `TASK & MULTICAST symbols`
could be extracted
- `scheduler`
- `buffers` 
- `matcher`
some modules could depend on symbols. so either we create circular dependency between modules or we extract symbols in a separate package as well

shinima:
Should we extract them into multiple modules like `@redux-saga/delay`, `@redux-saga/channel`, `@redux-saga/is`, or should we just extract them all into one extra module `@redux-saga/utils`?

If we extract them into multiple modules, then putting all symbols in `@redux-saga/symbols` seems good to me.

jeffvandyke:
Was there discussion somewhere I can't find where this was decided to be a good idea? Seeing this issue, I get very nervous about having to add about 5 new dependencies just for a basic usage of redux-saga - JavaScript package dependency lists are quite long enough. Is this direction of extracting separate packages always a good idea? And also, are the effects separate enough from internal implementation where it makes sense to extract them?

shinima:
@jeffvandyke I think this modular packages separation should be internal. Users only need to add redux-saga as the dependency.

By the way, I'm also curious about how we extract modular packages in the end, but at the moment we still have some space to improvement the modularity.

shinima:
@restrry  I think we could make `effectTypes` public. Further more, I'd like to remove `asEffect` and expose the structure of effects(though [it breaks encapsulation](https://github.com/redux-saga/redux-saga/pull/1477#issuecomment-395701484)), which results in much simpler usages like [this](https://github.com/redux-saga/redux-saga-devtools/blob/ec5c9a834171a301e638ac229fd49db2292fc986/src/store/selectors.js#L13-L60). I will make a PR to see if it makes a good difference.

Andarist:
> Should we extract them into multiple modules like @redux-saga/delay, @redux-saga/channel, @redux-saga/is, or should we just extract them all into one extra module @redux-saga/utils?

Former - more packages -> better modularization

> If we extract them into multiple modules, then putting all symbols in @redux-saga/symbols seems good to me.

👍 

[...]

PR #1262, opened 2017-11-13
Adds inspection to the public api

mikesol:
See https://github.com/redux-saga/redux-saga/issues/1260


Andarist:
Thanks for the PR. I find the idea useful and would like to merge this in in some form.

I'd like to finish https://github.com/redux-saga/redux-saga/pull/1263 first, as it has direct influence on your PR. There is not really much to do there, I'd like to hoist `effectRunners`, so the map can be reused, instead of creating a new one each time `redux-saga` gets an effect. Some effect runners depend on the variables from the closure, so I need to inspect each one, write them down, gather them under single state object so it could be passed to the respective effect runners in the place of `effectId` argument. If you want to help with that I highly encourage you to do so :)

Once the mentioned PR gets merged in, we could adjust your PR, review it some more and merge it in. Hope you understand.

src/internal/io.js
```diff
@@ -241,3 +241,26 @@ export const asEffect = {
   getContext: createAsEffectType(GET_CONTEXT),
   setContext: createAsEffectType(SET_CONTEXT),
 }
+
+const nullOrWhat = (what, obj) => obj && (!what ? obj : obj[what])
+
+const _get = (what, ...args) => effect =>
+  args.reduce((maybe, type) => maybe || nullOrWhat(what, asEffect[type](effect)), null)
+
+export const inspection = {
```
Andarist: in general it's way better not to export objects, when they expose only static methods, as exporting objects in most cases prevents tree-shaking

src/utils.js
```diff
@@ -1,5 +1,5 @@
 export { delay, noop, is, deferred, arrayOfDeferred, createMockTask, cloneableGenerator } from './internal/utils'
 
 export { SAGA_ACTION, TASK } from './internal/symbols'
-export { asEffect } from './internal/io'
+export { asEffect, inspection } from './internal/io'
```
Andarist: would creating a new entry point - `redux-saga/inspection` (like `/effects` and `/utils`) make sense for you here?

I'd like also to split `redux-saga` in few packages https://github.com/redux-saga/redux-saga/issues/1213 . However for the time being `redux-saga` would mainly stay the same, reexporting other packages etc. The reason for the change is just being able to install subpackages in other projects when the `redux-saga` itself is not needed. In such case - should `inspection` become other official subpackage or do you see it as part of the core?

mikesol: Thanks for the feedback! I've taken a stab at reorganizing it based on your recommendations, but as I don't have a lot of JavaScript experience, it may not be what you're looking after (I didn't know what tree-shaking was, for example, and still am a bit fuzzy on the concept after reading about it).  Lemme know!

mikesol:
Hey! Sorry, I missed your initial message about #1263 . That looks like a really cool project! Unfortunately it's a bit too much for me to chew right now - mostly, we are focusing on open sourcing our code base at Meeshkan, but once we have a bit more time, we'd love to contribute to redux-saga in a more general way!  With respect to our saga-teller project, we've refactored it to take a general matcher function as an argument, so there is nothing in the base functionality that depends on a non-public API. Of course, the matcher function will need to ping the internal api for the time being, but that is easy to change.

Andarist:
Ok, I will have this PR hanging around here for some time, until I finish #1263 and then I can even just make necessary adjustments and merge it in. Thanks for the work!

Andarist:
Thanks for the PR, after procrastinating on it too much and rethinking some things - we've decided that effects' shapes are stable enough to recommend just checking them like regular objects if needed, you can build ur own inspection utils if you want ofc, but we dont feel its particularly useful as part of the core library, especially considering the use cases for inspection rather niche.

mikesol:
Agreed - if the API is stable, there's no reason to merge this PR as writing object verification code is trivial. The only thing I'd ask then, if you guys have time, is to make this a documented feature when you're ready: even a one-liner in the docs saying that the internal representation of objects is stable and that any backwards-incompatible change will be reported in release notes.  But of course that's low priority compared to the other stuff you have going on. Great work, as usual!

restrry:
makes sense, I added a note to CHANGELOG.md https://github.com/redux-saga/redux-saga/pull/1700/commits/473cabf70c8be46780a511fabe3eaf4640bf6fb8

PR #1263, opened 2017-11-14
Formalizing saga's state

Andarist:


src/internal/io.js
```diff
@@ -2,26 +2,26 @@ import { IO, SELF_CANCELLATION } from './symbols'
...
+export const CANCELLED = 'CANCELLED'
+export const FLUSH = 'FLUSH'
+export const GET_CONTEXT = 'GET_CONTEXT'
+export const SET_CONTEXT = 'SET_CONTEXT'
 
 const TEST_HINT =
   '\n(HINT: if you are getting this errors in tests, consider using createMockTask from redux-saga/utils)'
 
-const effect = (type, payload) => ({ [IO]: true, [type]: payload })
+const effect = (type, payload) => ({ [IO]: type, [type]: payload })
```
yelouafi: this is a refactoring I wanted to do long time ago.

BTW why dynamic `[type]` property and not just `type` and `payload` properties ? (would also makes inspection easier in devtools) . the purpose of `[IO]` was to avoid name clashes as the user could yield any object so needed

Andarist: hi @yelouafi ! 👋 

if we remove `IO` won't it create the same problem? How today's situation is different in terms of name clashes?

Andarist: also - without `IO` its hard to determine if the object is an effect for the purpose of introduced `is.effect`, hard to distinguish it from any other `{ type, payload }` objects, its a popular shape i.e. for redux actions

Andarist: OTOH we could do `{ [IO]: type, payload }` instead of the dynamic `type` property

yelouafi: I meant 3 props : 

```js
{
  [IO]: true  //as a signature
  type: TAKE,
  payload: ...
}
```

then it could be consumed like

```js
if(obj[IO] === true) {
  switch(obj.type) {
    case TAKE:
    ...
  }
}
```

the first test could maybe be put behind a little function `isEffect(obj)`

Andarist: Yeah, that would totally work. Gonna play with it later. Thanks for the input, it's nice to know that you still have an eye on the project 👍 

I've started to refactor a little bit to extract things (like effect runners) from the `proc`'s closure - but as things, especially those runners, access closure's variables I've started to look into what they access and I will try to gather those things under single saga's `state` variable - this potentially could unlock possibility to create custom effects in the user land.

Andarist:
Note to myself: I have no idea how `task` and mainTask` work in the `proc`. Ideally (?) we should be able to hide both in 1 abstraction.

Andarist:
Idea - is effect serializability a goal? By encapsulating effect runner in an effect we could slim down the core library, because the interpreter wouldn't have to know upfront about all possible effects and their respective runners.

And effects would still be somewhat serializable, such `JSON.stringify` would simply omit a runner key.

yelouafi:
> I have no idea how task and mainTask work in the proc

for example if we have this saga

```js
function* parent() {
  yield fork(child1)
  yield fork(child2)
  yield call(someApi)
}

// inside another saga
const parentTask = yield fork(parent)
```

`task` is basically the descriptor for for `parentTask`. It will terminate after the `yield call(someApi)` resolves but will also wait for the 2 previously forked tasks to terminate.

`mainTask` represents only the sequence of 3 yields in parent's body. It ends as soon as the `yield call(someApi)` resolves. I used it to track when the body of the generator has terminated. Using a task internally allows `forkQueue` code to be generic. I just add mainTask here

https://github.com/redux-saga/redux-saga/blob/441b00959d51c2edc9efd4bdf7216d308f56344c/src/internal/proc.js#L61

If later attached forks are added then all tasks (mainTask + attached forks) are tracked using the same logic

PR #1265, opened 2017-11-19
Fix export error in rootSaga section of tutorial

jaimemendozadev:
The original tutorial shows the `rootSaga` function being exported via `export default`. For beginners, docs don't specify you need to remove the `export` keyword from `incrementAsync` and `watchIncrementAsync` functions in `sagas.js` if you're going to use `export default function* rootSaga()`.


Andarist:
Thanks for the PR. Im wondering if doing those snippets with `diff` would be beneficial, like:
```diff
-export function* incrementAsync() {
+function* incrementAsync() {
  yield delay(1000)
  yield put({ type: 'INCREMENT' })
}
```

jaimemendozadev:
Thanks for merging it! I leave the `diff` up to your good judgment!

PR #1273, opened 2017-11-27
Updated testing documentation

alanraison:
An attempt to add some detail and clarity to the testing documentation.


Andarist:
Love this contribution ❤️

I think you might be interested in `effectMiddlewares` feature introduced in [v1.0.0-beta.0](https://github.com/redux-saga/redux-saga/releases/tag/v1.0.0-beta.0), which would allow you to resolve effects with custom data instead of mocking the API (although in the end those are pretty similar things)

Issue #1130, opened 2017-08-15
[bug] Forking an effect creator causes it to run


Andarist:
Reproduced issue: https://codesandbox.io/s/68rm79Qg7

**yield fork(call, fn)**
its ok, as `call` is just a function so it shouldnt be treated any different

**yield fork(() => call(fn))**
its not ok, `fn` gets executed, returned effect is interpreted, it should not happen, may lead to unexpected behaviours

**yield fork(takeEvery, 'pattern', fn)**
its not ok, `takeEvery` uses `fork` under the hood, and it seems that nested `fork(fork, fn, ...)` gets 'unwrapped' and unwrapped task descriptor (one for `takeEvery('pattern', fn)` / `fork(fn, ...)`) gets returned to the caller saga.

Probably similar thing happens when returning iterators - need to be checked.

Returning promises / iterators from `call` is working, but should be treated as special cases.

Andarist:
**yield fork(() => call(fn))** case need to be checked further. But I suspect its as simple as - returned value gets passed to the interpreter.

Im still a little bit unsure if it should get interpreted. On one hand its just a "yielded" value with a `{done: true}`, OTOH its a little bit magical and might be unexpected, its quite implicit behaviour.

I.e. for of loops ignore the "done" values, but i.e. Promises auto-flatten themselves when chained (also not sure if that behaviour is good on Promises)

Andarist:
somewhat related https://github.com/TheNavigateur/proposal-pipeline-operator-for-function-composition/issues/3

nhducit:
currently we can use **yield fork(call, fn)** instead of **yield fork(() => call(fn))** to avoid this bug. How about **yield fork(takeEvery, 'pattern', fn)**? Do we have any workaround solution?

Andarist:
Just don't use any of those, those effects are not intended to be used like that.

restrry:
I think problem with `call` is related to [this problem as well](https://github.com/redux-saga/redux-saga/pull/1017). we need to align approach for both cases @shinima @Andarist

shinima:
> Just don't use any of those, those effects are not intended to be used like that.  👍

We should always **create a forking effect** instead of **forking an effect creator**.


[...]

PR #1283, opened 2017-12-05
allow thunks in delay value

caub:
This commit will simply allow to do:

```js
const session = yield call(delay, 1000, call(fetchApi, '/session'));
```
instead of 
```js
const [session] = yield all([
	call(fetchApi, '/session'),
	call(delay, 1000)
]);
```
(to request the session, with a duration of at least 1000ms (enforce a minimal loading delay))

and can be tested the same way, more easily since `all` returns an array

```js
assert.deepEqual(saga.next().value, call(delay, 1000, call(fetchApi, '/session')));
```
vs
```js
assert.deepEqual(saga.next().value, all([
	call(fetchApi, '/session'),
	call(delay, 1000)
]));
```

### DIFF
--- a/effects.d.ts
+++ b/effects.d.ts
@@ -548,4 +548,4 @@ export function throttle<A extends Action, T1, T2, T3, T4, T5, T6>(
   arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5, arg6: T6,
   ...rest: any[]): ForkEffect;
 
-  export function delay<T>(ms: number, val?: T): CallEffect;
+export function delay<T>(ms: number, val?: T | (() => T)): CallEffect;
--- a/src/internal/utils.js
+++ b/src/internal/utils.js
@@ -98,8 +98,9 @@ export function arrayOfDeferred(length) {
 
 export function delay(ms, val = true) {
   let timeoutId
+  let v = typeof val === 'function' ? val() : val
   const promise = new Promise(resolve => {
-    timeoutId = setTimeout(() => resolve(val), ms)
+    timeoutId = setTimeout(() => resolve(v), ms)
   })
 
   promise[CANCEL] = () => clearTimeout(timeoutId)
--- a/test/typescript/delay.ts
+++ b/test/typescript/delay.ts
@@ -12,4 +12,10 @@ function testDelay() {
     const e: boolean = res;
     const r: string = res;
   });
+
+  delay(1, () => 'foo').then(res => {
+    // typings:expect-error
+    const e: boolean = res;
+    const r: string = res;
+  });
 }
--- a/utils.d.ts
+++ b/utils.d.ts
@@ -12,7 +12,7 @@ import {
 import {Task, Channel, Buffer, SagaIterator} from "./index";
 
 export function delay(ms: number): Promise<true>;
-export function delay<T>(ms: number, val: T): Promise<T>;
+export function delay<T>(ms: number, val: T | (() => T)): Promise<T>;
 
 export const TASK: string | symbol;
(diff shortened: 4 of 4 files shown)


caub:
I'm not familiar with TypeScript, what signature should I put to allow to pass a thunk in second arg?

```ts
export function delay<T>(ms: number, val?: T): CallEffect;
```

test: 

```ts
delay(1, () => 'foo').then(res => {
  // typings:expect-error
  const e: boolean = res;
  const r: string = res;
});
```

Andarist:
Hi! Thanks for the PR. Could you specify a use case for this? I don't see any advantage in the given scenario of nested `call`s vs `all`. 

I personally also ain't sure if returned effects (not yielded ones) should resolve automatically (related https://github.com/redux-saga/redux-saga/issues/1130). 

And also I'm pretty sure that the actual function call should be moved into the `setTimeout` if we want to add this (I realize though that it's probably not what you had in mind.)

caub:
The usecase I had is in first comment. Currently I'm doing
```js
all([
	call(fetchApi, '/session'),
	call(delay, 1000)
])
```
but it's less practical, less easy to test and more verbose than 
```js
call(delay, 1000, call(fetchApi, '/session'))
```
because here's how I can test the all approach: (`eq` is `assert.deepEqual`)
```js
const saga = getSession();
eq(saga.next().value, all([
	call(fetchApi, '/session'),
	call(delay, 1000)
]));
eq(saga.next([session]).value, put({type: SIGNIN, value: session}));
```
notice the `[session]` because, that's the form returned by `all`

In comparison with delay + value:
```js
const saga = getSession();
eq(saga.next().value, call(delay, 1000, call(fetchApi, '/session')));
eq(saga.next(session).value, put({type: SIGNIN, value: session}));
```

Andarist:
IMHO this use case is obscuring the saga mechanism for a little benefit of slightly shorter/terser syntax. 

If you want to use it in such form I'd really prefer you creating a custom helper, don't have a perfect name for it, but the implementation would be:
```js
const callWithPostponedResult = (ms, saga, ...args) => all([call(delay, ms), call(saga, ...args)])
```

The only thing it doesn't do is unwrapping your result from the array, but that could also be achieved with other implementation:
```js
const postponedResult = (ms, saga, ...args) => function* () {
	const res = yield all([call(delay, ms), call(saga, ...args)])
	return res[1]
}

yield call(postponedResult, 1000, fetchApi, '/session')
```

You could possibly also migrate to `v1-beta` version and use `effectMiddlewares` option to ease your integration or wait for custom effects feature (dunno when it is gonna happen, first step PR is [here](https://github.com/redux-saga/redux-saga/pull/1263)).

caub:
Well, I don't know for now, no worry it's not important
I would have preferred to avoid another function, there are already many

and also because the current delay already works with things like `delay(500, delay(200, 'test'))`, this promise gives 'test' after 500ms, so my PR wasn't really changing this current behavior. I just needed to use `call` for tests

Andarist:
> and also because the current delay already works with things like delay(500, delay(200, 'test')), this promise gives 'test' after 500ms, so my PR wasn't really changing this current behavior. I just needed to use call for tests

It probably would also work with `delay(1000, call(fetchApi, '/session'))` - although it wouldn't act as `all` but rather like a waterfall - the call would get executed after 1 second.

That's however just a side effect of how redux-saga works under the hood. There is no intention for any of this pattern to actually work (`delay(500, delay(200, 'test'))` and `delay(1000, call(fetchApi, '/session'))`)

caub:
oh yes, indeed thanks
closing so

Issue #1240, opened 2017-10-30
redux-saga doesn't play well with webpack/sinon/mocha


frankandrobot:
In the sample code below, `sagaWithUndefinedError` has an undefined variable. The first problem is that the error thrown is `'uncaught', 'at _callee` (and not an "undefined error"). This is exacerbated by the fact that the stack trace points to the line number in the bundle created by webpack and not the line number in the original source code. (Note that non-saga code does not have this problem). 

So to sum up
1. Error messages aren't that helpful.
2. Stack traces don't seem to be aware of source maps and instead point to the webpack bundle.

As you can imagine, this makes it a little hard to write unit tests for sagas. As a workaround, we've resorted to asserting all dependencies and manually logging `error.message`s (which seem to be more helpful). 

Oh and the code below uses the `rstudio/redux-mock-store-middleware`, which I'm making public right now. 

```js
describe('stupidly simple saga', function() {
  function *sagaWithUndefinedError() {
    notDefinedError;
    yield put({type: 'go!'});
  }
  const sagaMiddleware = createSagaMiddleware();
  const middlewares = [sagaMiddleware];
  mockStore(undefined, middlewares);

  it('should point to original source code', function() {
    sagaMiddleware.run(function *() {
      yield* sagaWithUndefinedError();
    });
  });
});
```

### stack trace
```bash
ERROR: 'uncaught', 'at _callee
 sagaWithUndefinedError$@http://localhost:9876/base/test.webpack.js?df53c259522a5a6adb603b3cf28e680ed13619e5:201633:13
tryCatch@http://localhost:9876/base/test.webpack.js?df53c259522a5a6adb603b3cf28e680ed13619e5:175617:37
invoke@http://localhost:9876/base/test.webpack.js?df53c259522a5a6adb603b3cf28e680ed13619e5:175851:22
defineIteratorMethods/</prototype[method]@http://localhost:9876/base/test.webpack.js?df53c259522a5a6adb603b3cf28e680ed13619e5:175669:16
tryCatch@http://localhost:9876/base/test.webpack.js?df53c259522a5a6adb603b3cf28e680ed13619e5:175617:37
maybeInvokeDelegate@http://localhost:9876/base/test.webpack.js?df53c259522a5a6adb603b3cf28e680ed13619e5:175913:18
invoke@http://localhost:9876/base/test.webpack.js?df53c259522a5a6adb603b3cf28e680ed13619e5:175825:32
[...]

PR #1289, opened 2017-12-08
remove custom error wrapping

jonhester:
My stack traces with webpack and babel were a lot less useful after adding redux-saga to my application since they never pointed to the webpack bundle instead of the original source files. I removed the error wrapping as discussed in #1240. Getting rid of the error wrapping fixed this for me so maybe it will at least fix that part of @frankandrobot's issue as well.

### DIFF
--- a/src/internal/proc.js
+++ b/src/internal/proc.js
@@ -167,11 +167,7 @@ export default function proc(
   const logError = err => {
     let message = err.sagaStack
 
-    if (!message && err.stack) {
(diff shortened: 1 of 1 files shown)


Andarist:
We should go even further - removing `sagaStack` entirely, that way `logError` probably could be removed too.

jonhester:
Thanks for looking at this so quickly. I removed `sagaStack` and the `logError` method. I'm just calling `log('error', error)` now instead.

Andarist:
Thanks, I need to revisit each of those calls some time to determine if they are needed there, but that's a matter of followup PR to this.

Issue #1252, opened 2017-11-05
Make gh-pages branch protected


Andarist:
I've deleted by accident today `gh-pages` branch which contains the documentation and automatically our docs site went down.

Would be desirable to have this branch protected, but this mean we can't force push to it and that's what currently [our publication script](https://github.com/redux-saga/redux-saga/blob/1877d2ea3173e61b01407edf6475f25116b386df/package.json#L42) does. It needs to be tweaked to comit the content and push it so we could make `gh-pages` protected.

lukePeavey:
@Andarist Is this just a matter of removing the `--force` flag from the `push` command, or is it more involved than that?

Andarist:
I imagine it requires removing the `--force` flag and changing the behaviour of the script to commit new content each time

lukePeavey:
Thanks for the response.  I wasn't able to figure out how to implement this.

Andarist:
I got an advice from a friend who knows git way better than me, that it would probably require:
- `GIT_WORK_TREE=docs/ git checkout gh-pages` - which should unpack `gh-pages` into `docs`
- `GIT_WORK_TREE=docs/ git commit -am 'update book'`
- and then just do a push

This recipe is probably missing something, but at least it is a suggestion what would need to be explored to implement this.

lukePeavey:
Ok, that definitely helps! At first glance I didn't fully understand what was going on in the `docs:publish` script, but after breaking it down and doing some reading about gitbooks deployment, i think i have a better handle on it. 

Im going to some work on this and see if I can figure it out.

lukePeavey:
So far, I haven't been able to get this working with the `GIT_WORK_TREE` approach. I did find another possible solution using the [gh-pages module](https://www.npmjs.com/package/gh-pages). I came across it in [this article](https://medium.com/@gpbl/how-to-use-gitbook-to-publish-docs-for-your-open-source-npm-packages-465dd8d5bfba).

It accomplishes the same thing as the current publication script, but doesn't use `--force` push to overwrite the `gh-pages` branch each time. 

From the documentation
> It will create a temporary clone of the current repository, create a gh-pages branch if one doesn't already exist, copy over all files from the base path, or only those that match patterns from the optional src configuration, commit all changes, and push to the origin remote.
>
>If a gh-pages branch already exists, it will be updated with all commits from the remote before adding any commits from the provided src files.

The publication script would be changed to
```sh
[...]

PR #1290, opened 2017-12-10
Modify publication script so gh-pages can be protected

lukePeavey:
Possible solution for #1252

I modified the publication script to use the [gh-pages](https://github.com/tschaub/gh-pages) package to deploy the documentation site. It performs the same task as the current script but doesn't force push to gh-pages. Instead, a new commit will be added each time the script is run. 

I've tested this and it seems to be working fine. The branch generated by the new script is identical to the current one. 

There is one caveat...  if the gh-pages branch is deleted from the remote repository, you need to run `gh-pages-clean` to clear the package's cache before running the publish script again. Otherwise you'll get an error. (see [this issue](https://github.com/tschaub/gh-pages/issues/17)).

This can be run manually if needed `node_modules/.bin/gh-pages-clean`

Another possibility would be to add it to the `docs:clean` script so it runs automatically before each publish. This makes the script a bit slower (because `gh-pages` will create a temporary clone of the repo each time instead of using the cached one). 

Let me know if you questions/thoughts. Im happy to provide additional info or update the PR.


Andarist:
The changes are fine as far as I can tell, if they are working :)

> There is one caveat... if the gh-pages branch is deleted from the remote repository, you need to run gh-pages-clean to clear the package's cache before running the publish script again. Otherwise you'll get an error. (see this issue).

Can this become a problem? The whole point is to make the remote branch protected, so it won't be possible to delete it.

lukePeavey:
>Can this become a problem? The whole point is to make the remote branch protected, so it won't be possible to delete it.

Its pretty much a non-issue since branch will protected. I just thought i should mention it as something to be aware of (in case you ever had to take the branch down for some reason)

lukePeavey:
I've tested the script on [this repository](https://github.com/lukePeavey/redux-saga-clone/) if you want to take a look.

Andarist:
Cool, gonna merge in and see how it plays out. Thanks for your hard work, I really appreciate it 👍

lukePeavey:
Glad i was able to contribute!! I just hope it works

PR #1296, opened 2017-12-19
Link to effects list from "Declarative Effects"

willem-h:
While using redux-saga in my own work I've noticed that I Google for
"redux effects" expecting to find a list of the effects that are
available. The first result links to the [Declarative Effects](https://redux-saga.js.org/docs/basics/DeclarativeEffects.html) page
which doesn't contain a list or a link to one. This leads to me racking
my brain to figure out where I saw that list last time.

Adding a link in the first paragraph and another at the bottom of the
page fixes this slight frustration.


Andarist:
Thanks!

Issue #1199, opened 2017-10-07
[DX] Babel plugin


Andarist:
It should be possible to create babel plugin to provide better error stack traces for our sagas in the similar way React and others do it with component stack traces.

neurosnap:
This would be a great addition.  Often times our error stack traces from redux-saga are not helpful.

Andarist:
@neurosnap what information such babel-enhanced error message could possibly have? I thought especially about saga traces - each `call`, `fork`, `spawn` could be tracked and the stack from the root saga could be reported.

Andarist:
This should work very similar to:
https://babeljs.io/docs/plugins/transform-react-jsx-source/
https://babeljs.io/docs/plugins/transform-react-jsx-self/

Handling those things in the redux-saga itself in the development mode is ofc also needed.

restrry:
what's your idea behind that plugin?
what kind of information it has to log?

1. place where effect was created?  that'll work for errors that occurs during calling effect creator, but won't work if error occurs in saga's body 
```js
function* a(){
    yield call(() => unitilisedVariable()); <- can be logged as error when `call` effect was created
}
function* b(){
    unitilisedVariable() <- can't be identified, since occurs not in effect creator. what should be included in error stack in that case?
}


```

2. or place where saga function was created?
```js
function* a(){
     unitilisedVariable() <- now we can add in stack trace info where function a* was declared
}
```

3. both ?

Andarist:
I haven't thought this through yet, but the general idea was to log saga branch - from the root to the failing effect, creating sort of a stack from it, but not exactly the error stack, but the saga stack (similar to what React does for component stacks).

While I know how to manipulate babel's AST to some extent, I have not yet integrated anything with source maps and I guess we are looking for some sort of such integration.

For starters we could annotate each generator (we could detect if it yields an effect, but this could potentially cause some false negatives and annotating every generator shouldn't interfere with anything imho) with source line and inferred name with static properties. In addition we'd have to attach to each yielded value (let's assume everything is an effect) "parent saga" static annotations. That way those helpful infos could flow through the saga's runtime and it could collect it in case of an error to construct a custom error message.

restrry:
I started working on the plugin https://github.com/restrry/babel-plugin-transform-redux-saga-source
but still open for new ideas and need help to integrate it in redux-saga itself

russiancow:
[...]

PR #1310, opened 2018-01-18
[WIP] add support for error stack enhancer babel plugin

restrry:
original issue https://github.com/redux-saga/redux-saga/issues/1199

PR just to start discussion how it's better implement babel plugin and add support in redux-saga core

## TODO
- [ ] decide how to handle `meta`
- [ ] add tests

## Note
links aren't clickable yet. because there are several possibilities
- open file in devtools, but we need to understand how to map path to bundler URI (`webpack://`, for example)
- open file in text editor, requires local server

## Screens
error message with babel plugin

error message without plugin


src/internal/proc.js
```diff
@@ -151,6 +151,19 @@ function createTaskIterator({ context, fn, args }) {
       )
 }
 
+function getLocation(instrumented) {
+  return (
+    // TODO move symbol creation to redux-saga side
+    instrumented[Symbol.for('babel-plugin-transform-redux-saga-source')] || instrumented.__source || null
```
restrry: NOTE: we can add symbol no to extend saga effect descriptor. the most simple first step is add global symbol

Andarist: well, it could also always be importable from the `redux-saga`, so we dont have to rely on globals

src/internal/proc.js
```diff
@@ -178,9 +192,10 @@ export default function proc(
     Creates a new task descriptor for this generator, We'll also create a main task
     to track the main flow (besides other forked tasks)
   **/
-  const task = newTask(parentEffectId, name, iterator, cont)
-  const mainTask = { name, cancel: cancelMain, isRunning: true }
-  const taskQueue = forkQueue(name, mainTask, end)
+  const task = newTask(parentEffectId, meta, iterator, cont)
+  // NOTE: not sure where tasks are used, so left the mainTask interface as is
```
Andarist: what do you mean by "where tasks are used"? how did it matter for the implementation?

Andarist:
Gr8 work! I really like this idea (as you know). Originally I had something like this in mind - https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html#component-stack-traces , just for "sagas traces". I'd suspect it being more noticeable by the users.

Open questions:
- is this [technique](https://github.com/facebook/react/blob/fccd11bec00f64ccd455163634739d1a4b40020b/packages/shared/invokeGuardedCallback.js#L64) useful for us?
- would it be possible to hook into [react-error-overlay](https://www.npmjs.com/package/react-error-overlay) with our errors? 

I'll give this a spin this weekend, to see how it actually works - I know that I've promised to do it before, just having loads of unexpected stuff in my life lately and didn't have my head in the game here.

src/internal/proc.js
restrry: agree, I've even checked how [lodash-webpack-plugin](how https://github.com/lodash/lodash-webpack-plugin/blob/master/src/index.js#L48-L86) does it. but haven't tried to apply that technique.

src/internal/proc.js
restrry: On the one hand, we don't use `name` as independent property, but part of `meta` (code consistency). On the other hand, I'm not sure whether we need to create `meta` descriptor for `mainTask` (what's location in that case?)

restrry:
1. firstly, we need to decide what information we want to show for user. imho, that should be effect + task information in case of error, since exception could be thrown from effect creator and from task body.
deciding on content, we can format error as we want.

2.
> - is this technique useful for us?
> - would it be possible to hook into react-error-overlay with our errors?

I'll check and get back when PoC is ready

thanx for feedback!

src/internal/proc.js
Andarist: Hm, I think it would really be a matter of inserting an import for appropriate export from redux-saga package. It seems that lodash is doing some more magic there in the quoted code - for some other reasons.

src/internal/proc.js
Andarist: `mainTask` is completely internal, see [this comment](https://github.com/redux-saga/redux-saga/pull/1263#issuecomment-347485969). I dont think it needs to have any meta attached to it (regarding location etc)

Andarist:
1. gonna prepare a proposal of the content/format for various cases (although ideally they should boil down to a single format, dunno if that will be possible at the moment though)

2. cool, both of those are possible enhancements and not the core of this feature anyway

Andarist:
Simple example:
```js
const someAPI = () => {
	if (Math.random() > 0.5) {
		throw new Error('example error')
	}

	return 42
}

function* fetchSaga() {
	yield call(someAPI)
}

function* rootSaga() {
	yield fork(fetchSaga)
}
```

should create similar message:
```
The above error occurred when executing "call(someAPI)" effect:
	in call(someAPI)
	created by fetchSaga
	created by rootSaga
```

It seems to me that most valuable are:
- failed effect
- "saga trace" - only names of chained "parent" tasks

At the moment it seems to me that displaying effects that created certain sagas would just clutter the message, alternative message could be created:
```js
The above error occurred when executing "call(someAPI)" effect:
	in call(someAPI) (at fetchSaga)
	created by fork(fetchSaga) (at rootSaga)
```

Thoughts?

Additionally we could list sagas that got cancelled because of the error (siblings of the failed top most saga in the tree).

src/internal/proc.js
```diff
@@ -151,6 +151,19 @@ function createTaskIterator({ context, fn, args }) {
...
+  )
+}
+export function getMetaInfo(fn) {
+  return {
+    name: fn.name || 'anonymous',
+    location: getLocation(fn),
+  }
+}
+
 export default function proc(
```
Andarist: it seems to me that we could accept saga (generator) instead of iterator here and just use saga's static properties later when needed - no need for meta object as arguments, we could even (possibly) eliminate passing explicit `name` arg

Andarist: looking through the code now and im not sure if it can actually be done - but would certainly be worth checking

restrry: I looked through the `proc` code and not sure that's possible without refactoring, so I'd prefer keep that out of scope of that PR

restrry:
@Andarist could you check the current version? I added example that you could run with `npm run error-demo` (could be deleted before merge, also clean up package.json)
output for the current version
- sync error in put effect


- error in call effect


Notes:
- if we don't show effects for the whole failed sagas stack it could be hard to understand which effect created the failed task (especially if we have conditions in the code or `race`. on the other hand, saga shouldn't be so huge to have such a problem)
- we format failed saga stack and log it only when error bubbles to the root, probably we should just update `sagaStack`  while error is bubbling and let user decide what to do with error stack (imho, `sagaStack` should be provided to users, in our app we use it to report error in sagas)
- I also prefer to simplify approach with naming & meta info and get rid of  `getMetaInfo`

regarding:
> is this technique useful for us?
> would it be possible to hook into react-error-overlay with our errors?
> Additionally we could list sagas that got cancelled because of the error (siblings of the failed top most saga in the tree).
still working on it, could be done in follow up requests

Andarist:
> if we don't show effects for the whole failed sagas stack it could be hard to understand which effect created the failed task (especially if we have conditions in the code or race. on the other hand, saga shouldn't be so huge to have such a problem)

Fair point, although source mapped line numbers, should mitigate this issue, right? I'm not really holding onto any of the proposed formats.

> we format failed saga stack and log it only when error bubbles to the root

We should log it whenever detached or root saga fails - you could very well mean the same, just thought this needs clarification, so we are on the same page.

> probably we should just update sagaStack while error is bubbling and let user decide what to do with error stack (imho, sagaStack should be provided to users, in our app we use it to report error in sagas)

I don't really feel that comfortable about extending built-ins such as `Error`. Seems a little bit clunky, maybe we could just pass both original javascript error and `sagaStack` to the `onError` handler? If it would bring more complexity than it is worth, then we certainly shouldn't do it and we should just stick to the `sagaStack`.

> still working on it, could be done in follow up requests

Sure thing, those are just possible nice enhancements, not the core of the feature.

**Note**
I haven't yet read the new code, just replying to your post.

restrry:
> We should log it whenever detached or root saga fails - you could very well mean the same, just thought this needs clarification, so we are on the same page.

correct

> If it would bring more complexity than it is worth, 
the cheapest way that I seen could be to use Symbol to extend error object with custom data and after formatting saga stack pass 2 values in userland - error object and formatted saga stack as string. other ideas that I thought about require significant changes in the `proc.js`, but I'm still not so comfortable with its source code

restrry:
current implementation doesn't show location for effects created via `fmsIterator`, there are several possible solutions:
- show place where effect was emitted
- show just taskIterator name, but since `runForkEffect` is used for both iterators & generators, we have to change meta calculating process, something like 
```js
const meta = taskIterator.name ? { name: taskIterator.name  } : getMetaInfo(fn);
```

not sure what is preferred option

Andarist:
I've looked through the code and it's certainly nice that we can get some readable output even without babel plugin!

Need to check out how React handles this without babel plugin and with it, and what benefits [jsx-self](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-react-jsx-self) gives to the error output.

I'd like to re-check if we could somehow omit passing those meta objects around and maybe pass something else around that would already be tied to meta, like this `__self` or something.

This might not be possible though or would rely on babel plugin entirely and probably babel should only enhance our error messages and not be a requirement for them as the cost of having this in runtime seems to be negligible. 

Some play testing is still ahead of me, but it looks great and I hope to wrap this up in following week, let's ship it soon 🚀

src/internal/proc.js
```diff
@@ -105,8 +124,8 @@ function forkQueue(name, mainTask, cb) {
     addTask,
     cancelAll,
     abort,
-    getTasks: () => tasks,
-    taskNames: () => tasks.map(t => t.name),
```
restrry: didn't find that it was used somewhere, so renamed

src/internal/utils.js
```diff
@@ -134,7 +134,7 @@ const kThrow = err => {
 }
 const kReturn = value => ({ value, done: true })
 export function makeIterator(next, thro = kThrow, name = 'iterator') {
-  const iterator = { name, next, throw: thro, return: kReturn }
+  const iterator = { meta: { name }, next, throw: thro, return: kReturn, isSagaIterator: true }
```
restrry: had to wrap in `meta`  in sake of consistency, probably it's better to flatten `meta` completely (switch back to name & location properties)

restrry:
> Need to check out how React handles this without babel plugin and with it, and what benefits jsx-self gives to the error output.
if I got correct [jsx-source](https://www.npmjs.com/package/babel-plugin-transform-react-jsx-source) adds location information, [jsx-self](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-react-jsx-self) returns component descriptor itself, could be useful if we wanna get failed effect arguments, for example
 
> I'd like to re-check if we could somehow omit passing those meta objects around and maybe pass something else around that would already be tied to meta, like this __self or something.
I don't like how we handle `meta` either, probably you have idea how to improve the code

> This might not be possible though or would rely on babel plugin entirely and probably babel should only enhance our error messages and not be a requirement for them as the cost of having this in runtime seems to be negligible.
Plugin should be used in dev mode only, so don't think that affects prod runtime a lot.

I'd say that current feature state is MVP enough to test it, clean up a bit. I'm planning to write tests after code review

next PRs could be 
- improve DX with pause on exception
- support `react-error-overlay`

Andarist:
Just to let u know - I'm working on making this repo a monorepo managed by lerna. When it's done I plan to import your babel plugin repo here (ideally with git history untouched), then make some minor changes and publish.

AriaMinaei:
Not sure if you guys have considered it, but capturing a proper stack trace with the correct line numbers along with the jumps between the effects can probably be done without a babel plugin. Promise libraries do it that way at least. Here is [when](https://github.com/cujojs/when/blob/8a7785a3f8bc405f82b429434b6dbdab418f745c/monitor/PromiseMonitor.js)'s method for example.

Andarist:
@AriaMinaei thanks for sharing, this is an interesting technique but imho it seems more old-fashioned and heavy than a "simple" babel plugin. Note that we will show "nicer" (than today) errors even without babel plugin.

Slapbox:
What's the best way to get started with this now? Redux Saga is the best thing sliced bread, but the error tracing is no fun.

Andarist:
@Slapbox We are working on merging this PR - https://github.com/redux-saga/redux-saga/pull/1374 and when we do that (should be soon) I'm going to publish a new beta release with it.

PR #1322, opened 2018-02-02
fix export in introcution docs

### LINKED ISSUE #1321: helloSaga() is not defined in introduction documentation step
Hi everyone.

I was following the steps of the introduction documentation and i have found a example that doesnt work.

The documentation shows a refactor that looks like this:

``` javascript
import { put, takeEvery, all } from 'redux-saga/effects'


function* incrementAsync() {
  yield delay(1000)
}


function* watchIncrementAsync() {
  yield takeEvery('INCREMENT_ASYNC', incrementAsync)
}


// notice how we now only export the rootSaga
// single entry point to start all Sagas at once
  yield all([
    helloSaga(),
    watchIncrementAsync()
  ])
}
```

Notice how the helloSaga function is not defined.

MarcoNicolodi:
Hello. 

This is my first pull request it should fix #1321

### DIFF
--- a/docs/introduction/BeginnerTutorial.md
+++ b/docs/introduction/BeginnerTutorial.md
@@ -170,23 +170,24 @@ Now we have 2 Sagas, and we need to start them both at once. To do that, we'll a
 import { delay } from 'redux-saga'
 import { put, takeEvery, all } from 'redux-saga/effects'
 
+function* helloSaga() {
+  console.log('Hello Sagas!')
+}
 
 function* incrementAsync() {
   yield delay(1000)
   yield put({ type: 'INCREMENT' })
 }
 
-
 function* watchIncrementAsync() {
   yield takeEvery('INCREMENT_ASYNC', incrementAsync)
 }
-
 
 // notice how we now only export the rootSaga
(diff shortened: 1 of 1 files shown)


Andarist:
More correct fix is to include `helloSaga` in this snippet - `incrementSaga` is supposed to be wrapped by `watchIncrementAsync` and not ran by its own.

Andarist:
Thanks for the contribution!

MarcoNicolodi:
I think this PR should be reopened. As I followed the documentation, i started the testing section, and it tests the `incrementAsync` saga. Therefore, `incrementAsync` should be also exported. Should i add and `export` to all functions of the saga so they can be imported individually for testing purposes? I think if we dont, people would be misguided that they should erase the individual exports, making the test in the documentation to fail. 

Sorry i didnt notice it early.

Andarist:
Thanks for noticing. IMHO export shouldnt be added here. The point of this tutorial is to add piece by piece and at the point of this snippet there is no need to add export here.

A note about the need to add it when it's imported in the snippet above would be nice though.

PR #1331, opened 2018-02-14
Move redux to peer dependency

JonathanMontane:
This PR removes the somewhat unnecessary dependency on `redux` of this library.

`redux-saga` only uses the `compose` method from `redux`, which can be reimplemented in 10 lines of code.

I moved the dependency on `redux` to peerDependencies, because this library was designed as a plugin to redux, and this is the [recommended place](https://nodejs.org/en/blog/npm/peer-dependencies/) to declare such relationships.

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "url": "https://github.com/redux-saga/redux-saga/issues"
   },
   "homepage": "https://redux-saga.js.org/",
-  "dependencies": {
+  "peerDependencies": {
     "redux": "^3.5.1"
   },
   "devDependencies": {
@@ -105,6 +105,7 @@
     "react": "^15.0.0",
     "react-dom": "^15.0.0",
     "react-redux": "^4.4.5",
+    "redux": "^3.5.1",
     "rimraf": "^2.4.3",
     "rollup": "^0.50.0",
     "rollup-plugin-babel": "4.0.0-beta.1",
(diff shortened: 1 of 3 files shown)


Andarist:
`redux` is not a peerDep of this library, while it's primary target audience are in fact redux apps it can successfully be used in other setups too (i know numerous people using it outside of redux) thanks to `runSaga` API. Therefore making it a peer is not really true and might discourage people from experimenting with other setups. It might be optionalDependency, but I was never sure how to use them properly.

Also what this duplication of `compose` gives us? If used without `redux` tree-shakers just should remove unused parts of `redux` and leave us just with `compose`, if used with `redux` (which is the most common setup) it won't get duplicated.

JonathanMontane:
I don't think that everyone uses tree-shakers, although people probably should, so this may have a negative impact on some users.

This duplication of `compose` also guarantees that the maintenance of this project is not impacted by changes in the redux API or publication issues (I'm thinking of the left-pad incident for instance).

I agree that the peerDep is probably too much and you could probably remove the dependency altogether.

Andarist:
> I don't think that everyone uses tree-shakers, although people probably should, so this may have a negative impact on some users.

But it benefits (I realize it's not much, just few bytes, but still) majority of users.

> This duplication of compose also guarantees that the maintenance of this project is not impacted by changes in the redux API or publication issues (I'm thinking of the left-pad incident for instance).

left-pad incidents don't happen too often. All software depends on other software one way or another and imho reinventing the wheel somewhat contradicts OSS goals.

Ofc if `redux` would change anything regarding `compose` I'd ditch it right away and just inline their current implementation into the `redux-saga`, but I trust them not to break semver so having dep on it should be safe.

To sum it up - I'm not entirely sure what best solution for this is, I'm just not convinced that dropping this dep solves anything and having just yet another compose util just result in increased bundle size of redux-saga, maintenance burden etc.

Issue #1315, opened 2018-01-23
Helper ignoring coming action when previous action is still running


tomoyukim:
Hi, I locally created a helper named `takeOnce` like below in order to ignore coming action when the previous action is still running.

```javascript
function* takeOnce(pattern, saga, ...args) {
  const lastTask;
  while (true) {
    const action = yield take(pattern);
    if (_.isNil(lastTask) || !lastTask.isRunning()) {
      lastTask = yield fork(saga, ...args.concat(action));
    }
  }
}
```

I'm happy if it's supported by the library itself as authorized helper and if this is helpful, I'd like to create a new PR. So, what do you think about it?

Andarist:
`takeOnce` would mean for me "take once and exit", so I'd appreciate a better (more descriptive) name. It seems like you have a kinda `leadingDebounce` mechanism which takes only a leading action in a sequence until it finishes a task - `debounce` most often is parametrized with some time parameter, so it's probably not a best name for this though.

It can also be implemented a little bit easier (without introspecting `task`):
```js
function* GIVE_ME_NAME(pattern, saga, ...args) {
	cosnt chan = actionChannel(pattern, buffers.dropping(1))
	while (true) {
		const action = yield take(chan)
		yield call(saga, ...args, action)
		yield flush(chan)
	}
}
```

If we can come up with a better name I would be glad to merge this addition. Keep in mind though that internally helpers (takeEvery, takeLatest, throttle) are not implemented with generators syntax, you'd have to implement this with internal `fsmIterator` - if you need any help with understanding how it works I can explain it.

tomoyukim:
Thank you for your positive and kind feedback.

> If we can come up with a better name I would be glad to merge this addition.

Another idea comes to mind is `takeExclusive` or `takeExclusively` because I think it looks like somehow mutual exclusion. What about this?

I've tried to write code with `fsmIterator` referring another `take*` helpers based on my above snippet but I feel your implementation is better than me. So, I'll try to rewrite it with your idea.

BTW, I guess I need to include testing and maybe document when I create PR. It seems contribution guid is under discussion according to https://github.com/redux-saga/redux-saga/issues/1287. So, would you let me know what I should do for the PR?

Andarist:
Let's start with providing a `fsmIterator`-based implementation and tests. Docs would ofc be great too - adding it to the API docs with an example usage would be superb.

tomoyukim:
OK. Thank you for your support. I'll prepare for the PR.

[...]

Issue #589, opened 2016-10-14
Is there a better way to do takeFirst?


ekeric13:
So the takeEvery and takeLatest helper functions are quite useful.
But in a situation where you want to take the first take, but ignore the rest, such as a user clicking the delete button multiple times, what is the best way to write your saga watcher function so that it only takes the first delete action dispatch?

By modifying takeLatest I got a function that looks like this:
```
function* takeFirst(pattern, saga, ...args) {
  const task = yield fork(function* () {
    let firstTask = true;
    while(true) {
      const action = yield take(pattern);
      if (firstTask) {
        firstTask = false;
        yield call(saga, ...args.concat(action));
        firstTask = true;
      }
    }
  });
  return task;
}
```

Now in a watcher you might be able to do this:
```
export function* watchDeleteWidgetsSaga() {
  yield takeFirst(DELETE_WIDGETS_SAGA, deleteAndUpdateWidgets);
}
```

Is this a good way to "takeFirst"? Or are there already saga helpers and effect creators in the existing saga api that would help me do this? Or is "takeFirst" just an anti-pattern completely - if so why?

Andarist:
call is a blocking effect so you can just skip this whole firstTask flag.
In ur fork just put take and call in a while (true) loop. Effect will be
the same and what u r looking for

thebucknerlife:
@Andarist any advice on how to test a custom helper like takeFirst? We've followed the same pattern as above. I've been trying to pattern our test off of https://github.com/yelouafi/redux-saga/blob/master/test/sagaHelpers/takeEvery.js.

Andarist:
I would rather base this one on https://github.com/yelouafi/redux-saga/blob/master/test/sagaHelpers/takeLatest.js , rather than on `takeEvery`. You can dispatch your action, multiple times (so only first will get handled), then resolve promise in the `call`ed worker so it gets unblocked. Push to the `actual` some time on the way and just check in the end that it contains only one action

nibblesnbits:
@ekeric13, do you have a final version of this function?  I imagine it wouldn't be too bad to do myself, but if you've got a tested version, I'd love to see it.

ekeric13:
@nibblesnbits 
The code @Andarist has above is what you are looking for
```
function* takeFirst(pattern, saga, ...args) {
  const task = yield fork(function* () {
    while(true) {
      const action = yield take(pattern);
      yield call(saga, ...args.concat(action));
    }
  });
  return task;
}
```

[...]

PR #1332, opened 2018-02-16
Adds takeLeading as new helper

tomoyukim:
Hi all,

I tried to add new helper which takes the first and ignore the rest.

For discussion in the background, see https://github.com/redux-saga/redux-saga/issues/1315 and https://github.com/redux-saga/redux-saga/issues/589


Andarist:
Looks really good 👍 need to spend more time with it later to read all the docs and stuff

docs/api/README.md
```diff
@@ -263,6 +265,57 @@ const takeLatest = (patternOrChannel, saga, ...args) => fork(function*() {
 
 You can also pass in a channel as argument and the behaviour is the same as [takeLatest(pattern, saga, ...args)](#takelatestpattern-saga-args).
 
+### `takeLeading(pattern, saga, ...args)`
+
+Spawns a `saga` on each action dispatched to the Store that matches `pattern`. And automatically ignores
```
Andarist: > And automatically ignores any coming new `saga` task if it's still running.

This is imho slightly inaccurate. It would be better to just say that it blocks until spawned saga completes AND then starts to listen for a pattern again.

docs/api/README.md
```diff
@@ -263,6 +265,57 @@ const takeLatest = (patternOrChannel, saga, ...args) => fork(function*() {
...
+any coming new `saga` task if it's still running.
+
+Each time an action is dispatched to the store. And if this action matches `pattern`, `takeLeading`
```
Andarist: This paragraph described the behaviour again in slightly inaccurate way. Would be easier to say that it's listening for the actions when it doesn't run a saga.

docs/api/README.md
```diff
@@ -263,6 +265,57 @@ const takeLatest = (patternOrChannel, saga, ...args) => fork(function*() {
...
+
+#### Example
+
+In the following example, we create a simple task `fetchUser`. We use `takeLeading` to
+start a new `fetchUser` task on each dispatched `USER_REQUESTED` action. Since `takeLeading`
+ignores any new coming task after it's started, we ensure that if a user triggers multiple consecutive
+`USER_REQUESTED` actions rapidly, we'll only keep on running with the leading action
+
+```javascript
+import { takeLatest } from `redux-saga/effects`
```
Andarist: this should be `takeLeading`

test/sagaHelpers/takeLeading.js
```diff
@@ -0,0 +1,61 @@
+/* eslint-disable no-unused-vars, no-constant-condition */
```
Andarist: personally i dont like disabling rules at the top of the file, its hard to determine why they were disabled

would prefer to mute rules at call sites, if possible

tomoyukim:
@Andarist Thank you for your review and feedbacks. I've tried to improve them. Would you check them at your convenience?

Andarist:
Thanks ❤️

AvalonCV:
Hi there,
thanks for takeLeading - this is just what I was looking for. My IDE is still a bit uncertain about it, so I want to ask if it would be possible to add takeLeading to the effects typescript definitions as well? I assume this could be a simple one-liner (just as takeLatest is).
Best regards

Andarist:
This helper is not released yet - it will be soon-ish, but as part of v1-beta line (which should be safe to use, breaking changes are minor & the library is well-tested). TS typings are community driven, so I encourage you to prepare a PR for this :)

PR #1342, opened 2018-02-24
Update BeginnerTutorial.md

rmacy:
Later in the tutorial, we tell the user how to make tests for `incrementAsync` but we do not export the function in `sagas.js`, so users have to go back and add export after seeing errors in console.


Andarist:
Thanks!

PR #1365, opened 2018-03-15
Fix babel plugin tests

Andarist:


.eslintignore
```diff
@@ -1 +1 @@
-**/test/**/expected.js
```
Andarist: actually i have included `actual.js` (or how those files are named at the moment) on purpose, i know we sometimes might want to write code that breaks our lint rules, but we can always use eslint-ignore comments for that

 WDYT?

Andarist:
Why does `{{filename}}` work? is it a built-in feature of tape?

.eslintignore
restrry: imho `fixtures` could be excluded completely
1 we don't care about styles there, those are for machine, not for people
2 I'd prefer don't touch them if we decide to change code style once

Andarist:
Oh, it was already in the code and not the addition in the PR - thanks for clarifying.

The nice thing about expected files is that they might be auto generated on the initial run, with that template in we lose that trait. Wondering if we could preserve it by not using this template - I think it's doable by replacing `transformFileSync` in test runner with `transformSync` and passing a `filename` option to it. That way we could just trim the filename to the project's root to avoid conflicts between machines.

.eslintignore
Andarist: I don't have a strong opinion on this, but I don't believe that "those are for machine" is a right statement here.

They are our tests and they need be reviewable by humans for them to have any purpose.

packages/babel-plugin-redux-saga/README.md
```diff
@@ -46,23 +46,14 @@ saga2[_SAGA_LOCATION] = {
...
-
 ## Usage
 
 1. with babel
 ```js
 babel.transform(content, {
     sourceMaps: true,
     plugins: [
-        'babel-plugin-transform-redux-saga-source', { /* options */ }
+        'babel-plugin-redux-saga', { /* options */ }
```
Andarist: we have a typo here, should be `['babel-plugin-redux-saga', { /* options */ }]` or mention of the options should get removed from here

packages/babel-plugin-redux-saga/test/index.js
```diff
@@ -96,7 +87,14 @@ utilTests.forEach(function(config) {
     var sourcePath = path.join(__dirname, 'fixtures', config.fixture, 'source.js')
     var sourceMapPath = path.join(__dirname, 'fixtures', config.fixture, 'source.js.map')
 
-    var actual = processFile(sourcePath, sourceMapPath, config.options || {}, config.pluginOptions || {})
+    var inputSourceMap = fs.existsSync(sourceMapPath)
+      ? JSON.parse(fs.readFileSync(sourceMapPath).toString())
+      : undefined
+
+    var sourceCode = fs.readFileSync(sourcePath).toString()
+    const testCaseName = path.join(config.fixture, 'source.js')
```
Andarist: u are mixing `const` and `var` here, minor nit, dont have to be changed right away

restrry: good catch. I wanted to support old node engines so avoided using es6+ syntax (redux-saga package.json doesn't contain any requirements for node version)

PR #1372, opened 2018-03-22
fix docs typo

Revinand:


Revinand:
Thank you for the great library!

PR #1374, opened 2018-03-23
Test babel plugin for babel@6, babel@7

restrry:
### Description
- redux-saga tries to use global symbol for location
- babel plugin imports `globalSymbolNames` (from redux-saga) to get name for `location` global symbol. probably `globalSymbolNames` shouldn't be part of public api of redux-saga. 
- babel plugin has been cleaned from babel@7 API. now it can be used for both babel@6, babel@7
- tests are re-written accordingly + switched to jest runner


packages/babel-plugin-redux-saga/package.json
```diff
@@ -9,17 +9,20 @@
   ],
   "main": "./src/index",
   "scripts": {
-    "test": "node ./test/index.js | tap-difflet"
+    "test": "jest"
```
Andarist: 👍

packages/babel-plugin-redux-saga/package.json
```diff
@@ -9,17 +9,20 @@
...
   },
   "devDependencies": {
     "@babel/core": "7.0.0-beta.39",
     "@babel/preset-env": "7.0.0-beta.39",
-    "tap-difflet": "^0.7.0",
-    "tape": "^4.8.0",
+    "babel-core": "^6.26.0",
```
Andarist: huh, funny that we do not have to fallback to more custom solution because babel got migrated to scoped packages

packages/core/src/index.js
```diff
@@ -3,7 +3,7 @@ export default middleware
 
 export { runSaga } from './internal/runSaga'
 export { END, eventChannel, channel, multicastChannel, stdChannel } from './internal/channel'
-export { CANCEL, SAGA_LOCATION } from './internal/symbols'
+export { CANCEL, SAGA_LOCATION, globalSymbolNames } from './internal/symbols'
```
Andarist: we do not import/require it from the babel plugin (and probably we won't, using global symbol registry is "good enough"), so probably better to just hide it in the internals, probably not even worth to keep it as a map as we have only a single item there

I'd just go with same, duplicated, implementation for SAGA_LOCATION in both core and babel plugin (and wouldnt also export SAGA_LOCATION from here, until somebody requests it)

packages/babel-plugin-redux-saga/src/index.js
```diff
@@ -1,31 +1,36 @@
...
 
 module.exports = function(babel) {
   var { types: t } = babel
   var sourceMap = null
-  var symbolNameBinding = null
 
-  function getAssignementWithSymbolProperty(name, symbolName) {
-    return t.memberExpression(t.identifier(name), t.identifier(symbolName), true)
+  function getAssignementWithSymbolProperty(name) {
+    return t.memberExpression(
```
Andarist: I have lately really enjoyed using `@babel/template`, you can access it from the `babel` object here too, without need to import it yourself):
```j
var { types: t, template } = babel
```

We can only use `template` (no `template.ast`, `template.expression`, etc) to keep babel@6 compatibility.

IMHO it improves the readability quite a lot.

packages/babel-plugin-redux-saga/src/index.js
```diff
@@ -1,31 +1,36 @@
...
+      ),
+      true);
   }
 
-  function getObjectExtensionNode(path, functionName) {
-    if (!symbolNameBinding) {
-      const { name: locationSymbolName } = importHelper.addNamed(path, locationSymbolExportName, locationSymbolPath, {
-        nameHint: hintedLocationSymbolExportName,
-      })
+  function getAssignementWithPlainProperty(name) {
```
Andarist: symbols are not enumerable, so maybe we should wrap this fallback output with `Object.defineProperty`?

packages/babel-plugin-redux-saga/src/index.js
```diff
@@ -142,8 +146,8 @@ module.exports = function(babel) {
     CallExpression(path, state) {
```
Andarist: shouldnt we just visit `yield`s? probably would make reasoning about this easier

Andarist: not sure on which line I should comment about this, but it should be easy to replace those IIFEs with comma separated expressions
```js
var _ref;
yield _ref = call(smthelse), _ref[_SAGA_LOCATION] = { fileName: ..., lineNumber: ... } ,_ref;
```
Can be achieved with things like `scope.generateDeclararedUidIdentifier('ref')` and `path.replaceWithMultiple` - or smth like that 😉

restrry:
@Andarist  addressed + refactored

packages/babel-plugin-redux-saga/src/index.js
```diff
@@ -112,18 +76,21 @@ module.exports = function(babel) {
...
       if (path.node.generator !== true) return
 
-      const functionName = path.node.id.name
-      const objectExtensionNode = getObjectExtensionNode(path, functionName)
-
+      var functionName = path.node.id.name
       var locationData = calcLocation(path.node.loc, state.file.opts.filename, state.opts.basePath)
 
-      var declaration = assignLoc(objectExtensionNode, locationData.fileName, locationData.lineNumber)
+      const extendedDeclaration = extendDeclarationWithLocation({
```
Andarist: technically this is also possible:
```js
const saga = function* saga() {}
```

and won't be visited with the plugin right now.

packages/babel-plugin-redux-saga/src/index.js
Andarist: @restrry WDYT?

packages/babel-plugin-redux-saga/src/index.js
```diff
@@ -1,75 +1,40 @@
...
-          ]),
-        ),
-      ),
-      t.returnStatement(t.identifier(tempVarId)),
-    ]
-    const container = t.functionExpression(t.identifier(fnId), [], t.blockStatement(body))
-    return t.callExpression(container, [])
+  const extendExpressionWithLocation = template(`
+    (function reduxSagaSource() {
+      return Object.defineProperty(TARGET, SYMBOL_NAME, {
```
Andarist: When using real symbols we don't have to use `Object.defineProperty`, but using it consistently makes the code easier, so let's leave it that way for now.

packages/babel-plugin-redux-saga/src/index.js
```diff
@@ -142,23 +109,26 @@ module.exports = function(babel) {
...
+      var isParentYield = path.parentPath.isYieldExpression()
+      var isGrandParentYield = path.parentPath.parentPath.isYieldExpression() // NOTE: we don't check whether parent is logical / binary / ... expression
       if (!isParentYield && !isGrandParentYield) return
 
       if (!node.loc) return
       // if (path.parentPath.node.delegate) return // should we ignore delegated?
 
       var file = state.file
       var locationData = calcLocation(node.loc, file.opts.filename, state.opts.basePath)
+      var sourceCode = path.getSource();
```
Andarist: with babel@7 u can use more accurate `toString`:
```js
var { hasOwnProperty: hasOwn } = {}
var sourceCode = hasOwn.call(path, 'toString') ? path.toString() : path.getSource();
```

packages/babel-plugin-redux-saga/test/__snapshots__/runner.test.js.snap
```diff
@@ -0,0 +1,976 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`babel v6 attach source to declaration 1`] = `
```
Andarist: hm, those snapshots are imho less readable. Would it be possible to go back to separate files and suffix expected files with babel's version?

packages/babel-plugin-redux-saga/src/index.js
restrry: make sense, but imho wrapping with `Object.defineProperty` makes code a bit more readable (for debugging, for example), so I'd stick to IIFE.

packages/babel-plugin-redux-saga/test/__snapshots__/runner.test.js.snap
restrry: if it's less readable for other maintainers, I'll change the structure 👍

packages/babel-plugin-redux-saga/src/index.js
restrry: yes, that was done intentionally to unify approach

packages/babel-plugin-redux-saga/test/__snapshots__/runner.test.js.snap
Andarist: I'd appreciate the change, imho separate files are easier to browse & inspect + syntax highlighting works in editors

restrry: done

packages/babel-plugin-redux-saga/src/index.js
restrry: I forgot that wanted to add FunctionExpression. Done.
from `error-demo` example
before:


after:

packages/babel-plugin-redux-saga/src/index.js
Andarist: what do u think about this one?

> shouldnt we just visit yields? probably would make reasoning about this easier

@restrry

packages/babel-plugin-redux-saga/test/fixtures/effect-object-props/babel7-expected.js
```diff
@@ -0,0 +1,20 @@
+function* withEffectObjectProps() {
+  yield function reduxSagaSource() {
+    return Object.defineProperty(race({
+      timeout: delay(3000),
+      cannelled: take('CANCELLED')
```
Andarist: nit: typo `cannelled` -> `cancelled`

packages/babel-plugin-redux-saga/src/index.js
restrry: I did it in earlier implementation, but found several places in our codes base when people used expressions, like
```js
function*() {
  yield foo.bar(1, 2, 3) || {}
}
// or
function* () {
  yield (foo? bar() : baz())
}
```
we can try to get rid of expression support, that should simplify code

Andarist: well, idiomatic way of working with sagas is generally yielding effects only which are just objects - so generally it should be quite "safe" to wrap even more complex expressions, as their completion values should be objects, right?

restrry: actually yes, if we're using Symbol. but in case of using fallback to string key is not so 'safe'. probably, that's a minor concern and we can return to the initial implementation and wrap only yielded values

Andarist: string key will still be used with `defineProperty`, what concern do u see about this in comparison with Symbol situations? Seems to me that both of them are the same regarding "safety" of this

packages/babel-plugin-redux-saga/src/index.js
```diff
@@ -5,38 +5,58 @@ var globalSymbolName = '@@redux-saga/LOCATION'
...
+   *  Object.defineProperty(TARGET, SYMBOL_NAME, {
+   *    value: {
+   *      fileName: FILENAME,
+   *      lineNumber: LINE_NUMBER,
+   *      sourceCode?: SOURCE_CODE
+   *    }
+   *  });
+   */
+  function createLocationExtender(node, useSymbol, fileName, lineNumber, sourceCode){
+    return t.callExpression(
```
restrry: unfortunately, using template mechanism leads to cryptic error if we process yieldExpression and use `regenerator` package at the same time
`babel Expected type "Expression" with option {}`

packages/babel-plugin-redux-saga/src/index.js
restrry: totally agree, changed

packages/babel-plugin-redux-saga/src/index.js
Andarist: Could u explain how I could reproduce it so I could debug later? Exact input and options used (I guess we are talking about most recent version of this plugin). I have both today's and tomorrow's evenings already planned but I can work on it on Saturday and hopefully release it then

restrry: just redefine `createLocationExtender` body to
```js
  function createLocationExtender(node, useSymbol, fileName, lineNumber, sourceCode){
    const extendExpressionWithLocation = template(`
      Object.defineProperty(TARGET, SYMBOL_NAME, {
          value: {
            fileName: FILENAME,
            lineNumber: LINE_NUMBER
          }
        });
    `);
    const extendedDeclaration = extendExpressionWithLocation({
        TARGET: node,
        SYMBOL_NAME: getSymbol(useSymbol),
        FILENAME: t.stringLiteral(fileName),
        LINE_NUMBER: t.numericLiteral(lineNumber),
      })

    return extendedDeclaration;
}
```
and run any existing test containing 
```js
    options: { presets: ['env'] },
```

Andarist: You should unpack `expression` from returned value, so:
```js
return extendExpressionWithLocation.expression;
```

Babel don't know how did you want to parse this code and it assumes statement here - in babel@7 there are actually various versions of `template`, ie. `template.expression`, `template.statement` and others - but we cant use them if we want to keep babel@6 compat for now.

Also It would be better for perf to cache `extendExpressionWithLocationTemplate` in outer scope - just before `createLocationExtender` declaration.

Andarist: I've pushed out the fix for this

restrry: great explanation, I wish I could find that information before 👍

restrry:
🚢

PR #1377, opened 2018-03-26
examples error: counter/src/sagas/index.js

ylksty:
16:17-22 "export delay was not found in redux-saga/effects


Andarist:
Have u tried running a bootstrap command? With new version (not yet released) it is an effect - https://github.com/redux-saga/redux-saga/blob/7827597d3855edbed1e5e2dc7a219e9dbe859106/packages/core/src/effects.js#L25

titulus:
related to #1490

PR #1388, opened 2018-04-09
babel-plugin: build relative path by default

restrry:
old behavior - plugin generates absolute path, to get relative path we should pass `basePath` option
new behavior - plugin generates relative path (to cwd), to get absolute path we should pass `useAbsolutePath: true` option


Andarist:
are those relative paths clickable in most scenarios?

restrry:
not yet, requires additional investigation.
to open them in editor we need server (or middleware)
to open in devtool `sources` tab we need to map file path from message to file path from `sources` tab. that's bundler dependent part. for example, webpack, builds file path in the next way `webpack://{filename}`, also that part is configurable on bundler level, so we can't predict exact prefix/schema.
anyway I'm continue working on that investigation, probably we can do that later

Issue #1395, opened 2018-04-11
Call() Silently Kills All Sagas for an Uncaught Promise Rejection


max-winderbaum:
When we pass a promise into `call()` that gets rejected asynchronously, then the `onError` function passed into `createSagaMiddleware` does not ever fire and all sagas stop working.

If the `call()` is wrapped in a `try/catch` block then it doesn't kill all the sagas, but we'd still love it if all errors within sagas get passed into the `onError` function!

Repro: https://runkit.com/max-winderbaum/5ace4c686e1db2001293b4b4

Please let me know if there's anything else I can do to help this along! I'm also looking into how to create a PR to fix this, but I'd love some help and/or a point in the right direction.

Andarist:
This is most likely caused by rejecting with `undefined` instead of an actual error. Is there any reason why this happens in your app? It would be good to guard against that in redux-saga though - PRs are welcome.

PR #1398, opened 2018-04-12
Safely Handle Various Arguments in Calls to Promise.reject()

max-winderbaum:
Resolves redux-saga/redux-saga#1395

I found the issue with redux-saga/redux-saga#1395, and there are several layers:

1. Rejecting a promise with undefined does actually crash sagas. This is a fairly easy fix.

2. A promise with an unhandled reject() that has an argument other than an Error() object causes that rejection to skip the onError method and just write to the console. This one is a bit more hairy, because the failure is silent in that we can't log it on our end.

The simplest possible solution I can think of is to convert the first argument passed into reject() into an Error() object for promises only... this is a bit dishonest because it's not technically what got passed into reject() but it's not so bad because we're already handling reject() like it's an error anyways

@Andarist Thoughts?

### DIFF
--- a/packages/core/src/internal/proc.js
+++ b/packages/core/src/internal/proc.js
@@ -186,7 +186,7 @@ export default function proc(
 
   const logError = err => {
     log('error', err)
-    if (err.sagaStack) {
+    if (err && err.sagaStack) {
       log('error', err.sagaStack)
     }
   }
@@ -338,11 +338,11 @@ export default function proc(
       })
 
       if (!task.cont) {
-        if (result.sagaStack) {
+        if (result && result.sagaStack) {
           result.sagaStack = sagaStackToString(result.sagaStack)
         }
 
-        if (result instanceof Error && onError) {
+        if (onError) {
           onError(result)
         } else {
           // TODO: could we skip this when _deferredEnd is attached?


packages/core/src/internal/proc.js
```diff
@@ -473,7 +473,14 @@ export default function proc(
     } else if (is.func(promise.abort)) {
       cb.cancel = () => promise.abort()
     }
-    promise.then(cb, error => cb(error, true))
+    promise.then(cb, error => {
+      if (error instanceof Error) {
```
Andarist: I dont think we should actually convert those not-errors to errors. Let them flow through the system and guard against them at later point - before attaching properties to them and before logging them

packages/core/src/internal/proc.js
```diff
@@ -338,15 +338,20 @@ export default function proc(
...
       if (!task.cont) {
-        if (result.sagaStack) {
+        if (result && result.sagaStack) {
           result.sagaStack = sagaStackToString(result.sagaStack)
         }
 
         if (result instanceof Error && onError) {
           onError(result)
+        } else if (onError) {
+          const errorString = result && result.toString() ?
```
restrry: if user throws null or undefined, why we can't just pass it to `onError` as is?

max-winderbaum: Sending an error is important to not introduce a breaking change - we have no idea how many people in the world are expecting an Error object to be passed into `onError`, and I didn't want to break their code.

max-winderbaum: This also has the added benefit of sending up a stack trace so someone could attempt to debug what happened during the error based on that

Andarist:
@max-winderbaum dont worry about CI, its a separate problem

packages/core/src/internal/proc.js
Andarist: You are targeting at the moment v1-beta line (and I really wouldnt like to do any development on 0.x line), breaking changes can happen during a beta period and this also shouldnt break too many ppl, so it's fine to just pass what was thrown "as is" to `onError`

max-winderbaum:
@Andarist Ok, this should do it!

Andarist:
Thanks!

PR #1401, opened 2018-04-15
update docs for delay effect

restrry:
today I updated my project to `redux-saga#master` and encountered that delay is effect now, but docs still point to the old path

related changes https://github.com/redux-saga/redux-saga/commit/eef6f89b4c59356ff99bbeee33cfd615b38039aa


Andarist:
Downside of this change is that we have 2 delays now :s might be confusing for newcomers, but 🤷‍♂️

PR #1402, opened 2018-04-15
Fixed location of the delay function in the beginner tutorial

IvanJov:
Hi! I found one small issue in the beginner tutorial, this PR fixes it. Thanks!

### DIFF
--- a/docs/introduction/BeginnerTutorial.md
+++ b/docs/introduction/BeginnerTutorial.md
@@ -133,7 +133,7 @@ Now we will introduce another Saga to perform the asynchronous call. Our use cas
 Add the following code to the `sagas.js` module:
 
 ```javascript
-import { delay } from 'redux-saga/utils'
+import { delay } from 'redux-saga'
 import { put, takeEvery } from 'redux-saga/effects'
 
 // ...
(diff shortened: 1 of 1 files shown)


Andarist:
Actually this is correct for v1-beta .

IvanJov:
@Andarist Hmm, okay. So on this line https://github.com/redux-saga/redux-saga/pull/1402/files#diff-91df1104cefdb22cf3dc49af49294dd9L282, there was already an import directly from `redux-saga`. Should I change it everywhere to `redux-saga/utils`?

restrry:
as @Andarist [noticed](https://github.com/redux-saga/redux-saga/pull/1401#issuecomment-381434597) we have 2 different `delay` functions across code base & documentations. 
imho, `delay` from effects is part of api, `delay` from utils is internal thing. So probably that tutorial should be changed to use another function that returns a promise, since the whole tutorial explains how redux-saga can handle both effects and promises

IvanJov:
@restrry Okay, that makes sense. I am not sure I understand what do you mean with `that tutorial should be changed to use another function that returns a promise`?

restrry:
@IvanJov for example, instead of importing delay from utils, we can define function
```
const delay = (ms) => new Promise (res => setTimeout(res, ms))
```
now referring to it, in context of the tutorial, makes sense.

Andarist:
good idea @restrry , that way we wouldnt confuse ppl about having 2 `delay` functions

IvanJov:
Gotcha, let me change that!

IvanJov:
@Andarist @restrry Please check it again!

restrry:
👍  from me

Andarist:
Thanks!

PR #1414, opened 2018-04-23
[types] Add `race([...effects])` definition

ggarek:


restrry:
have you rebased against master?

Andarist:
Let's see if master passes now. Thanks for the contribution!

PR #1419, opened 2018-04-25
Fix range of supported redux versions. Fix #1306

### LINKED ISSUE #1306: update to redux 4
redux 4 is coming! https://github.com/reactjs/redux/releases/tag/v4.0.0-beta.1

it seems not to introduce many breaking changes, so it should work with saga as well, but saga has currently a dependency on redux 3.x, so that at least needs to be updated

fjorgemota:
Well, hope that everything is OK, but I'm not really sure that only these changes suffices.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   },
   "homepage": "https://redux-saga.js.org/",
   "dependencies": {
-    "redux": "^3.5.1"
+    "redux": ">=0.10 <5"
   },
   "devDependencies": {
     "@babel/cli": "7.0.0-beta.39",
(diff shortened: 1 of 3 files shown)


Andarist:
Thanks!

package.json
```diff
@@ -39,7 +39,7 @@
   },
   "homepage": "https://redux-saga.js.org/",
   "dependencies": {
-    "redux": "^3.5.1"
+    "redux": ">=0.10 <5"
```
restrry: is it compatible with version < `3.5.1`? I don't know why it was selected before

Andarist: Previously I've just installed redux and it got me latest version at that time in the range. Later I've backtracked in which version `compose` got introduced to redux and it was 0.10

fjorgemota: Curiously, on the tests I've made the installed redux version was 3.5.2, because react and some other peer dependencies seem to have not been updated to support redux 4.0 yet (like redux-devtools, for example).

But I still think that external users should not suffer from that problem and may use redux 4.0 easily with this change.

PR #1423, opened 2018-04-25
Update plugin readme + fix examples

restrry:
- update readme
- examples use latest `saga` related packages
- fix real-world, cancellable-counter examples

### DIFF
--- a/packages/babel-plugin-redux-saga/README.md
+++ b/packages/babel-plugin-redux-saga/README.md
@@ -1,14 +1,17 @@
-## NOTE: plugin is still in development mode, so use it on your own risk
+## NOTE: plugin is still in beta, so use it on your own risk
 
 Babel plugin for code instrumenting by extending `redux-saga` code fragments with additional meta-data. Meta-data contains information about code fragment location and other details, that could be consumed by developer tools or libraries.
+Adding the plugin improve logging for errors thrown in your sagas.
+Example of setup and demo are available [here](../../examples/error-demo)
 
 ## Example
 
 Source:
 
 ```js
+// src/sagas/index.js
 function* saga1(){
-    yield foo(1, 2, 3);
+    yield call(foo, 1, 2, 3);
 }
 
 function* saga2(){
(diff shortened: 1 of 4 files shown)


examples/async/package.json
```diff
@@ -7,7 +7,7 @@
     "test": "cross-env NODE_ENV=test babel-node test/sagas.js | tap-spec"
   },
   "dependencies": {
-    "redux-saga": "^1.0.0-beta.0"
+    "redux-saga": "file:../../packages/core"
```
Andarist: actually I would prefer not relaying on those packages being in a monorepo and stuff, with "real" dependencies we can in example easily upload those examples to things like codesandbox.io etc and it should download its copies of those deps

examples/real-world/.babelrc.js
```diff
@@ -9,7 +9,6 @@ module.exports = {
       '@babel/env',
       {
         loose: true,
-        modules: false,
         forceAllTransforms: true,
```
Andarist: why the change here? this demo uses webpack and it's better not to transpile modules when using it

packages/babel-plugin-redux-saga/README.md
```diff
@@ -19,44 +21,43 @@ function* saga2(){
...
-        res[_SAGA_LOCATION] = {
-            fileName: "{{filename}}",
-            lineNumber: 2,
-            code: "foo(1, 2, 3)"
-        };
-        return res;
-    }();
+    yield Object.defineProperty(call(foo, 1, 2, 3), Symbol.for("@@redux-saga/LOCATION"), {
+        value: {
+            fileName: "{{pathToFile}}",
```
Andarist: in readme it would be more clear to use some dummy path here like, `../src/sagas/index.js`

examples/async/package.json
restrry: ok. I'll revert it

examples/real-world/.babelrc.js
restrry: `server.js` is not built with webpack. second option is use CJS for server.js

packages/babel-plugin-redux-saga/README.md
restrry: ok

packages/babel-plugin-redux-saga/README.md
```diff
@@ -19,44 +22,43 @@ function* saga2(){
...
+})
 ```
 
 ## Usage
 
 1. with babel
 ```js
-babel.transform(content, {
-    sourceMaps: true,
+// .babelrc
```
Andarist: imho would be good to use a valid JSON here

Andarist:
> now there are 3 local dependencies:

I think lerna should update used versions across all packages when bumping a version, but we'll see.

> server.js and its dependencies are not handled with webpack.

This should be covered with [this](https://github.com/redux-saga/redux-saga/blob/30646f5f15b72ca23cbccba45427d83024376fc2/examples/real-world/.babelrc.js#L20), but I forgot to use env var for this script, we should define `start` script as:
```js
"start": "cross-env BABEL_ENV=cjs node --require @babel/register server.js"
```
and I think it should work.

OTOH not sure why this example has webpack config and if it's used anywhere 😟

restrry:
1 suggested approach works, but  actually it does the same thing as `module: false`. isn't it?
2 webpack.config.js is used [here](https://github.com/redux-saga/redux-saga/blob/master/examples/real-world/server.js#L6)

Andarist:
1. suggested approach is to use opt-in to cjs format, while `modules: false` is an opt-out. In this case it's kinda an equivalent of `modules: 'commonjs'`

I generally like using `modules: false` everywhere as a rule of thumb and opting into cjs on demand with env vars. Generally tooling which supports esm supports cjs too and when using cjs u end up with unoptimized builds and in most cases you are not even warned about it. cjs-only tools fail immediately on esm though and let you opt-in into cjs format easily, so the problem is easy to spot.

Issue #1368, opened 2018-03-17
Why fsmIterator?


NeoLegends:
Hey!

When reading the source code of redux-saga, I stumbled about the fsmIterator approach you're using to implement some of the effects. May I ask why you chose to go that way and implement the state machine yourself instead of using ES6 generators and `yield`?

Thanks!

Andarist:
Hi @NeoLegends! Hope you enjoy reading the source, if you have any questions you can always reach out to me. If you have any thoughts on how it could be improved (both the API/behaviour or the implementation) I would gladly hear you out and ofc you can always make a PR 😉 

Back to the original question - reason is quite simple, with poor-ish es6+ support (many people still need to support IE10) we didn't want to force people to transpile `redux-saga` source to use it. Most people usually ignore `node_modules` in bundlers' rules.

NeoLegends:
I'd like to have something like a combineLatest effect!

It would allow listening to multiple action types and would return an object like the one returned from `race` as soon as its got one action of every type. If it encounters one action multiple times it'll take the latest one.

Andarist:
> It would allow listening to multiple action types and would return an object like the one returned from race as soon as its got one action of every type. If it encounters one action multiple times it'll take the latest one.

This would match rx semantics, right?

> because we cannot return anything custom through yield.

Not sure if I follow could u explain?

The idea seems good, I always wanted to implement more built-in helpers. Do you want to work in this?

NeoLegends:
> This would match rx semantics, right?

Exactly.

> Not sure if I follow could u explain?

As far as I currently understand redux-saga, I cannot have true user-defined effects that return data through the yield like:

```js
// Not possible to return data through yield from the library user perspective today
// because combineLatest is just a library function that uses `fork` and not a real effect
//                         ⇣
const { action1, action2 } = yield combineLatest({
  action1: 'DO_SOMETHING_1',
  action2: 'DO_SOMETHING_2',
});

// use action1, action2
```

Instead, as a user, I must always pass a saga as parameter that is invoked / forked with the data like:

```js
yield combineLatest({
  action1: 'DO_SOMETHING_1',
  action2: 'DO_SOMETHING_2',
  function*({ action1, action2 }) {
    // use action1, action2
  },
});

// cannot use action1, action2 here :/
```

> The idea seems good, I always wanted to implement more built-in helpers. Do you want to work in this?

Sounds good. :) I'll see what I can do.

Andarist:
[...]

PR #1429, opened 2018-05-04
Implement combineLatest effect

NeoLegends:
In response to #1368 I'm finally coming around to implementing the effect. However, I'm a little lost on how the actual implementation within saga's proc.js has to work. Would love some assistance, currently it isn't working as intended, because `runRaceEffect` apparently cannot be used the way I want to use it.

### DIFF
--- a/packages/core/effects.d.ts
+++ b/packages/core/effects.d.ts
@@ -423,6 +423,23 @@ export interface SetContextEffect<C extends object> {
 export function setContext<C extends object>(props: C): SetContextEffect<C>;
 
 
+export type GenericCombineLatestEffectDescriptor<T> = {[key: string]: T};
+
+export interface GenericCombineLatestEffect<T> {
+  COMBINE_LATEST: GenericCombineLatestEffectDescriptor<T>;
+}
+
+export type CombineLatestEffectDescriptor = GenericCombineLatestEffectDescriptor<Effect>;
+
+export interface CombineLatestEffect {
+  COMBINE_LATEST: CombineLatestEffectDescriptor;
+}
+
+export function combineLatest(effects: {[key: string]: Effect}): CombineLatestEffect;
+
+export function combineLatest<T>(effects: {[key: string]: T}): GenericCombineLatestEffect<T>;
+
+
 export interface RootEffect {
   root: true;
   saga(...args: any[]): Iterator<any>;
--- a/packages/core/src/effects.js
+++ b/packages/core/src/effects.js
@@ -23,4 +23,5 @@ export {
   takeLeading,
   throttle,
   delay,
+  combineLatest,
 } from './internal/io'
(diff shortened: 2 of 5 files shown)


Andarist:
I would prefer this being not an actual effect, but rather a saga helper - built on top of internal `fsmIterator`. Let's keep effects (handled by proc.js) small and focused. `combineLatest` can be expressed as a composition of other effects.

As saga is sort of a state machine I'm not sure how `combineLatest`  should work - at least looking at your written test. The same thing can be achieved with just `yield all([take(a), take(b)])`, cant it?

I think `combineLatest` should work more like `takeEvery` - so like this:
```js
function* saga([action_a, action_b]) {
  // do smth
}
const task = yield combineLatest([type_a, type_b], saga)
```

WDYT?

Andarist:
> The difference would be that combineLatest would always use the latest take, while your example would resolve with the first instance of each action that was taken.

Sure, but using it like this:
```js
const [action_a, action_b] = yield combineLatest([type_a, type_b])
```
is not suitable for generators (as they are stateful functions) - cannot inject back new values after we resolve it (`combineLatest`) for the first time.

With `race` the story is different because it alone has no long-lived state, it just resolves and doesnt have to remember prior invocations. While `combineLatest` would have to remember previous invocations and there is no way to glue them together (if you want to use it like in your example), because there is no way to judge if the received effect is the same one that was received before (ofc we could try some caching per saga, but this would overcomplicate things **a lot**)

NeoLegends:
I‘m not sure whether I understand you correctly? Remember the original issue? I think you were talking about the same there until I made myself more clear?

CombineLatest would not have any long-lived-state and would not hae to remember previous invocations. It would resolve only when it has seen an action of all types, discarding duplicate ones.

NeoLegends:
Currently on mobile but combineLatest is supposed to work like: block the saga until you have collected at least one action of every specified type. If you see multiple ones of one type while still waiting for others, replace the previously saved action of that type with the new one.

Andarist:
> I‘m not sure whether I understand you correctly? Remember the original issue? I think you were talking about the same there until I made myself more clear?

I've remembered the issue, just not exact discussion around it. Thanks for clarifying.

Unfortunately I cannot merge this at the moment (until we figure out custom effects), desired effect can be expressed easily as a composition of other effects - just like `takeEvery`, `takeLatest` & `throttle`.

Having it as a built-in effect (handled by `proc`) would prevent removing it from production bundles for people who are not interested in it.

I'd gladly accept version based on `fsmIterator` though.

NeoLegends:
> Unfortunately I cannot merge this at the moment (until we figure out custom effects), desired effect can be expressed easily as a composition of other effects - just like takeEvery, takeLatest & throttle.

You're right, but the ergonomics benefits of this are real. `takeEvery`, `takeLatest`,  and `throttle` are all effects that fire multiple times. This justifies they take a second saga to fork when they resolve, since that function will be called potentially lots of times. `combineLatest` would only resolve once. Thereby a version based on `fsmIterator` has severely crippled readability, because the forked saga would only run once (per effect invocation). Such an implementation does not suffice, unfortunately.

NeoLegends:
I see, seems like I cannot convince you yet. ;)

Wanna leave this open if you decide otherwise in the future?

Andarist:
Sorry that you've put some effort to this PR already, but as stated - to keep the API small & focused I don't want to put more helpers in the core itself. 

I'm going to close this now, but feel free to add any comments if you'd like to.

PR #1430, opened 2018-05-06
Add Retry helper

restrry:
the beginning of conversation: https://github.com/redux-saga/redux-saga/pull/1205
the current implementation is based on `fmsIterator`


packages/core/src/internal/sagaHelpers/retry.js
```diff
@@ -0,0 +1,29 @@
...
+export default function retry(maxTries, delayLength = 0, fn, ...args) {
+  let counter = maxTries
+
+  const yCall = ({ done: false, value: call(fn, ...args) })
+  const yDelay = { done: false, value: delay(delayLength) }
+
+  return fsmIterator(
+    {
+      q1() {
+        return ['q2', yCall, null, 'q10']
```
restrry: we could refactor it to object to simplify signature

Andarist: personally i kinda like tuple-approach, it's really concise - however this is not using it "correctly", tuples should be of fixed size and we return different lengths here - also [accessing out of bounds indexes](https://github.com/redux-saga/redux-saga/pull/1430/files#diff-bcf10a3064550fa33b1534dcf38328f0R37) is a v8 deopt and should be avoided

to sum it up - i agree that we should refactor it to object, i propose `{ nextQ, errQ, eff, updater }`, although we could also go with `{ nextState, errorState, effect, stateUpdater }`, wdyt?

packages/core/src/internal/sagaHelpers/retry.js
```diff
@@ -0,0 +1,29 @@
+import fsmIterator, { qEnd } from './fsmIterator'
+import { call, delay } from '../io'
+
+export default function retry(maxTries, delayLength = 0, fn, ...args) {
+  let counter = maxTries
+
+  const yCall = ({ done: false, value: call(fn, ...args) })
```
Andarist: this is wrapped in parenthesis unnecessarily

packages/core/src/internal/sagaHelpers/retry.js
```diff
@@ -0,0 +1,29 @@
+import fsmIterator, { qEnd } from './fsmIterator'
+import { call, delay } from '../io'
+
+export default function retry(maxTries, delayLength = 0, fn, ...args) {
```
Andarist: using a default parameter in the middle of argument list is rather futile

restrry: yeah, I forgot to remove it :)

packages/core/src/internal/sagaHelpers/retry.js
restrry: 👍 `{ nextState, errorState, effect, stateUpdater }`

Andarist: Sure thing!

PR #1432, opened 2018-05-08
update plugin readme with example and install instruction

restrry:


packages/babel-plugin-redux-saga/README.md
```diff
@@ -6,47 +6,26 @@ Example of setup and demo are available [here](../../examples/error-demo)
...
-    yield 2;
-}
-Object.defineProperty(saga2, Symbol.for("@@redux-saga/LOCATION"), {
-  value: {
-    fileName: "src/sagas/index.js",
-    lineNumber: 5
-  }
-})
+```sh
+npm i -D babel-plugin-redux-saga
```
Andarist: -D is a short hand that many people might not recognize, let's just use more explicit `--save-dev` for clarity

restrry: Done!

Issue #1309, opened 2018-01-18
Revamp testing docs


Andarist:
There is an excellent article by @sh1989 with comparison of redux-saga testing approaches/libs:
http://blog.scottlogic.com/2018/01/16/evaluating-redux-saga-test-libraries.html

It would be super cool to have this text incorporated into our official docs.

zgavin1:
Hi @sh1989 and @Andarist. I'm interested in helping improve the testing docs.

If you had anymore thoughts about this since January, let me know. I can't quite say yet whether concepts from Scott's text should just be more of an addendum or more integrated into the whole document. I'll dive in an see what is best for now.

Andarist:
@zgavin1 Thanks for the contribution offer! We really appreciate it.

The only thing that comes to my mind beside what's already covered by the article is `effect` is that `effectMiddlewares` (available in v1-beta) can be used to intercept effects in tests to implement integration-like tests. Idea is that you can just create a real store, start a saga, intercept/resolve async effects with `effectMiddlewares` and assert on things like state updates etc.

Andarist:
closed with #1433

PR #1433, opened 2018-05-10
#1309 Revamp testing docs

zgavin1:
@Andarist 

Here's a first attempt, I'm super open if you or any others have more suggestions to add or refine this.

#1309

### DIFF
--- a/docs/advanced/Testing.md
+++ b/docs/advanced/Testing.md
@@ -220,3 +220,191 @@ See also: Repository Examples:
 https://github.com/redux-saga/redux-saga/blob/master/examples/counter/test/sagas.js
 
 https://github.com/redux-saga/redux-saga/blob/master/examples/shopping-cart/test/sagas.js
+
+## Testing libraries
+
+While both of the above testing methods can be written natively, there exist several libraries to make both methods easier. Additionally, some libraries can be used to test sagas in a *third* way: recording specific side-effects (but not all).
+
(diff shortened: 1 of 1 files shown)


zgavin1:
I'll see what happened with the build when I have a moment this afternoon.

docs/advanced/Testing.md
```diff
@@ -220,3 +220,256 @@ See also: Repository Examples:
...
+  // assert that your saga does nothing unexpected
+  assert.true(actualEffects.length === 2);
+});
+```
+
+### redux-saga-tester
+
+A final library to consider for integration testing. this library provides a `sagaTester` class, to which you instantiate with your store's initial state and your reducer.
+
+To test your saga, the `sagaTester` instance `start()` method with your saga and it's argument(s). This runs your saga to it's end, and then you may assert that effects occured, actions were dispatched and the state was updated as expected.
```
Andarist: should be `its` instead of `it's`

Andarist:
This looks great 😍 Don't think there is anything to add right now, we can always work on it later too. I've noticed only a single small grammar mistake.

zgavin1:
Cool! Thanks for reviewing it- I'll make that change and find the build issue later tonight.

Andarist:
Build issue is not related to your changes, I won't block this PR because of it.

Andarist:
Thanks, this is AWESOME 🎉

Issue #1316, opened 2018-01-25
Explain root sagas and sagaMiddleware.run better


Andarist:
@lydell has rightfully noticed [here](https://github.com/redux-saga/redux-saga/issues/760#issuecomment-360403490) that those topics might be confusing.

> As of 2018-01-25 the beginner tutorial says:
> 
> export default function* rootSaga() {
>   yield all([
>     helloSaga(),
>     watchIncrementAsync()
>   ])
> }
> As a beginner, I’m going to go with that until I run into problems.
> 
> It would be awesome if there was a section in the docs discussing different ways of implementing the root saga, mentioning pros and cons of each. Also mention what multiple calls to sagaMiddleware.run would mean.

smashercosmo:
Here is my related question on stackoverflow: https://stackoverflow.com/questions/39438005/what-is-the-idiomatic-way-of-starting-rootsaga/47623129

zgavin1:
I'm interested in working on this. I'll find some time in the next day or two.

Andarist:
That would be great!

younesmln:
I'm wondering about the status of this issue because I saw the a related PR has been merged but it still open
I'm interested in helping out if its still an issue

zgavin1:
@Andarist Unless there's more info we want to include, this issue could be closed.

PR #1445, opened 2018-05-19
Issue #1316 explain root sagas

zgavin1:
The only thing missing is I did not address the request:

> Also mention what multiple calls to sagaMiddleware.run would mean.

I don't know what that means. I assume that you would just have duplicates of your root saga started, and therefore duplicates of every effect. That's what happened when tried it in the demo tutorial repo example. I'm not sure why someone would expect differently.


Andarist:
Any reason why you have decided to close the PR?

zgavin1:
Yeah I just wanted to clean up the commit history, I. Not sure what I did
but it was ugly, I just got pulled away right before I was going to submit
again.

PR #1446, opened 2018-05-20
Add docs for root saga implementations.

zgavin1:
The only thing missing is I did not address the request:

> Also mention what multiple calls to sagaMiddleware.run would mean.

I don't know what that means. I assume that you would just have duplicates of your root saga started, and therefore duplicates of every effect. That's what happened when tried it in the demo tutorial repo example. I'm not sure why someone would expect differently.


zgavin1:
@Andarist anything else I can add here for this issue?

docs/basics/RootSaga.md
```diff
@@ -0,0 +1,84 @@
...
+}
+```
+
+This is really just one of a few ways to implement your root. Here, the `all` effect is used with an array of your sagas in a non-blocking fashion. Other root implementations may help you better handle errors and more complex data flow.
+
+Contributor @slorber mentioned in issue #760 several other common root implementations. To start, there are some other implementations you may see that will behave similarly to the  tutorial root saga behavior:
+
+```javascript
+export default function* root() {
+  yield [
```
Andarist: this array should be wrapped in `all`

docs/basics/RootSaga.md
```diff
@@ -0,0 +1,84 @@
...
+
+```javascript
+export default function* root() {
+  yield fork(saga1)
+  yield fork(saga2)
+  yield fork(saga3)
+}
+```
+
+While making use of `fork` rather than `all`, the behavior in your app is the same: all of your sub-sagas are started. The difference between these examples and the original implementation is in execution under the hood. `fork` is non-blocking and so allows the `rootSaga` in these two cases to finish while the child sagas are kept running and blocked by their internal effects.
```
Andarist: I would probably explain that `all([fork, fork, fork])` gives back a single effect (`all`) to be handled to the saga runtime and `fork fork fork` gives a single fork effect 3 times. Ofc ultimately it doesn't matter and both would execute those effects in the same order etc.

docs/basics/RootSaga.md
```diff
@@ -0,0 +1,84 @@
+# Setting up your root Saga
+
+In the beginner tutorial it is shown that your root saga will look something like this
+
+```javascript
+export default function* rootSaga() {
+  yield all([
+    helloSaga(),
```
Andarist: Not sure if mentioning yielding iterators (`helloSaga()`'s result in this case) in yet another place is a good thing, while those are technically handled by redux-saga I would restrain from using it over call/fork which should be the recommended way.

docs/basics/RootSaga.md
```diff
@@ -0,0 +1,84 @@
...
+}
+```
+
+While making use of `fork` rather than `all`, the behavior in your app is the same: all of your sub-sagas are started. The difference between these examples and the original implementation is in execution under the hood. `fork` is non-blocking and so allows the `rootSaga` in these two cases to finish while the child sagas are kept running and blocked by their internal effects.
+
+Error handling in all three implementations is the same. Any error will terminate the root saga and subsequently all other children. (`fork` effects are still connected to their parent.)
+
+## Keeping the root alive
+
+In practice, these implementations aren't terribly practical because your rootSaga will terminate on the first error in any individual child effort or saga. Ajax requests in particular would put your application at the mercy of the status of any endpoints your application makes requests to.
```
Andarist: `effort` -> `effect`

docs/basics/RootSaga.md
```diff
@@ -0,0 +1,84 @@
...
+
+```javascript
+export default function* root() {
+  yield spawn(saga1)
+  yield spawn(saga2)
+  yield spawn(saga3)
+}
+```
+
+Once again, though, uncaught errors are still problematic on a saga-by-saga basis as each saga can still be terminate by such a failure. It's *safer*, but does not provide a meaningful solution.
```
Andarist: We could put more emphasize here (and probably a little bit above too) about need to handle thrown errors. It's developer's responsible to handle potential errors to prevent his/her app from crashing. Making `spawn` a solution for unhandled errors might sometimes hide broken parts from the developer's eyes and should be used carefully. 

It might be wort adding that `spawn` might be considered similar to error boundary in React. The difference is that you can't handle errors & recover when using spawn, but it can be used as extra safety measure at some level of the saga tree, cutting off a single feature or something and not letting the whole app crash.

docs/basics/RootSaga.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Keeping everything alive
+
+We need to think about catching errors separately for each individual saga, and giving that saga the opportunity to restart. In #570 user @granmoe proposed an implementation like this:
+
+```javascript
+function* rootSaga () {
```
Andarist: there was also a discussion about sync errors [here](https://github.com/redux-saga/redux-saga/pull/644#issuecomment-272236599), this could be included here

In general I think we should mention those things as possibility, but should also mention that we do not actually recommend applying it bluntly. Again - it IMHO nicely relate to error boundaries in react. React docs recommend using them for error logging & displaying information to the user about broken part of the page, BUT it is not recommending developers to try to render the same broken subtree again (restarting is imho such a rerender attempt)

docs/basics/RootSaga.md
zgavin1: Well the context here is to reference the root saga implementation in the tutorial, just as a jumping off point. I could use stronger language to suggest that this way is is *not* recommended, but it seems necessary to include atleast some type of nod to the tutorial's root saga.

docs/basics/RootSaga.md
zgavin1: Very cool, thank you for pointing me to that thread. The compositional wrapper proposed look like a pretty cool solution.

I will do my best to integrate that information while trying to stay on the topic of root saga implementations and not going to deep into error handling strategies.

Andarist:
@zgavin1 PR is ready to be merged now, but I think you've rebased or merged something slightly in a wrong way as this PR shows numerous other commits that would be merged with it. Could u prepare a fresh PR with just ur changes?

zgavin1:
Yeah, sorry didn't notice that.

Issue #1447, opened 2018-05-20
Circular reference when building with Rollup


matthewgertner:
I'm trying to build my project that uses `redux-saga` using Rollup, and it reports the following:

```
(!) Circular dependency: node_modules/redux-saga/es/internal/io.js -> node_modules/redux-saga/es/internal/sagaHelpers/index.js -> node_modules/redux-saga/es/internal/sagaHelpers/takeEvery.js -> node_modules/redux-saga/es/internal/io.js
(!) Circular dependency: node_modules/redux-saga/es/internal/io.js -> node_modules/redux-saga/es/internal/sagaHelpers/index.js -> node_modules/redux-saga/es/internal/sagaHelpers/takeLatest.js -> node_modules/redux-saga/es/internal/io.js
(!) Circular dependency: node_modules/redux-saga/es/internal/io.js -> node_modules/redux-saga/es/internal/sagaHelpers/index.js -> node_modules/redux-saga/es/internal/sagaHelpers/throttle.js -> node_modules/redux-saga/es/internal/io.js
```

Is this a known issue or something I should not worry about?

Andarist:
It's both 😄 You don't have to worry about - this circular dep shouldn't cause any trouble, but it's also a known issue. Would you maybe like to restructure the project here to get rid of it? We would appreciate the PR!

matthewgertner:
Ok I'll try to take a look but I'm expecting some `redux-saga` swag if I fix this! 😆

Andarist:
There is already pending PR (#1453 ) for this by @ziel5122

matthewgertner:
Ah ok... guess I'll have to wait for my `redux-saga` coffee mug then.

PR #1453, opened 2018-05-23
changed project structure to fix circular dependency #1447

ziel5122:
- moved functions dependent on 'sagaHelpers' imports to a separate file (io-helper-dep)
- added second export to 'effects' for new file/functions

 https://github.com/redux-saga/redux-saga/issues/1447

As I am not familiar with Rollup

### DIFF
--- a/packages/core/src/effects.js
+++ b/packages/core/src/effects.js
@@ -18,10 +18,13 @@ export {
   flush,
   getContext,
   setContext,
+  delay,
+} from './internal/io'
+
+export {
   retry,
   takeEvery,
   takeLatest,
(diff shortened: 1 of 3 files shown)


packages/core/src/effects.js
```diff
@@ -18,10 +18,13 @@ export {
...
+
+export {
   retry,
   takeEvery,
   takeLatest,
   takeLeading,
   throttle,
-  delay,
-} from './internal/io'
+} from './internal/io-helper-dep.js'
```
Andarist: please change this filename to `io-helpers.js`

PR #1454, opened 2018-05-23
Debounce helper

restrry:
implementation is based on https://github.com/redux-saga/redux-saga/pull/688
the initial discussion is available in the aforementioned PR


packages/core/src/internal/sagaHelpers/debounce.js
```diff
@@ -0,0 +1,41 @@
...
+  const setAction = ac => action = ac
+  const setRaceOutput = ro => raceOutput = ro
+
+  return fsmIterator(
+    {
+      q1() {
+        return {nextState: 'q2', effect: yTake, stateUpdater: setAction}
+      },
+      q2() {
+        return action === END ?
```
restrry: I saw that comment https://github.com/redux-saga/redux-saga/pull/688#pullrequestreview-12891349 but haven't checked yet. I'll check the source code later and we can remove those checks from all helpers once

restrry: I added test for both pattern and channel finished with END, seems that it can be removed across the code base. 
would you mind checking added tests are correct?

PR #1456, opened 2018-05-27
Removing mentions of .done and replacing with .toPromise()

JSoet:
Just updating the readme documentation

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -1240,7 +1240,7 @@ test('my oddOrEven saga', assert => {
     a.equal(
       data.gen.next(2).value,
       'even',
-      'it should yield "event"'
+      'it should yield "even"'
     );
 
     a.equal(
@@ -1261,7 +1261,7 @@ test('my oddOrEven saga', assert => {
     );
 
     a.equal(
-      data.clone.next().toPromise(),
+      data.clone.next().done,
       true,
       'it should be done'
     );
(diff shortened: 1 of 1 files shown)


docs/api/README.md
```diff
@@ -1244,7 +1244,7 @@ test('my oddOrEven saga', assert => {
     );
 
     a.equal(
-      data.gen.next().done,
+      data.gen.next().toPromise(),
```
restrry: that's iterator's flag showing whether the iterator is finished or not. so no needs for changes

Andarist: @JSoet would u like to remove those changes here? The only needed change is [this one](https://github.com/redux-saga/redux-saga/pull/1456/files#diff-0f72f7af9b98c6cb0c52d489836df1e8R965).

JSoet:
Yep, I will but I'm on vacation for the next week and didn't get to this
before leaving, sorry, and I don't think I can make the changes on my
mobile, so if it can wait until I'm back then I'll do it.

Andarist:
No problem, we are not in a rush with this. Enjoy your vacation 👍

JSoet:
K, I changed the other 2 instances back, I assume they were both iterators then. I also noticed a typo in the test output above so changed that while I was at it. Sorry for the delay and the original mistake, and thanks for making a great library :)

Andarist:
Thanks!

Issue #1459, opened 2018-05-29
getContext doesn't work with Symbol


jyboudreau:
#### Description of the bug/issue

Trying to get a context property that is using a Symbol as a key throws an error.

#### Steps to reproduce the bug/issue

1. Set a context using `setContext` effect that uses a Symbol.
2. Get the context using the Symbol as the property

#### Example
https://codesandbox.io/s/pyxx82z0lq

### Actual results
An Error is thrown as the type is being validated as a string. I noticed that in the discussions around context that it seemed Symbol was being considered a valid use case.

### The Expected results
The context is retrieved.

### Environment information

redux-saga v1.0.0-beta.1
see [example](https://codesandbox.io/s/pyxx82z0lq) for other info

PR #1462, opened 2018-05-30
getContext doesn't work with Symbol

restrry:
fix for https://github.com/redux-saga/redux-saga/issues/1459

### DIFF
--- a/packages/core/src/internal/utils.js
+++ b/packages/core/src/internal/utils.js
@@ -25,6 +25,9 @@ export const object = {
         target[i] = source[i]
       }
     }
+    if(Object.getOwnPropertySymbols){
+      Object.getOwnPropertySymbols(source).forEach(s => target[s] = source[s])
+    }
   },
 }
 
@@ -216,4 +219,10 @@ export function createAllStyleChildCallbacks(shape, parentCallback) {
   }
 
   return childCallbacks
+}
+
+export function or(...fns){
+  return function(value){
+    return fns.some(fn => fn(value))
+  }
 }
--- a/packages/core/__tests__/interpreter/context.js
+++ b/packages/core/__tests__/interpreter/context.js
@@ -34,3 +34,32 @@ test('saga must handle context in dynamic scoping manner', () => {
     expect(actual).toEqual(expected)
   })
 })
+
+test('saga must handle Symbol as valid context key', () => {
+  let actual = []
+  const a = Symbol('a')
+  const b = Symbol('b')
+  const context = { [a]: 1 }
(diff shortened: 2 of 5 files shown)


jyboudreau:
Thanks @restrry for a quick fix! 

Btw, I noticed that [here](https://github.com/redux-saga/redux-saga/blob/master/packages/core/src/internal/io.js#L196) we also assert that the argument is a string. Probably needs to change too.

packages/core/src/internal/utils.js
```diff
@@ -39,12 +39,18 @@ export const is = {
 }
 
 export const object = {
-  assign(target, source) {
-    for (let i in source) {
-      if (hasOwn(source, i)) {
-        target[i] = source[i]
+  assign: Object.assign || function (target) {
```
Andarist: this actually won't work as expected. Using `Object.assign` works fine, but `for .. in` doesn't iterate through symbols

Andarist:
@jyboudreau What's ur motivation behind using Symbols for this? Handling Symbols is slightly inconvenient and I'm wondering if we have to do this.

packages/core/src/internal/utils.js
restrry: 1 @Andarist that was done intentionally, since if you use Symbol, you have Object.assign defined in your browser. and I didn't want to add additional check for `Object.getOwnPropertySymbols` existence
2 implementation was borrowed from babel https://github.com/babel/babel/blob/2abd7839e16e3fa63fc381e3ec6ddd83efdf703e/packages/babel-helpers/src/helpers.js#L359

Andarist: > @Andarist that was done intentionally, since if you use Symbol, you have Object.assign defined in your browser. and I didn't want to add additional check for Object.getOwnPropertySymbols existence

In world with polyfills that is unfortunately not true in all the cases. You can polyfill Symbols and skip polyfilling `Object.assign`.

restrry:
updated

Andarist:
Ok, so generally this seems to be ready to merge. But it got me wondering - there is some inconsistency between setContext & getContext effects, 1 accept a map of properties to set and one accept a single prop to get.

Unifying both to accept single props would make the whole problem with symbols just go away (we wouldnt have to use Object.assign or any kind of equivalent) - we'd just do regular assignments. OTOH it would be far less ergonomic to use :s

restrry:
@Andarist OTOH if you need to assign several props you can use. or we can provide such a helper
```js
[
  ['a', 1],
  ['b', 2]
].forEach(([key, value]) => setContext(key, value))
```

PR #1463, opened 2018-05-30
chore: npm doesnt support node 10

restrry:
```sh
npm does not support Node.js v10.3.0
npm WARN npm You should probably upgrade to a newer version of node as we
npm WARN npm can't make any promises that npm will work with this version.
npm WARN npm Supported releases of Node.js are the latest release of 4, 6, 7, 8, 9.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
-  - "8"
-  - "10"
+  - 8
+  - 9
 
 install:
   - npm run bootstrap


restrry:
@Andarist by default node v10.0.0-v10.2.x are shipped with npm 5.6 https://nodejs.org/en/download/releases/

we can add more sophisticated logic to run our test in different environments  or run all test only on node v10.3 with npm 6. whdyt?

Andarist:
Do u know why those has started to fail anyway? What has changed? Weren't tests passing on node@10 previously?

restrry:
I suspect Travis made the latest node release (10.3.0) available
node v10.2.1, npm 5.6.0  ✅  https://travis-ci.org/redux-saga/redux-saga/jobs/383595522
node v10.3.0 npm 6.1.0  🔴 https://travis-ci.org/redux-saga/redux-saga/jobs/387408815

as I can see the latter build was run with npm 6.1.0, so we need to figure out what causes the problem.
log with node v10.3.0, npm 6.1.0 (installed via n) 
```
 npm i
npm ERR! asyncWrite is not a function
npm ERR! asyncWrite is not a function
npm ERR! pna.nextTick is not a function
```

Andarist:
Can we just pin the version to 10.2.1 for the time being? If not then just merge this in its current form for the time being.

PR #1464, opened 2018-06-01
clean up channel end

restrry:
updated version of https://github.com/redux-saga/redux-saga/pull/724 by @Andarist

### DIFF
--- a/packages/core/src/internal/proc.js
+++ b/packages/core/src/internal/proc.js
@@ -1,6 +1,6 @@
 import {
   CANCEL,
-  CHANNEL_END as CHANNEL_END_SYMBOL,
+  TERMINATE,
   TASK,
   TASK_CANCEL as TASK_CANCEL_SYMBOL,
   SELF_CANCELLATION,
@@ -40,20 +40,18 @@ function getIteratorMetaInfo(iterator, fn) {
   return getMetaInfo(fn)
 }
 
-// TODO: check if this hacky toString stuff is needed
-// also check again whats the difference between CHANNEL_END and CHANNEL_END_TYPE
+// toString is used to have ability to log actions in devtools.
 // maybe this could become MAYBE_END
 // I guess this gets exported so takeMaybe result can be checked
-export const CHANNEL_END = {
-  toString() {
-    return CHANNEL_END_SYMBOL
-  },
-}
+
 export const TASK_CANCEL = {
   toString() {
     return TASK_CANCEL_SYMBOL
   },
 }
+
+const shouldTerminate = (res) => res && (isEnd(res) || res[TERMINATE] || res === TASK_CANCEL)
+const isTerminate = res => res && res[TERMINATE]
 
 /**
   Used to track a parent task and its forks
@@ -294,8 +292,8 @@ export default function proc(
           This will jump to the finally block
         **/
         result = is.func(iterator.return) ? iterator.return(TASK_CANCEL) : { done: true, value: TASK_CANCEL }
-      } else if (arg === CHANNEL_END) {
-        // We get CHANNEL_END by taking from a channel that ended using `take` (and not `takem` used to trap End of channels)
+      } else if (isTerminate(arg)) {
+        // We get TERMINATE flag, i.e. by taking from a channel that ended using `take` (and not `takem` used to trap End of channels)
         result = is.func(iterator.return) ? iterator.return() : { done: true }
       } else {
         result = iterator.next(arg)
@@ -483,7 +481,7 @@ export default function proc(
         return
       }
       if (isEnd(input) && !maybe) {
-        cb(CHANNEL_END)
+        cb({ ...input, [TERMINATE]: true })
         return
       }
       cb(input)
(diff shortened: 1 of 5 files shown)


packages/core/src/internal/proc.js
```diff
@@ -40,21 +40,19 @@ function getIteratorMetaInfo(iterator, fn) {
   return getMetaInfo(fn)
 }
 
-// TODO: check if this hacky toString stuff is needed
-// also check again whats the difference between CHANNEL_END and CHANNEL_END_TYPE
+// toString is used to have ability to log actions in devtools.
```
restrry: do we really need it? browser devtools can show Symbol correctly. any custom solution could achieve it with `String(symbol)`

packages/core/src/internal/proc.js
```diff
@@ -40,21 +40,19 @@ function getIteratorMetaInfo(iterator, fn) {
...
-  },
-}
+
 export const TASK_CANCEL = {
   toString() {
     return TASK_CANCEL_SYMBOL
   },
 }
 
+const shouldTerminate = (res) => res && (isEnd(res) || res[TERMINATE] || res === TASK_CANCEL)
```
Andarist: I get a feeling that something is off with our `prettier`/`lint-staged` config, single arrow param should be reformatted without parens 🤔

this needs to be investigated, but not as part of this PR

packages/core/src/internal/proc.js
Andarist: I think this comment was referring to https://github.com/redux-saga/redux-saga-devtools but this project is unmaintained atm, so probably it doesnt matter much and as mentioned other workarounds exist

restrry: should we delete the construction in this case?

Andarist: which one? you already have deleted `CHANNEL_END` from here, which I think is fine 👍 

I think this comment "toString is used to have ability to log actions in devtools." might be removed too.

packages/core/src/internal/proc.js
```diff
@@ -294,8 +286,8 @@ export default function proc(
           This will jump to the finally block
         **/
         result = is.func(iterator.return) ? iterator.return(TASK_CANCEL) : { done: true, value: TASK_CANCEL }
-      } else if (arg === CHANNEL_END) {
-        // We get CHANNEL_END by taking from a channel that ended using `take` (and not `takem` used to trap End of channels)
+      } else if (isTerminated(arg)) {
```
Andarist: wouldnt `shouldTerminate` be more appropriate? just thinking out loud, let me know what do u think

packages/core/src/internal/proc.js
```diff
@@ -40,21 +34,19 @@ function getIteratorMetaInfo(iterator, fn) {
...
-  },
-}
+
 export const TASK_CANCEL = {
   toString() {
     return TASK_CANCEL_SYMBOL
   },
 }
 
+const isTerminated = res => res && res[TERMINATE]
```
Andarist: could u explain [this comment](https://github.com/redux-saga/redux-saga/pull/1464#discussion_r192986758)? Im not sure if I get what case are you talking about

packages/core/src/internal/proc.js
```diff
@@ -40,21 +40,19 @@ function getIteratorMetaInfo(iterator, fn) {
...
-}
+
 export const TASK_CANCEL = {
   toString() {
     return TASK_CANCEL_SYMBOL
   },
 }
 
+const shouldTerminate = (res) => res && (isEnd(res) || res[TERMINATE] || res === TASK_CANCEL)
+const isTerminate = res => res && res[TERMINATE]
```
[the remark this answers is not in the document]
restrry: in normal flow `isTerminate` is being used in `next` callback, which receives the result of `digestEffect`. but when we kick off generator for the very first time we call `next()` without passing any result.
so at the initial step `arg, isErr` are undefined ==> `isTerminate` will be called with undefined as well

packages/core/src/internal/proc.js
restrry: well, actually we just received request to terminate, so probably you are right and it should be called `should...`

packages/core/src/internal/proc.js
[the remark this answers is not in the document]
Andarist: gotcha

packages/core/src/internal/proc.js
```diff
@@ -483,7 +475,7 @@ export default function proc(
         return
       }
       if (isEnd(input) && !maybe) {
-        cb(CHANNEL_END)
+        cb({ ...input, [TERMINATE]: true })
```
Andarist: can u think of any use case for extending the `input` with a flag? maybe we could just pass special completion value - `TERMINATE` here, like this:
```js
if (isEnd(input) && !maybe) {
  cb(TERMINATE)
  return
}
```

I know I'm probably one who wrote this spread here, but can't think of any use case of it now.

If we would follow that path, then we could skip implementing `isTerminated` altogether and just use direct `TERMINATE` comparisong, i.e.
```js
 } else if (arg === TERMINATE) {
```
and
```js
const isCompleted = res => res && (isEnd(res) || res === TERMINATE || res === TASK_CANCEL)
```

restrry: checked, also didn't find any cases, updated

PR #1465, opened 2018-06-01
Consider avoiding assumptions about how easy sagas/generators are

janhoogeveen:
First off, I really like the documentation of redux-sagas and it does help me learn to understand sagas in small steps.

One thing I couldn't avoid however is to feel uneasy about how the docs talk about beginner examples. Concepts and examples are presented as "simple" and "simply do this" while in all honesty some of the concepts are pretty hard to grasp, especially for beginners. 

I'd propose to change a lot of these sentences to use the word _basic_ instead. Basic means _the examples are slimmed down to their fundamentals_, but doesn't imply they are _easy_ or _simple to grasp_.

After all, if something is easy or not differs per person.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 [![OpenCollective](https://opencollective.com/redux-saga/backers/badge.svg)](#backers)
 [![OpenCollective](https://opencollective.com/redux-saga/sponsors/badge.svg)](#sponsors)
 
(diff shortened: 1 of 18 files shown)


Andarist:
Thanks for the PR. I agree with u, that we should probably avoid such words.

Issue #1399, opened 2018-04-14
Some questions about source code.


shinima:
Question 1:  I found that `mainTask.isMainRunning` is set to `false` [here](https://github.com/redux-saga/redux-saga/blob/d98344e4fd6a7be561bd26f58d5c831e3b14c1b8/packages/core/src/internal/proc.js#L314) and [here](https://github.com/redux-saga/redux-saga/blob/d98344e4fd6a7be561bd26f58d5c831e3b14c1b8/packages/core/src/internal/proc.js#L321), but I cannot find where `mainTask.isMainRunning` is used. I am wandering whether it is a mistake.

Question 2: [Here](https://github.com/redux-saga/redux-saga/blob/d98344e4fd6a7be561bd26f58d5c831e3b14c1b8/packages/core/src/internal/proc.js#L301) that a `CHANNEL_END` will make the iterator jump to finally block without cancelling its attached forks. I know its behavior, but I cannot understand its intention. Can you explain why? An example will be grateful.

Andarist:
1. probably some leftover, would u like to send a PR removing those assignments? OTOH there is a chance that we should set `isRunning` to false there, but it would require more thorough investigation

2. truth to be told - im not sure about this, `CHANNEL_END` is a somewhat hacky solution for SSR - I guess the intention is that every started task should be completed until it hits `take` effect which ends this particular task with `CHANNEL_END` but not necessarily its forks - they should end in the same way (by getting `take` effect)

shinima:
1. I have made a PR 😄 
2. Let's just leave it alone.

And... Question 3:

[Here](https://github.com/redux-saga/redux-saga/blob/7b415a2408f9fb9002abae1112dce3b6bb5f6ed6/packages/core/src/internal/proc.js#L774-L778) we are using properties of iterator to track the parent-task status, but actually [_the parent task is the  aggregation of the main tasks + all its forked tasks_](https://github.com/redux-saga/redux-saga/blob/master/packages/core/src/internal/proc.js#L62). 
So, in current implementation, there are confusing situations if we run the following code:

```JavaScript
function* genFn() {
  yield fork(backgroundTask)
  // mainTask.isRunning is set to false here
  // But since the backgroundTask is still running, so the parent-task is still running
  // We are using iterator._isRunning to track parent-task so iterator._isRunning is true.
  // It's confusing that:
  //    mainTask.isRunning === false && iterator._isRunnig === true
}
```

[...]

PR #1468, opened 2018-06-03
internal: Rename `iterator._XXX` to more explicit names.

shinima:
From #1399

- Rename `_isRunning` to `_isTaskRunning` (same for `_isCancelled` and `_isAborted`)
- Rename `_result` to `_taskResult`; rename `_error` to `_taskError`.

### DIFF
--- a/packages/core/src/internal/proc.js
+++ b/packages/core/src/internal/proc.js
@@ -237,8 +237,8 @@ export default function proc(
       We need to check both Running and Cancelled status
       Tasks can be Cancelled but still Running
     **/
-    if (iterator._isRunning && !iterator._isCancelled) {
-      iterator._isCancelled = true
+    if (iterator._isTaskRunning && !iterator._isTaskCancelled) {
+      iterator._isTaskCancelled = true
       taskQueue.cancelAll()
       /**
         Ending with a Never result will propagate the Cancellation to all joiners
@@ -253,7 +253,7 @@ export default function proc(
   cont && (cont.cancel = cancel)
 
   // tracks the running status
-  iterator._isRunning = true
+  iterator._isTaskRunning = true
 
   // kicks up the generator
   next()
@@ -320,11 +320,11 @@ export default function proc(
   }
 
   function end(result, isErr) {
-    iterator._isRunning = false
+    iterator._isTaskRunning = false
     // stdChannel.close()
 
     if (!isErr) {
-      iterator._result = result
+      iterator._taskResult = result
       iterator._deferredEnd && iterator._deferredEnd.resolve(result)
     } else {
       addSagaStack(result, {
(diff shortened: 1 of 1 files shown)


Andarist:
This looks good, I'm only thinking that maybe keeping those on `iterator` is somewhat leaky. It feels wrong to mutate it. Maybe we could try to keep those flags on the task's instance instead? It's our custom object and makes more sense to keep it there. WDYT?

shinima:
I came up with several options:

### Option-1
Like in this PR now, rename `_isRunning` to `_isTaskRunning` (same for `_isCancelled` and `_isAborted`)
rename `_result` to `_taskResult`; rename _error to `_taskError`.

* pros: make the names more explicit.
> cons: keeping those on iterator is somewhat leaky

### Option-2
Convert `task.isRunning / task.isCancelled ...` from methods to properties, and use these properties to track the running status of the task. I have tried this option in my [little-saga](https://github.com/shinima/little-saga/blob/master/src/core/Task.js#L4-L9), and it works pretty well (passed almost all of redux-saga test cases).
* pros: It is simple, intuitive and explicit.
* cons: It introduces breaking changes since that the [API doc](https://github.com/redux-saga/redux-saga/tree/master/docs/api#task) already says `task.isRunning ...` are methods.

### Option-3

Like in #1404 , using properties and methods at the same time. Using properties `task._isRunning / task_.isCancelled ... ` to track running status, while still provides `task.isRunning() / task.isCancelled()`.

* pros: get rid of `iterator._xxx`
* cons: having methods and properties at the same time seems unnecessary

### My opinion
I think `Task.isRunning() / Task.isCancelled()...` are not popular APIs and we are toward version 1.0, so option-2 is the best option. How do you think ?

Andarist:
I'm not against introducing breaking changes for this, so option 2 is fine. It has additional drawback though - properties might imply that they can be changed and those should really be read-only outside of redux-saga code. So I'm kinda leaning towards option 3 at the moment.

> I have tried this option in my little-saga

That is really cool! What are your goals for this project? You might be also interested in [this project](https://github.com/briancavalier/forgefx).

Andarist:
@restrry WDYT about this?

shinima:
little-saga is a slim version of redux-saga. I made some breaking change (including using `task.isRunning / task.isCancelled ...` to track running status) to keep little-saga as simple as possible. And I wrote an article in Chinese to help people understand the internals of redux-saga/little-saga.

> properties might imply that they can be changed and those should really be read-only outside of redux-saga code

We can document that `task.isRunning / task.isCancelled` is read-only and add `readonly` modifier in typescript declaration file. On the other hand, I think most users will use join/cancel/cancelled effects 
rather than using these APIs to inspect the running status. These APIs could be think as advanced APIs and maybe we can trust these advanced APIs users 😆 .

@Andarist If you think option-2 is okay, I am glad to make a PR. Or if you think option-3 is better, I can reopen #1404.

restrry:
👍 for option-3.
pros: no breaking changes. and if we need add some addition logic (in future) for getting/setting those flags, we already have isolation from external api.
the only remark - does it make sense to rename mainTask.isRunning -> `mainTask._isRunning` to have  make them more consistent?

shinima:
So let's pick option-3.   `mainTask._isRunning` is very reasonable though mainTask is private in `proc()`. I will add commits in #1475.

PR #1470, opened 2018-06-04
Adding a few links to some pattern articles.

bfillmer:
A couple of my more popular articles on using Sagas to handle various use-cases.


Andarist:
Thanks, those are great !

PR #1475, opened 2018-06-05
Use properties of task to track the task running status.

shinima:
Implementation for  #1468  Option-3

### DIFF
--- a/packages/core/src/internal/proc.js
+++ b/packages/core/src/internal/proc.js
@@ -237,8 +237,8 @@ export default function proc(
       We need to check both Running and Cancelled status
       Tasks can be Cancelled but still Running
     **/
-    if (iterator._isRunning && !iterator._isCancelled) {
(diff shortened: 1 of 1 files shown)


Andarist:
Could u rebase this? CI fix got merged onto master today.

shinima:
Rebased.  @Andarist

shinima:
@restrry  I have changed  `mainTask.isRunning`  to `mainTask._isRunning`. How do you think?

Andarist:
Not sure if this rename is needed, mainTask is internal anyway. I'm also not strongly opposed to it either. Whatever u both work out regarding this one.

restrry:
regarding naming - my confusion was about having booleans with the same semantic `task._isRunning` and `mainTask.isRunning` , but `task.isRunning()` is a method at the same time.
we can leave everything as is, but I'd prefer to have any kind of unification for naming :)

PR #1477, opened 2018-06-07
Refactor effect structure to `{ [IO]: true, type, payload }`

shinima:
[In #1263](https://github.com/redux-saga/redux-saga/pull/1263#discussion_r150802414) you both agreed refactoring effect structure to `{ [IO]: true, type, payload }`.

But there are too many conflicting files to merge #1263 currently, so I extract the  refactoring-effect-structure part in this PR.

I will dive into #1263 to see if I can continue the work.


packages/core/src/internal/io.js
```diff
@@ -209,7 +210,7 @@ export function setContext(props) {
 
 export const delay = call.bind(null, delayUtil)
 
-const createAsEffectType = type => effect => effect && effect[IO] && effect[type]
+const createAsEffectType = type => effect => effect && effect[IO] && effect.type === type && effect.payload
 
 export const asEffect = {
```
shinima: Maybe we can remove `asEffect`, and place the logic in runEffect. So in `proc.js` it would look like:
```javascript
  if(effect[IO] === true) {
    const { type, payload } = effect
    if (type === 'TAKE') {
      runTakeEffect(payload, currCb)
    } else if (type === 'PUT') {
      runPutEffect(payload, currCb)
    } // ......
  }
}
```
`asEffect` is exported by `redux-saga/utils`.  Maybe it is not a good idea to refactor `asEffect`. How do you think about it?

Andarist: I don't know how `asEffect` was used outside of the core, let's just ditch it for now & wait to see if it breaks anyone's use case. We are in v1-beta period right now, so I'm not concerned with making some breaking changes.

I like refactoring `runEffect` in the way you have proposed here.

packages/core/src/internal/io.js
```diff
@@ -89,6 +89,7 @@ function getFnCallDesc(meth, fn, args) {
   }
 
   let context = null
+  // prettier-ignore
```
Andarist: whats the reason behind this comment?

shinima: In .lintstagedrc we've configured that staged files will run through `prettier` and `eslint --fix`.

If we run prettier only,  it will add semicolons before `[` and `(`.

image

These semicolons then are removed by `eslint --fix`.

packages/core/src/internal/io.js
shinima: `asEffect` seems like a inspection/test utils. (#1260)
I will continue refactoring `runEffect`. If we refactor the effect structure, it would be more easier for users to do inspections/tests.

packages/core/src/internal/proc.js
```diff
@@ -350,31 +351,33 @@ export default function proc(
...
+      : (type === EFFECT_TYPES.FORK)           ? runForkEffect(payload, effectId, currCb)
+      : (type === EFFECT_TYPES.JOIN)           ? runJoinEffect(payload, currCb)
+      : (type === EFFECT_TYPES.CANCEL)         ? runCancelEffect(payload, currCb)
+      : (type === EFFECT_TYPES.SELECT)         ? runSelectEffect(payload, currCb)
+      : (type === EFFECT_TYPES.ACTION_CHANNEL) ? runChannelEffect(payload, currCb)
+      : (type === EFFECT_TYPES.FLUSH)          ? runFlushEffect(payload, currCb)
+      : (type === EFFECT_TYPES.CANCELLED)      ? runCancelledEffect(payload, currCb)
+      : (type === EFFECT_TYPES.GET_CONTEXT)    ? runGetContextEffect(payload, currCb)
+      : (type === EFFECT_TYPES.SET_CONTEXT)    ? runSetContextEffect(payload, currCb)
+      : kThrow(`runEffect(effect...) invalid effect type: ${type}`)
```
shinima: `kThrow('runEffect(effect...) invalid effect type: ${type}')` could be thought as a fallback `runCustomEffect`. 
Once #1479 is merged, I think we may use `env.runCostumEffect` here, that will allow running user-defined effects.

restrry:
there are some external consumers of `asEffect`. it seems to be used more as `is.effect.fork(..`
https://github.com/jfairbank/redux-saga-test-plan/blob/master/src/expectSaga/parseEffect.js
https://github.com/jfairbank/redux-saga-test-plan/blob/master/src/expectSaga/provideValue.js
so if we delete it, we need to provide an alternative solution

shinima:
We may export `IO` symbol so that user could directly use `obj && obj[IO] && obj.type === 'FORK'` to test if an object is an fork-effect. I think exposing the structure of effects to users make effects easier to understand.
Maybe I should keep `asEffect` now but add a deprecate warning.

restrry:
that breaks encapsulation, imho we should provide a helper that hides effect shape

shinima:
@restrry I agree that we shouldn't break encapsulation. I need to do more investigation before refactoring `asEffect`. I'll reverting deleting `asEffect`.

packages/core/src/internal/io.js
restrry: so we need to change our setup? otherwise we have to add that comment manually for all cases

shinima: I think we can turn off *no-extra-semi* rule in ESLint. Using prettier to remove extra semicolons is enough. 😃

Andarist:
I think we can safely remove `asEffect` for the time being and just re-add it later in other form - probably some mix of current `asEffect` and [this](https://github.com/redux-saga/redux-saga/pull/1262/files). This API seems not like a core feature, but rather an inspection util. So IMHO would be better to just take it out of the picture for now.

Andarist:
I've liked the `effectTypes` change that I think you have reverted, I think it should get reinstated - only I would create a separate file for it and import them with a namespace import (`import * as effectTypes`), that way the code could be slightly better minified than in case of exported object.

packages/core/src/internal/proc.js
```diff
@@ -330,31 +330,31 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
...
-      : (data = asEffect.cancel(effect))        ? runCancelEffect(data, currCb)
-      : (data = asEffect.select(effect))        ? runSelectEffect(data, currCb)
-      : (data = asEffect.actionChannel(effect)) ? runChannelEffect(data, currCb)
-      : (data = asEffect.flush(effect))         ? runFlushEffect(data, currCb)
-      : (data = asEffect.cancelled(effect))     ? runCancelledEffect(data, currCb)
-      : (data = asEffect.getContext(effect))    ? runGetContextEffect(data, currCb)
-      : (data = asEffect.setContext(effect))    ? runSetContextEffect(data, currCb)
-      : /* anything else returned as is */        currCb(effect)
-    )
+    if (is.promise(effect)) {
```
Andarist: we should remove `return`s from this function, as this function `runEffect` shouldnt actually return anything to the caller (we can ofc use `return` to exit the function, just we shouldnt return any value)

shinima:
I've removed the unnecessary `return` keywords and turned off no-extra-semi eslint rule.

PR #1479, opened 2018-06-07
internal: Saga env

shinima:
Currently, function `proc()` has too many parameters (9 parameters). However, after root saga starts, some parameters never changes. Most of the calls to `proc()` just pass them down to child sagas. These parameters includes:

 * stdChannel
 * dispatch
 * getState
 * options.sagaMonitor
 * options.logger
 * options.onError
 * options.middleware

These parameter could be thought as **saga running environments**. In this PR I pack them into a single object called `env`. Other `proc()` parameters are task-specific, that means different calls to `proc()` would pass different values to these parameters.

Running sagas can read from the `env` to complete specific tasks, for example, logging an unhandled error. Running sagas should treat the `env` read-only and should not mutate `env`.

IMHO, introducing `env` will bring the following benefits:

* reduce the number of parameters of proc()
* better distinction between **Running Environments** and **Task Specific Parameters**
* Like [this post](https://github.com/redux-saga/redux-saga/pull/1263#discussion_r150841895) says, we may let the user to pre-process the `env` object before root-saga starts. Maybe users could inject custom effect-types and effect-runners by setting something like `env.runCustomEffects(...)`, and in `proc()` when we meet effects with unknown type, we use `env.runCustomEffects(...)` to run the effect. Anyway, introducing `env` would make it easier to extend redux-saga.

### DIFF
--- a/packages/core/src/internal/proc.js
+++ b/packages/core/src/internal/proc.js
@@ -149,19 +149,8 @@ function createTaskIterator({ context, fn, args }) {
       )
 }
 
-export default function proc(
-  iterator,
-  stdChannel,
-  dispatch = noop,
-  getState = noop,
-  parentContext = {},
-  options = {},
-  parentEffectId = 0,
-  meta,
-  cont,
-) {
-  const { sagaMonitor, logger, onError, middleware } = options
-  const log = logger || _log
+export default function proc(iterator, parentContext = {}, env, parentEffectId = 0, meta, cont) {
+  const log = env.logger || _log
 
   const logError = err => {
     log('error', err)
@@ -318,8 +307,8 @@ export default function proc(
           result.sagaStack = sagaStackToString(result.sagaStack)
         }
 
-        if (onError) {
-          onError(result)
+        if (env.onError) {
+          env.onError(result)
         } else {
           // TODO: could we skip this when _deferredEnd is attached?
           logError(result)
@@ -379,7 +368,7 @@ export default function proc(
 
   function digestEffect(effect, parentEffectId, label = '', cb) {
     const effectId = nextEffectId()
-    sagaMonitor && sagaMonitor.effectTriggered({ effectId, parentEffectId, label, effect })
+    env.sagaMonitor && env.sagaMonitor.effectTriggered({ effectId, parentEffectId, label, effect })
 
     /**
       completion callback and cancel callback are mutually exclusive
@@ -396,8 +385,12 @@ export default function proc(
 
       effectSettled = true
       cb.cancel = noop // defensive measure
-      if (sagaMonitor) {
-        isErr ? sagaMonitor.effectRejected(effectId, res) : sagaMonitor.effectResolved(effectId, res)
+      if (env.sagaMonitor) {
+        if (isErr) {
+          env.sagaMonitor.effectRejected(effectId, res)
+        } else {
+          env.sagaMonitor.effectResolved(effectId, res)
+        }
       }
       if (isErr) {
         crashedEffect = effect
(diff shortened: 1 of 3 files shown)


packages/core/src/internal/proc.js
```diff
@@ -318,11 +297,11 @@ export default function proc(
           result.sagaStack = sagaStackToString(result.sagaStack)
         }
 
-        if (onError) {
-          onError(result)
+        if (env.onError) {
```
Andarist: we could be more explicit here and do `if (typeof env.onError === 'function')`

packages/core/src/internal/runSaga.js
```diff
@@ -47,18 +47,27 @@ export function runSaga(options, saga, ...args) {
...
+      log('error', err.sagaStack)
+    }
+  }
+
   const middleware = effectMiddlewares && compose(...effectMiddlewares)
 
-  const task = proc(
-    iterator,
+  const env = {
     channel,
```
Andarist: i would prefer naming this `stdChannel` as its not just any channel

packages/core/src/internal/proc.js
```diff
@@ -149,27 +148,7 @@ function createTaskIterator({ context, fn, args }) {
...
-  const log = logger || _log
-
-  const logError = err => {
-    log('error', err)
-    if (err && err.sagaStack) {
-      log('error', err.sagaStack)
-    }
-  }
-
+export default function proc(iterator, parentContext, parentEffectId, meta, env, cont) {
```
Andarist: imho we should reorder those args

`env` & `iterator` should be first 2, im a little bit more inclined to put `env` as first as its constant and making it first it would be easier to curry `proc` (although not something we actually want to do)

`meta` should become very last as its the least important one and I think we should also check if its possible to put `parentEffectId` in the `meta` object too

OTOH `cont` is callback and it seems reasonable to put it last, maybe `cont` should be the one before last one (`meta`)?

let me know what do u think

Andarist:
Very nice change 👍

packages/core/src/internal/runSaga.js
shinima: Yes, `channel` is too vague. How do you think naming this `defaultChannel` ?  

`stdChannel`  is already a function defined in *src/internal/channel.js*, and I am confused why it is a **standard channel**.

packages/core/src/internal/proc.js
shinima: I think moving `env` to first is a good idea. I do not have opinions for the order of other arguments. 

Though very important, `proc()` is only called in two other places: one in `runForkEffect()`, and one in `runSaga()`. I will move `env` to first now, and we can refactor the order later on if necessary.

packages/core/src/internal/proc.js
shinima: I've add check for `onError` in runSaga, so it must be a function if pass `if (onError)` here.

shinima:
How do think we rename `env.stdChannel` to `env.defaultChannel` ?  `stdChannel` is already a function defined in src/internal/channel.js, and I am confused why it is a **standard** channel.

packages/core/src/internal/proc.js
restrry: why not use hash to imitate named arguments (instead of working with positional ones as we do now)?

shinima: It's just a matter of flavor. I prefer positional arguments for most cases. And for options, I prefer object/hash.

shinima:
I've updated the code, please have a look when you are free @Andarist .

packages/core/src/internal/runSaga.js
restrry: `stdChannel` from `src/internal/channel.js` is channel factory. isn't it? dunno worth renaming it or not

restrry:
👍 overall

Andarist:
After a thought - let's just keep the `channel` there, for the `proc` it doesn't really matter what kind of channel it is. It only expects certain interface.

As to `stdChannel` - not the best name, didnt think of any better name at a time of implementing it. It's just a multicasting channel with some scheduling involved, but the scheduling thing is very proc-specific.

PR #1481, opened 2018-06-07
internal: Use env.finalizeRunEffect to simplify the implementation of digestEffect.

shinima:
This PR is based on #1479. If we introduce the `env`, then we can use `env.finalizeRunEffect` which converts runEffect to finalRunEffect [as the lines commented](https://github.com/redux-saga/redux-saga/blob/b956a0913028ffcc486c830290012c6ec9ff751d/packages/core/src/internal/proc.js#L433-L436). 

This PR remove passing `middleware` to `proc()`, since the running saga now depends `env.finalizeRunEffect` instead of `middleware`. This simplifies the implementation of `digestEffect`.

`env.finalizeRunEffect` could be think as an environment function, because it is determined before the root saga starts and never changes once created.

I think it is a good use case of `env`. Maybe we could do more enhancement using `env` afterwards. 
How do you think?

### DIFF
--- a/packages/core/src/internal/proc.js
+++ b/packages/core/src/internal/proc.js
@@ -150,6 +150,7 @@ function createTaskIterator({ context, fn, args }) {
 
 export default function proc(iterator, parentContext, parentEffectId, meta, env, cont) {
   const taskContext = Object.create(parentContext)
+  const finalRunEffect = env.finalizeRunEffect(runEffect)
 
   let crashedEffect = null
   const cancelledDueToErrorTasks = []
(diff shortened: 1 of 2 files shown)


packages/core/src/internal/proc.js
```diff
@@ -150,6 +150,7 @@ function createTaskIterator({ context, fn, args }) {
 
 export default function proc(iterator, parentContext, parentEffectId, meta, env, cont) {
   const taskContext = Object.create(parentContext)
+  const finalRunEffect = env.finalizeRunEffect(runEffect)
```
shinima: Maybe `enhancedRunEffect` and `env.enhanceRunEffect` are better names. I don't like marking things as finalized, since we will change them again someday.

packages/core/src/internal/runSaga.js
```diff
@@ -62,6 +62,16 @@ export function runSaga(options, saga, ...args) {
   }
 
   const middleware = effectMiddlewares && compose(...effectMiddlewares)
+  const finalizeRunEffect = runEffect => {
```
Andarist: this in fact can be cached now,
```js
const finalizeRunEffect = !is.func(middleware) 
    ? runEffect
    : runEffect => {
      if () {
        return function finalRunEffect(effect, effectId, currCb) {
          const plainRunEffect = eff => runEffect(eff, effectId, currCb)
          return middleware(plainRunEffect)(effect)
        }
      }
    }
```

Issue #1476, opened 2018-06-06
Partially missing docs for call, cps, and fork


chrissantos1995:
### Description
My team is using the syntax below to explicitly pass `context` and `fn` to `fork`, `call` and `cps`:

### `fork({context, fn}, ...args)`

It seems that the [API reference](https://redux-saga.js.org/docs/api/) is missing documentation for this style of invocation -- do we know if these are being omitted intentionally?

If it's simply missing from the documentation -- let me know if you need help with the PR for adding these. 

As a followup, it'd be great to have these added to the flow-typed definitions. I can help with that as well if we want to move forward with this change.

shinima:
The code that handles this style of invocation is [here](https://github.com/redux-saga/redux-saga/blob/8555610cbd1f09176546edea7b01383714b8ed48/packages/core/src/internal/io.js#L94-L96). I think the documentation is missing this style of invocation. Welcome to make a PR.

The typescript declaration file already contains [type definitions for this style of invocation](https://github.com/redux-saga/redux-saga/blob/8555610cbd1f09176546edea7b01383714b8ed48/packages/core/effects.d.ts#L150-L155).

chrissantos1995:
Thanks shinima -- I opened a PR [here](https://github.com/redux-saga/redux-saga/pull/1484) for doc additions. I'll pursue the type additions in the flow-typed repo.

Andarist:
closed by #1484

PR #1484, opened 2018-06-09
Add docs for object invocation for call, cps, fork #1476

chrissantos1995:
Adds missing documentation for passing `context` and `fn` as properties in an object to `call`, `cps`, and `fork`.  [#1476](https://github.com/redux-saga/redux-saga/issues/1476)


shinima:
This looks good to me. But I would like to let @Andarist or @restrry check this PR since I'm not very good at English.

Andarist:
Thanks!

PR #1486, opened 2018-06-10
add codesandbox template for issues

restrry:
I added `cancellable-counter` example as a project for template, without babel setup though. what else should we add?


Andarist:
I think this is quite enough, I don't think babel can be configured on codesandbox anyway.

restrry:
jfyi It can be, I did another example with it, but adding 2 templates it's too much :)

Andarist:
Oh, good to know! Ives is making an impressive job on this thing 😍

PR #1487, opened 2018-06-10
Add root saga docs file.

zgavin1:
#1316 

Root Saga docs.


Andarist:
Thanks for your great work!

zgavin1:
Thank you @Andarist. I'm just starting to do more open source work, thanks for all your feedback and your responsiveness 😄

PR #1488, opened 2018-06-11
internal: using env.runCustomEffects to separate implementations of effect runners

shinima:
**EXPERIMENTAL** implementation of `env.runCustomEffects`. ~~This PR is based on #1477. Implementation of `env.runCustomEffects` is implemented in [this commit](https://github.com/redux-saga/redux-saga/pull/1488/commits/50a86d1fbaff04a5e264a37dfe099bdb3e02157e), other 3 commits are from the base PR.~~

In this PR, I add `runCustomEffects` to `env`. Every time `proc#runEffect` cannot find a core effect-runner for an effect, it will use `env.runCustomEffects` to run this effect. This allows us to define customized effect runners when constructing `env` in `runSaga()`. I adopt a similar way with `options.effectMiddlewares` that composes redux-like middlewares to enhance the `defaultRunCustomEffect`.  

I move the implementations of all/race/select effect runners from **proc.js** into **common-effect-middlewares.js**. And select effect runner is only defined when `options.getState` is a function.

What's passed to  `env.runCustomEffects`  should be carefully considered. Currently runCustomEffects is only used internally, so I passed `type, payload, env, currCb, effectId, digestEffect, taskContext` without in-depth consideration.

How do think about these changes? This experimental PR handles with `all / race / select`. Ideally, channel-related effect types (put / take / flush / actionChannel) could be handled in the same way.

### DIFF
--- a/packages/core/src/internal/common-effect-middlewares.js
+++ b/packages/core/src/internal/common-effect-middlewares.js
@@ -4,10 +4,14 @@ import { isEnd } from './channel'
 
 export const shouldComplete = res => isEnd(res) || shouldTerminate(res) || shouldCancel(res)
 
-export function makeSelectEffectMiddleware(getState) {
+export function createSelectEffectMiddleware(getState) {
   return next => arg => {
     if (arg.type !== effectTypes.SELECT) {
       return next(arg)
+    }
+    if (is.undef(getState)) {
+      arg.cb(new Error(`${effectTypes.SELECT} effect is only available when getState is defined`), true)
+      return
     }
     const { payload, cb } = arg
     const { selector, args } = payload
--- a/packages/core/src/internal/common-effect-middlewares.js
+++ b/packages/core/src/internal/common-effect-middlewares.js
@@ -1,0 +1,124 @@
+import { array, is, noop, shouldCancel, shouldTerminate } from './utils'
+import * as effectTypes from './effectTypes'
+import { isEnd } from './channel'
+
+export const shouldComplete = res => isEnd(res) || shouldTerminate(res) || shouldCancel(res)
+
+export function makeSelectEffectMiddleware(getState) {
+  return next => arg => {
+    if (arg.type !== effectTypes.SELECT) {
+      return next(arg)
+    }
+    const { payload, cb } = arg
+    const { selector, args } = payload
+    try {
+      const state = selector(getState(), ...args)
+      cb(state)
+    } catch (error) {
+      cb(error, true)
+    }
+  }
+}
+
+export function allEffectMiddleware(next) {
+  return arg => {
+    if (arg.type !== effectTypes.ALL) {
+      return next(arg)
+    }
+
+    const { payload: effects, cb, digestEffect, effectId } = arg
+    const keys = Object.keys(effects)
+
+    if (!keys.length) {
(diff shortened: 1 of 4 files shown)


Andarist:
Could u rebase this against the master? Would be cool to see only new changes in the diff for this PR

shinima:
Rebased.

packages/core/src/internal/common-effect-middlewares.js
```diff
@@ -0,0 +1,124 @@
...
+        keys.forEach(key => childCbs[key].cancel())
+      }
+    }
+
+    keys.forEach(key => digestEffect(effects[key], effectId, key, childCbs[key]))
+  }
+}
+
+export function raceEffectMiddleware(next) {
+  return arg => {
```
restrry: what's the benefit of  changing  implementation to middleware ?

shinima: IMHO, `proc()` is over complex now that all of the effect runners are implemented in proc. I think we could separate these effect types into three categories and make the whole implementation more clearer.

* core category: fork / join / cancel / cancelled.  These effects implement the redux-saga fork model and need access `task / forkQueue / mainTask` objects to accomplish their work.
* common category: like all and race. They do not access `task / forkQueue / mainTask` and have less dependencies on saga internals. They could be implemented in a separate file like this PR.
* channel-related category: including put / take / flush and could be implemented in another file too.

From proc's point of view, all effect types except those from the core category could be treated **customized**, so I pick `env.runCustomEffect` as the name and use it to run race/all effects.

Anyway, this is an experimental idea, we can discuss its pros&cons more thoroughly.

shinima:
The biggest disadvantage of this PR is that it introduces complexity. This PR shows how user-defined effect types could be implemented. But if we do not plan to support user-defined effects, I think keep ing the current implementation is not bad.

packages/core/src/internal/common-effect-middlewares.js
restrry: I do like idea to extract effects from proc. just not sure that we should implement all effects via middleware api, as you noticed that adds complexity. 
if we want to provide examples how to write custom effects we can provide them as a part of documentation.

shinima: This is a trade-off between extensibility and complexity. Effect types that redux-saga provides are very sufficient, in my coding practice I could always use built-in effect types to achieve my goals.

Currently we are going to release v1, I think we could ignore this pull request first and come back later.

Andarist:
The idea itself is nice and super thumbs up for experimenting with it 👍

I don't feel that we should use middleware approach for some of current core effects. They all seem quite core to me and while ofc we could remove them from the core I don't see much benefit from it. That being said - refactoring the core so that their implementation could be moved to separate files would be super nice.

Also I'm wondering about using middleware approach at all. Middleware's purpose seems to be more about introducing a transparent logic layers, whereas custom effects would have to have distinct types and thus a simpler effect type to effect runner map sounds like a more appropriate tool for the job.

shinima:
Yes, I agree that the middleware is too complicated for effect-runners. I switch a simpler mechanism using an effect-runner map. See #1524.

shinima:
Closing this PR. See #1524.

PR #1489, opened 2018-06-11
Fixing tests in BeginnerTutorial

jargot:
Tests are not passing the way they are written in the tutorial, so I exported delay() to be able to check deeply against it in tests (deepEqual). Everything passes now.


shinima:
Thanks!

PR #1490, opened 2018-06-11
fix import delay in docs

titulus:
there are mistake in several docs with `delay` util importing
```javascript
import { delay } from 'redux-saga/effects'
```

becouse `redux-saga/effects` does not exports `delay`, but `redux-saga` does itself.
So the right import line is:
```javascript
```

Implement this change for every doc where occurs

### DIFF
--- a/docs/advanced/ForkModel.md
+++ b/docs/advanced/ForkModel.md
@@ -18,7 +18,8 @@ Attached forks remain attached to their parent by the following rules
 For example say we have the following
 
 ```js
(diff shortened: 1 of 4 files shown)


Andarist:
It's already importable from `redux-saga/effects` when using `redux-saga@beta`

Docs in the repository might be already updated to use it from `redux-saga/effects`. Published docs to https://redux-saga.js.org/ shouldnt yet have this change.

PR #1491, opened 2018-06-12
Liftable channels

shinima:
It is just an experimental PR, feel free to reject it.

Currently, we are monkey-patching `channel.put` in several places. One is in **internal/channels.js**:

```javascript
export function stdChannel() {
  const chan = multicastChannel()
  const { put } = chan
  chan.put = input => {
    if (input[SAGA_ACTION]) {
      put(input)
      return
    }
    asap(() => put(input))
  }
  return chan
}
```

Another one is in **internal/middleware.js**:

```javascript
function sagaMiddleware({ getState, dispatch }) {
  const channel = stdChannel()
  channel.put = (options.emitter || identity)(channel.put)
  /* ...... */
}
```

In this PR, I add liftable channels to handle monkey-patching of `channel.put` . Calling `liftable(someMulticastChannel)` will return a liftable version that has a `lift(lifter)` method and a `clone()` method.

* `liftable.lift(lifter)`: lifter is a function that accepts the original *put*, and return a new *put*. The new *put* will replace the original one.
* `liftable.clone()`: lift method mutate channel object in-place. Calling `clone()` will return a cloned liftable, and lifts to the cloned will not affect the original.

I also lift the channel to a connected-channel using `wrapSagaDispatch(dispatch)` if `options.dispatch` is defined, so `env.dispatch` could be removed. 

How do you think about these changes to liftable channels?

### DIFF
--- a/packages/core/src/internal/channel.js
+++ b/packages/core/src/internal/channel.js
@@ -210,15 +210,34 @@ export function multicastChannel() {
   }
 }
 
-export function stdChannel() {
-  const chan = multicastChannel()
-  const { put } = chan
-  chan.put = input => {
-    if (input[SAGA_ACTION]) {
-      put(input)
-      return
+function liftable(chan) {
+  chan.lift = fn => {
+    if (process.env.NODE_ENV === 'development') {
+      check(fn, is.func(fn), 'channel.lift(lifter): lifter must be a function')
     }
+    chan.put = fn(chan.put)
+    if (process.env.NODE_ENV === 'development') {
+      check(fn, is.func(chan.put), 'channel.lift(lifter): lifter must return a function')
+    }
+    return chan
+  }
+
+  chan.clone = () => {
+    const { put, take, close } = chan
+    return liftable({ [MULTICAST]: true, put, take, close })
+  }
+
+  return chan
+}
+
+const scheduleSagaPut = put => input => {
+  if (input[SAGA_ACTION]) {
+    put(input)
+  } else {
     asap(() => put(input))
   }
-  return chan
+}
+
+export function stdChannel() {
+  return liftable(multicastChannel()).lift(scheduleSagaPut)
 }
--- a/packages/core/src/internal/proc.js
+++ b/packages/core/src/internal/proc.js
@@ -432,7 +432,7 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
     proc(env, iterator, taskContext, effectId, meta, cb)
   }
 
-  function runTakeEffect({ channel = env.stdChannel, pattern, maybe }, cb) {
+  function runTakeEffect({ channel = env.channel, pattern, maybe }, cb) {
     const takeCb = input => {
       if (input instanceof Error) {
         cb(input, true)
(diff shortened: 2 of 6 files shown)


packages/core/src/internal/runSaga.js
```diff
@@ -73,9 +73,16 @@ export function runSaga(options, saga, ...args) {
     }
   }
 
+  let connectedChannel = channel
+  if (is.notUndef(dispatch)) {
+    if (process.env.NODE_ENV === 'development') {
+      check(dispatch, is.func, 'dispatch must be a function')
+    }
+    connectedChannel = channel.clone().lift(() => wrapSagaDispatch(dispatch))
```
shinima: Lift the channel to be a connected-channel, so `env.channel` could be removed from `env`.

packages/core/test/runSaga.js
```diff
@@ -65,3 +66,58 @@ test('runSaga', assert => {
...
+test('runSaga with emitter', assert => {
+  assert.plan(1)
+
+  let actual = []
+
+  const em = mitt()
+  let state = null
+  em.on('action', action => (state = action))
+
+  const channel = stdChannel().lift(put => {
```
shinima: Connecting to an emitter is more straightforward than before as all the connecting logic is here and we do not need `dispatch`.

packages/core/src/internal/proc.js
```diff
@@ -462,7 +462,7 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
     asap(() => {
       let result
       try {
-        result = (channel ? channel.put : env.dispatch)(action)
+        result = channel.put(action)
```
shinima: `env.dispatch` is removed here. So running sagas do all the input/output via a channel.

packages/core/src/internal/channel.js
```diff
@@ -210,15 +210,34 @@ export function multicastChannel() {
...
+    chan.put = fn(chan.put)
+    if (process.env.NODE_ENV === 'development') {
+      check(fn, is.func(chan.put), 'channel.lift(lifter): lifter must return a function')
+    }
+    return chan
+  }
+
+  chan.clone = () => {
+    const { put, take, close } = chan
+    return liftable({ [MULTICAST]: true, put, take, close })
```
restrry: should it read `MULTICAST` from `chan`?

packages/core/src/internal/runSaga.js
```diff
@@ -73,9 +73,16 @@ export function runSaga(options, saga, ...args) {
     }
   }
 
+  let connectedChannel = channel
+  if (is.notUndef(dispatch)) {
+    if (process.env.NODE_ENV === 'development') {
+      check(dispatch, is.func, 'dispatch must be a function')
+    }
+    connectedChannel = channel.clone().lift(konst(wrapSagaDispatch(dispatch)))
```
shinima: Lift the channel to be a connected-channel, so `env.channel` could be removed from `env`.

shinima:
I come up with a better idea. I'll make another PR in a few days.

PR #1492, opened 2018-06-13
Update docs to v1 part 1

restrry:
- putResolve
- delay / takeEvery / takeLeading / takeLatest as effects
- debounce


shinima:
Great works 👍

Andarist:
While we are at it - `take.maybe` became `takeMaybe` so it's a good opportunity to rename it too.

restrry:
@Andarist done

Issue #1362, opened 2018-03-11
Wait for a task to be finished with success, error or cancelled


klis87:
I have a following saga:
```js
function* cancelTaskOnAction(actionType, task) {
  const { abortingAction, taskFinished } = yield race({
    abortingAction: take(actionType),
    taskFinished: join(task), // or task.done,
  });

  if (abortingAction) {
    console.log('lets abort this task and exit');
    yield cancel(task);
  } else {
    console.log('task finished or has been cancelled from somewhere else, so there is no need to wait for actionType anymore, lets exit to free memory');
  }
}
```

The problematic part is `taskFinished: join(task)`, it works when task finished either with ok response or an error, but `taskFinished` will never win this race if task was cancelled, for example by `takeLatest` somewhere. The saga will hang on `take(actionType)`, which is not good for memory.  Is this a bug or this is intentional? Is it possible to add an effect to my race like `taskAborted`?

I am using Redux-Saga v0.16.0

Andarist:
joining a cancelled task cancels a joiner (at least it should), does ur `cancelTaskOnAction` really hangs unresolved when `task` gets cancelled? Could u provide a repro on something like codesandbox.io ?

klis87:
It seems so, as after `task` is cancelled, this saga will hang until `take(actionType)` is fired. I will create a demo soon, I hope I will manage it today.

klis87:
I prepared demo: https://codesandbox.io/s/73qo3w8096

It looks like we have 2 problems, when `join` is used in `race`, when task finishes normally, it will win, but `winner` object won't contain winner's key. 2nd problem, when task is cancelled somewhere else, this saga will still run, demo shows 2 cases:
- 1st case - just click `Start new worker`.
- 2nd case - click `Start new worker` and then within 5s `Stop all workers`

Andarist:
1. the winner has `taskStopped` key, it's value just happens to be `undefined` because `worker` doesn't return anything
2. seems like a problem, gonna look into how it is handled internally and what we can do to fix this. Just to clarify - in your eyes `workerAborter` should get cancelled based on regular cancellation propagation  to joiners rule, right?

klis87:
re 1) My bad :)
re 2) When I raised this issue, when `task` is cancelled, I expected `taskFinished` to be the winner. I thought this is more logical, because as mentioned in docs : 
> Cancellation propagates downward

[...]

Issue #1354, opened 2018-03-06
END and the consequences on take effects inside races


chacestew:
Hi there,

How exactly should `take` effects behave inside a `race` when `END` is dispatched?

I am running into an issue on server rendering. If a saga is blocked on the race below...

```js
yield put(doSomethingAsync()) // this action is taken to trigger a fetch saga

const { success, error, timeout } = yield race({
      success: take(doSomethingAsyncSuccess),
      error: take(doSomethingAsyncFailure),
      timeout: delay(5000)
});
```
... and `END` is dispatched, then putting either `doSomethingAsyncSuccess` or `doSomethingAsyncFailure` does nothing and the saga hangs. Only after 5 seconds will the delay resolve to timeout and the saga can continue. 

Am I correct in thinking the race effect is not terminated but any `take` conditions inside it are? If so there a better way to handle such logic when dispatching `END`?

chacestew:
Reading https://github.com/redux-saga/redux-saga/issues/1117 this does seem to confirm the behaviour that **all** basic takes will terminate when `END` is dispatched via the `stdChannel`, so I'lI scrap that question.

Before closing - is there any intention for this behaviour to change in v1, relating to https://github.com/redux-saga/redux-saga/pull/824 or others? If there is anything already in beta to change or extend how `END` works, I'd be keen to test it in our implementation.

Andarist:
You could try to utilize `effectsMiddleware` feature from v1 to track and resolve other effects based on `END`.

Truth to be told `END` is a hacky solution in itself, if you have any proposal on how it could be improved, I'm all ears. What would you see handled differently?

Andarist:
https://github.com/redux-saga/redux-saga/pull/1493 got merged in, this should "resolve" ur issue by aligning the behaviour with `all` - `END` is going to cancel the saga waiting on `take`

PR #1493, opened 2018-06-13
Should finish race effect on END

restrry:
a breaking change. 
race should be finished if any of effects resolved with END (by analogy with `all`)
https://github.com/redux-saga/redux-saga/issues/1362
https://github.com/redux-saga/redux-saga/issues/1354

### DIFF
--- a/packages/core/test/interpreter/race.js
+++ b/packages/core/test/interpreter/race.js
@@ -8,32 +8,27 @@ import * as io from '../../src/effects'
 test('saga race between effects handling', assert => {
   assert.plan(1)
 
-  let actual = []
+  let resultOfRace = 'initial'
   const timeout = deferred()
   const middleware = sagaMiddleware()
   const store = applyMiddleware(middleware)(createStore)(() => {})
 
   function* genFn() {
-    actual.push(
-      yield io.race({
-        event: io.take('action'),
-        timeout: timeout.promise,
-      }),
-    )
+    resultOfRace = yield io.race({
+      event: io.take('action'),
+      timeout: timeout.promise,
+    })
   }
 
   const task = middleware.run(genFn)
 
-  Promise.resolve(1)
+  const expected = { timeout: 1 }
+  Promise.resolve()
     .then(() => timeout.resolve(1))
     .then(() => store.dispatch({ type: 'action' }))
-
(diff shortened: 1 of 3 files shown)


packages/core/test/interpreter/race.js
```diff
@@ -73,33 +73,33 @@ test('saga race between array of effects handling', assert => {
...
-    actual.push(
+    try {
       yield io.race({
         event: io.take('action'),
         timeout: timeout.promise,
-      }),
-    )
+      })
+    } finally {
+      cancelled = true
```
shinima: It seems incorrect to do assignment in finally block. `cancelled = true` will always executes here regardless of the race result.

restrry: oops. forgot to remove delay.

restrry: I aligned the test with happy path version to  emphasise the difference between them
https://github.com/redux-saga/redux-saga/blob/af4b1b41f6f336bb4668fe2be569819a5ca3ee62/packages/core/test/interpreter/race.js#L8

shinima:
Changes looks good to me. But I'm worrying about the cases when we place takeMaybe in race/all.

```js
function* gen() {
   const allResult = yield all([
    takeMaybe('some-action'), // will take an END
    somePromise,
  ])
  expect.deepEqual(allResult, [END, resolvedValueOfPromise])
}
```

When `takeMay` take an END from the channel, END is passed to `all` and then `all` terminates the task. So we could never take an END using takeMay if takeMay is nested in all/race.

I think maybe we should treat END as a normal result in all/race. How do you think ? Point out my mistakes if I'm wrong.

shinima:
I may not explain myself clearly... Actually I'm talking about `takeMaybe` in `all`. In current implementation, users could not use `takeMaybe` to take an END if `takeMaybe` is in `all`.

It my fault that `takeMaybe` is not related to this PR. I'll make another PR when this PR gets merged. This PR itself is very good that it makes `all` and `race` more consistent when dealing with END.

Andarist:
At least it's consistent with `all` now 👍 Thanks you both for great work, I'll try to catch up with other issues/PRs in following days

Andarist:
Damn, I cant remember what cases I've wanted to check 🤦‍♂️

PR #1496, opened 2018-06-16
Liftable stdChannel

shinima:
In this PR, stdChannels become liftable. A liftable channel has three extra methods:

* `liftable.lift(lifter)`: lifter is a function that accepts the original put, and return a new put. The new put will replace the original put of the channel. Method `lift` is used to handle monkey-patching of `channel.put`.
* `liftable._clone()`: `lift(fn)` mutates channel object in-place. Calling ` _clone()` will return a cloned liftable, and lifts to the cloned will not affect the original.
* `liftable._connect(dispatch)`: connect channel to external I/O. This method return a connected-channel. `dispatch` will be used as **the put handler** for the connected-channel.

`env.dispatch` is removed in this PR, and all input/output of running saga is done via  `env.channel`. This makes sagas a little purer (requires less knowledge about the outer world).

## Public changes:

* stdChannel has a new public method `lift`. Users could use this method to lift/enhance a stdChannel.

* `options.emitter` of `sagaMiddlewareFactory` is removed. Instead, `options.channel` is added, users could create a stdChannel and lift it with the emitter to achieve the same goal of the old `options.emitter`. This make `sagaMiddlewareFactory()` and `runSaga()` more consistent in options: `runSaga#options` only has two more fields (dispatch / getState) than `sagaMiddlewareFactory#options`.

### DIFF
--- a/packages/core/src/internal/channel.js
+++ b/packages/core/src/internal/channel.js
@@ -220,7 +220,7 @@ const scheduleSagaPut = put => input => {
 
 const wrapSagaDispatch = dispatch => action => dispatch(Object.defineProperty(action, SAGA_ACTION, { value: true }))
 
-function liftable(chan) {
+function multicastLiftable(chan) {
   chan.lift = fn => {
     if (process.env.NODE_ENV === 'development') {
       check(fn, is.func(fn), 'channel.lift(fn): fn must be a function')
@@ -234,7 +234,7 @@ function liftable(chan) {
 
   chan._clone = () => {
     const { put, take, close } = chan
-    return liftable({ [MULTICAST]: true, put, take, close })
+    return multicastLiftable({ [MULTICAST]: true, put, take, close })
   }
 
   chan._connect = dispatch => chan._clone().lift(() => wrapSagaDispatch(dispatch))
@@ -243,5 +243,5 @@ function liftable(chan) {
 }
 
 export function stdChannel() {
-  return liftable(multicastChannel()).lift(scheduleSagaPut)
+  return multicastLiftable(multicastChannel()).lift(scheduleSagaPut)
 }
--- a/packages/core/src/internal/channel.js
+++ b/packages/core/src/internal/channel.js
@@ -210,15 +210,38 @@ export function multicastChannel() {
   }
 }
 
-export function stdChannel() {
-  const chan = multicastChannel()
-  const { put } = chan
-  chan.put = input => {
-    if (input[SAGA_ACTION]) {
-      put(input)
-      return
-    }
+const scheduleSagaPut = put => input => {
+  if (input[SAGA_ACTION]) {
+    put(input)
+  } else {
     asap(() => put(input))
   }
+}
+
+const wrapSagaDispatch = dispatch => action => dispatch(Object.defineProperty(action, SAGA_ACTION, { value: true }))
+
+function liftable(chan) {
+  chan.lift = fn => {
+    if (process.env.NODE_ENV === 'development') {
+      check(fn, is.func(fn), 'channel.lift(fn): fn must be a function')
+    }
+    chan.put = fn(chan.put)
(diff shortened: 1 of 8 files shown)


packages/core/test/middleware.js
```diff
@@ -95,22 +95,22 @@ test('middleware options', assert => {
...
-    },
-  })
+  const emitter = emit => action => {
+    if (action.type === 'batch') {
+      action.batch.forEach(emit)
+      return
+    }
+    emit(action)
+  }
+  const channel = stdChannel().lift(emitter)
```
shinima: This shows how a customized emitter could be applied.

packages/core/src/internal/channel.js
```diff
@@ -210,15 +210,38 @@ export function multicastChannel() {
...
+    chan.put = fn(chan.put)
+    if (process.env.NODE_ENV === 'development') {
+      check(fn, is.func(chan.put), 'channel.lift(fn): fn must return a function')
+    }
+    return chan
+  }
+
+  chan._clone = () => {
+    const { put, take, close } = chan
+    return liftable({ [MULTICAST]: true, put, take, close })
```
restrry: why using `clone` assume that we're using multicast channel?

shinima: Only stdChannels have `lift / _clone / _connect` methods.  These methods are only used by stdChannel to monkey-patching its `put` and handle the scheduling of external IO.
Multicast channels or normal channels do not need to do such monkey-patching or scheduling, so they do not have these methods.

restrry: I see. but the current implementation doesn't tell us the only stdChannel could be passed in.

shinima: I agree this is not obvious that only stdChannel could be passed to `liftable`.  But `liftable` is internal (only `lift` method on stdChannel instance is public), so I think it's just fine.

restrry: we have other `channels` defined in this file. probably we can just rename method to makeLiftableStdChannel

restrry:
ok for me. @Andarist ?

packages/core/test/middleware.js
Andarist: this looks quite nice

Andarist:
Do u have any resources about the lifting concept explained or with just with other examples of this technique? Would like to understand the idea and how it's used better before moving with this forward.

shinima:
Not much resources/examples...  I found we were monkey-patching stdChannel.put somehow, so I just came up with `lift` as a unified way to monkey-patch stdChannel.put.

Here `lift` is just a simple mechanism like reducer-enhancer in redux, accepts the old *put* and returns the new *put*.  By lifting the stdChannel, we can move some logic/complexity into the `env.channel`, which makes `sagaMiddlewareFactory` more consistent with `runSaga`.

restrry:
I only read about `lifting` concept from Haskell, but seems that's a bit different (unwrap value, apply function, wrap). 
https://stackoverflow.com/questions/2395697/what-is-lifting-in-haskell
in js it similar to something to
```js
const boxA = {value: 1}
const boxB = {value: 2}
const lift = (f, ...args) => ({
   value: f(...args.map(i => i.value))
})
const sum = (x, y) => x + y

const sumBox = lift (sum, boxA, boxB)
```

shinima:
`lift` in [ramda](https://ramdajs.com/docs/#lift) (a popular javascript functional programming library) is much like lifting in Haskell.   `lift` in RxJS also has a similar meaning too.

In this PR, `lift` has different meaning and may not be a good name. What about change it to `enhance` or `enhancePut`?  I'm quite insensitive between lift and enhance, they have the same meaning when translated to Chinese.

restrry:
👍 for `enhancePut`

shinima:
See #1528.

PR #1506, opened 2018-06-27
Remove isEnd check in shouldComplete

shinima:
`END` is always handled by `runTakeEffect`,  so `isEnd` check is unnecessary in `shouldComplete`.

@restrry  This PR is based on #1493.


restrry:
agree

PR #1517, opened 2018-07-03
Add test for passing thrown error in onError

restrry:


packages/core/test/utils/createMockTask.js
```diff
@@ -0,0 +1,59 @@
...
+      assert.equal(
+        task.isRunning(),
+        false,
+        'Sagas must take consecutive actions dispatched synchronously on an action channel even if it performs blocking calls',
+      )
+      assert.end()
+    })
+    .catch(err => assert.fail(err))
+
+  // const generator = saga()
```
shinima: Could you delete these comments?

restrry: oops

packages/core/test/interpreter/onerror.js
```diff
@@ -26,6 +26,52 @@ test('saga onError is optional', assert => {
   })
 })
 
+test('saga passes thrown Error instance in onError handler', assert => {
```
Andarist: isnt this test a duplicate of "saga onError is called for uncaught error"?

restrry: didn't notice 👍 
changed test case

PR #1521, opened 2018-07-05
Setup eslint for react

shinima:
I found that eslint fails to run in examples folder. So I install `eslint-plugin-react` and eslint now will lint the examples folder too. I'll update the code in example folder gradually.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,5 +1,5 @@
 {
-  "extends": "eslint:recommended",
+  "extends": ["eslint:recommended", "plugin:react/recommended"],
   "env": {
     "browser": true,
     "node": true,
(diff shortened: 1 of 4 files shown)


.eslintrc.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "eslint:recommended",
+  "extends": ["eslint:recommended", "plugin:react/recommended"],
```
restrry: should we add `"plugin:react/recommended"` only for examples folder?

shinima: Agree. There are no react code in packages folder.

restrry:
ok

shinima:
Rebased this upon master.

PR #1522, opened 2018-07-05
Move prettier config to .prettierrc.json

shinima:
Move prettier config to root package.json. With better integration with WebStorm and VSCode, we can use prettier to format the code without running lint-staged.


Andarist:
While the change is good I would prefer creating a dedicated config file for this - such as `.prettierrc`. IMHO it's better to keep those configs in separate files - there are so many tools we already have and there might be more in the future that `package.json` can quickly become a dumpster for configs.

shinima:
Agree. I've moved the config to a separate file.

PR #1523, opened 2018-07-07
Upgrade babel version

shinima:
Upgrade @babel/xxx to version 7.0.0-beta.52. 

Rename **.babelrc.js** to **babel.config.js** which is supported in babel/babel#7358 (docs of babel.config.js can be found [here](https://babeljs.io/docs/en/next/babelconfigjs))

Add `{ decoratorsLegacy: true }` which is required after  7.0.0-beta.45 ( babel/babel#7734 ).


shinima:
I don't why tests with babel6 and `options: { presets: ['env'] }` starts to fail.  Babel6 adds `'use strict'` as babel7 does when `{ presets: ['env'] }` is present.  

Is it a desired behavior? Can you help me to fix this? @restrry

Andarist:
What issue is supposed to be fixed by `.babelrc.js` -> `babel.config.js` rename?

examples/async/babel.config.js
```diff
@@ -14,9 +13,7 @@ module.exports = {
       },
     ],
     '@babel/react',
-    '@babel/stage-2',
+    ['@babel/stage-2', { decoratorsLegacy: true }],
```
Andarist: do we even use decorators anywhere?

packages/core/package.json
```diff
@@ -59,13 +59,13 @@
...
-    "@babel/preset-react": "7.0.0-beta.39",
-    "@babel/preset-stage-2": "7.0.0-beta.39",
+    "@babel/cli": "7.0.0-beta.52",
+    "@babel/core": "7.0.0-beta.52",
+    "@babel/node": "7.0.0-beta.52",
+    "@babel/polyfill": "7.0.0-beta.52",
+    "@babel/preset-env": "7.0.0-beta.52",
+    "@babel/preset-react": "7.0.0-beta.52",
+    "@babel/preset-stage-2": "7.0.0-beta.52",
     "babel-plugin-annotate-pure-calls": "babel7",
```
Andarist: we could update this to `^0.3.0`

Andarist:
`"use strict"` is probably inserted because those files are considered to be modules (rather than scripts) and when transforming them to CJS this directive has to be added to achieve strict semantics

I suppose failing tests are just a matter of updating the fixtures, but you should run tests locally and check each one individually - if diff looks OK then just accept new one.

shinima:
The babel's doc [says](We recommend to use the babel.config.js format. Babel itself is using it.):
> We recommend to use the babel.config.js format. Babel itself is using it.

I personally prefer babel.config.js as it aligns with webpack.config.js and rollup.config.js.  Another minor advantage is that ESLint will check babel.config.js now ( ESLint does not check .babelrc.js before ).


> I suppose failing tests are just a matter of updating the fixtures, but you should run tests locally and check each one individually - if diff looks OK then just accept new one.

I am using Windows and tests of babel-plugin-redux-saga always fails due to path separators ( Windows uses `\` instead of `/` ) .  😂 I'll switch to a mac and see the diffs carefully.

shinima:
I've updated babel-plugin-annotate-pure-calls to ^0.3.0, edited fixtures and passed the tests. @Andarist

examples/async/babel.config.js
Andarist: We should remove `stage-2` preset anyway and use specific transforms if needed. I suspect the only thing we really use is `@babel/proposal-object-rest-spread`

shinima:
I have removed `@babel/preset-stage-2` and added `@babel/proposal-object-rest-spread` instead.   babel.config.js got updated too.   @Andarist

examples/async/babel.config.js
```diff
@@ -5,15 +5,16 @@ const cjs = BABEL_ENV === 'cjs' || NODE_ENV === 'test'
 module.exports = {
   presets: [
     [
-      '@babel/env',
+      '@babel/preset-env',
```
shinima: I've updated to use full package name of `@babel/xxx`.

restrry:
ok for me

shinima:
I am merging this PR now since I have other PRs depending on this.  Do not forget to run `npm install` and `lerna bootstrap` after this PR.

examples/async/babel.config.js
```diff
@@ -5,15 +5,16 @@ const cjs = BABEL_ENV === 'cjs' || NODE_ENV === 'test'
...
         modules: false,
         forceAllTransforms: true,
       },
     ],
-    '@babel/react',
-    ['@babel/stage-2', { decoratorsLegacy: true }],
+    '@babel/preset-react',
   ],
-  plugins: [cjs && '@babel/transform-modules-commonjs'].filter(Boolean),
+  plugins: [cjs && '@babel/plugin-transform-modules-commonjs', '@babel/plugin-proposal-object-rest-spread'].filter(
```
Andarist: isnt object-rest-spread included in preset-env? (not sure if its on by default or not though)

also - it accepts `loose` flag too and we should use it

great change about removing `stage-2` 👌

shinima: `object-rest-spread` is still in [stage-3](https://babeljs.io/docs/en/babel-preset-stage-3), so we need to keep it in the plugin list.

I'll add `loose` flag. And just out of curiosity, what's the benefits of this flag?

PR #1524, opened 2018-07-08
EXPERIMENTAL Introduce env.effectRunnerMap and support for custom effects

### LINKED ISSUE #1238: Custom effects
Just look into what it would take to add such functionality

shinima:
| Q                       | A                                           |
| ----------------------- | ------------------------------------------- |
| Fixed Issues?           | May fix #1238                               |
| Patch: Bug Fix?         | None                                        |
| Major: Breaking Change? | None                                        |
| Minor: New Feature?     | Support custom/user-defined effect          |
| Tests Added + Pass?     | Need to add new tests / Pass existing tests |
| Documentation           | Need to add documentation                   |
| Any Dependency Changes? |  None                                           |

**EXPERIMENTAL** implementation of `env.effectRunnerMap`. This PR is big since I separate implementations of core-effect-runners into different files. But the idea of this PR is rather simple.

In this PR, effect-runners have a unified function interface which can be seen in the new implementation of `runEffect` (the interface may need further discussion and refinement):

```javascript
    if (is.promise(effect)) {
      resolveIterator(effect, effectId, meta, currCb)
      const effectRunningContext = {
        effectId,
        task,
        taskContext,
        mainTask,
        taskQueue,
        digestEffect,
        resolvePromise,
        resolveIterator,
      }

      const effectRunner = env.effectRunnerMap[effect.type]

      if (is.notUndef(effectRunner)) {
        // Unified effect-runner function interface
        effectRunner(env, effect.payload, currCb, effectRunningContext)
      } else {
        currCb(new Error(`${effect.type} is not a valid effect type`), true)
      }
    } else {
      // anything else returned as is
      currCb(effect)
    }
  }
```

`env.effectRunnerMap` is an object that maps effect-type to effect-runner. It is initialized in `runSaga()` by combining `coreEffectRunnerMap` and `options.customEffectRunnerMap`. 

`coreEffectRunnerMap` is the runner-map that contains all the existing core effect-runners. I've moved the core effect-runners into folder *src/internal/coreEffectRunners/*, one file per runner.

Users could provide `options.customEffectRunnerMap` to define custom effect types. But note that the core effect types have higher priority over custom effect types.

### DIFF
--- a/packages/core/src/internal/coreEffectRunners/cancelled.js
+++ b/packages/core/src/internal/coreEffectRunners/cancelled.js
@@ -1,0 +1,3 @@
+export default function runCancelledEffect(env, effectPayload, cb, { mainTask }) {
+  cb(Boolean(mainTask._isCancelled))
+}
--- a/packages/core/src/internal/coreEffectRunners/cancelled.js
+++ b/packages/core/src/internal/coreEffectRunners/cancelled.js
@@ -1,3 +1,3 @@
-export default function runCancelledEffect(env, effectPayload, cb, { mainTask }) {
-  cb(Boolean(mainTask._isCancelled))
+export default function runCancelledEffect(env, payload, cb, { mainTask }) {
+  cb(mainTask._isCancelled)
 }
--- a/packages/core/src/internal/coreEffectRunners/select.js
+++ b/packages/core/src/internal/coreEffectRunners/select.js
@@ -1,13 +1,12 @@
 import * as effectTypes from '../effectTypes'
 import { is } from '../utils'
 
-export default function select(env, payload, cb) {
+export default function select(env, { selector, args }, cb) {
   if (is.undef(env.getState)) {
     cb(new Error(`${effectTypes.SELECT} effect is only available when getState is defined`), true)
     return
   }
 
-  const { selector, args } = payload
   try {
     const state = selector(env.getState(), ...args)
     cb(state)
--- a/packages/core/src/internal/coreEffectRunners/select.js
+++ b/packages/core/src/internal/coreEffectRunners/select.js
@@ -1,0 +1,17 @@
+import * as effectTypes from '../effectTypes'
+import { is } from '../utils'
+
+export default function select(env, payload, cb) {
+  if (is.undef(env.getState)) {
+    cb(new Error(`${effectTypes.SELECT} effect is only available when getState is defined`), true)
+    return
+  }
+
+  const { selector, args } = payload
+  try {
+    const state = selector(env.getState(), ...args)
+    cb(state)
+  } catch (error) {
+    cb(error, true)
+  }
+}
--- a/packages/core/src/internal/proc.js
+++ b/packages/core/src/internal/proc.js
@@ -1,41 +1,17 @@
-import { CANCEL, IO, TERMINATE, TASK, TASK_CANCEL, SELF_CANCELLATION } from './symbols'
-import * as effectTypes from './effectTypes'
+import { CANCEL, IO, TASK, TASK_CANCEL } from './symbols'
+import { addSagaStack, sagaStackToString } from './error-utils'
 import {
   noop,
   is,
   check,
   deferred,
   uid as nextEffectId,
-  array,
   remove,
   object,
-  makeIterator,
   createSetContextWarning,
+  shouldCancel,
+  shouldTerminate,
 } from './utils'
-
-import { getLocation, addSagaStack, sagaStackToString } from './error-utils'
-
-import { asap, suspend, flush } from './scheduler'
-import { channel, isEnd } from './channel'
-import matcher from './matcher'
-
-export function getMetaInfo(fn) {
-  return {
-    name: fn.name || 'anonymous',
-    location: getLocation(fn),
-  }
-}
-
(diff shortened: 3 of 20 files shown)


shinima:
If we do not plan to support custom effects in v1, this PR itself is good because it separates the implementation of existing core effect runners.

packages/core/src/internal/proc.js
```diff
@@ -335,23 +274,24 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
...
-      else if (type === effectTypes.JOIN) runJoinEffect(payload, currCb)
-      else if (type === effectTypes.CANCEL) runCancelEffect(payload, currCb)
-      else if (type === effectTypes.SELECT) runSelectEffect(payload, currCb)
-      else if (type === effectTypes.ACTION_CHANNEL) runChannelEffect(payload, currCb)
-      else if (type === effectTypes.FLUSH) runFlushEffect(payload, currCb)
-      else if (type === effectTypes.CANCELLED) runCancelledEffect(payload, currCb)
-      else if (type === effectTypes.GET_CONTEXT) runGetContextEffect(payload, currCb)
-      else if (type === effectTypes.SET_CONTEXT) runSetContextEffect(payload, currCb)
-      else currCb(effect)
+      const effectRunningContext = {
```
shinima: Maybe `sagaState` is a better name over `effectRunningContext`.    #1263  uses `sagaState`.

packages/core/src/internal/coreEffectRunners/cancelled.js
```diff
@@ -0,0 +1,3 @@
+export default function runCancelledEffect(env, effectPayload, cb, { mainTask }) {
+  cb(Boolean(mainTask._isCancelled))
```
restrry: no needs to cast to Boolean, it always Boolean https://github.com/redux-saga/redux-saga/blob/master/packages/core/src/internal/proc.js#L169

packages/core/src/internal/runSaga.js
```diff
@@ -73,6 +76,11 @@ export function runSaga(options, saga, ...args) {
     }
   }
 
+  const effectRunnerMap = {}
+  object.assign(effectRunnerMap, customEffectRunnerMap)
+  // core-effect-runners have higher priority over custom-effect-runners
+  object.assign(effectRunnerMap, coreEffectRunnerMap)
```
restrry: it still can be changed in user land, since `effectRunnerMap` is available via `env`. should we freeze object?

packages/core/src/internal/coreEffectRunners/select.js
```diff
@@ -0,0 +1,17 @@
+import * as effectTypes from '../effectTypes'
+import { is } from '../utils'
+
+export default function select(env, payload, cb) {
```
restrry: should we unify signatures across effects? sometimes we destructure it, sometimes use as is - it not obvious that all of them have the same signature

packages/core/src/internal/proc.js
```diff
@@ -335,23 +274,24 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
...
+        taskQueue,
+        digestEffect,
+        resolvePromise,
+        resolveIterator,
+      }
+
+      const effectRunner = env.effectRunnerMap[effect.type]
+
+      if (is.notUndef(effectRunner)) {
+        effectRunner(env, effect.payload, currCb, effectRunningContext)
```
restrry: have gut feeling that we need to change the order. `callback` is the last argument across the whole codebase

packages/core/src/internal/coreEffectRunners/cancelled.js
shinima: got it.

packages/core/src/internal/runSaga.js
shinima: I think it is enough that we document `env` should never be mutated.  Custom effects are advanced usages of redux-saga (if we plan to support custom effects), and users must take care of their effect runners.

packages/core/src/internal/coreEffectRunners/select.js
shinima: agree

packages/core/src/internal/proc.js
shinima: Yes, the effect runner interface need further discussion and refinement. I'm not sure what's the best interface at the moment.

Andarist:
I like the overall refactor that has been done here - extracting effect runners to separate files (out of the `proc` closure) is a really good thing. 

I'm not confident enough in exposing custom effects runner right now, but obviously those changes would help in possibly making this in the future - I would leave custom effects out of the scope of this PR.

At the moment `redux-saga` sits nicely under 7kb and ideally we'd like to keep the size under this threshold - I suspect though that the reported size here is inaccurate right now because this branch is quite old and that actual size is lower (i suspect though it's over 7kb, but I might be wrong).

This is an internal refactor, so it shouldn't block v1 release. Could you rebase this PR to see how things look then (especially regarding the library's size)?

shinima:
I'll try to rebase this in the following days.

shinima:
Close in favor of #1659.

PR #1526, opened 2018-07-10
Update examples (upgrade react/webpack version)

shinima:
| Q                       | A         |
| ----------------------- | --------- |
| Fixed Issues?           | None      |
| Patch: Bug Fix?         | None      |
| Major: Breaking Change? | None      |
| Minor: New Feature?     | None      |
| Tests Added + Pass?     | Passed    |
| Documentation           |           |
| Any Dependency Changes? | See below |

In this PR, I update react and webpack versions in our examples. I simplify some webpack config files by using webpack-dev-server and html-webpack-plugin. And I fix up most of the errors reported by ESLint in the examples. (Since lint-staged is configured to check *examples/* in #1521).

## Update list:

* UPDATE react and react-dom  -->  ^16.4.1
* UPDATE react-redux  -->  ^5.0.7
* UPDATE react-router  -->  ^3.2.1
* UPDATE webpack  -->  ^4.15.1
* ADD prop-types, because react16 does not support `React.PropTypes` anymore
* ADD webpack-cli, which is required if we use webpack4
* ADD webpack-dev-server and html-webpack-plugin, which results much simpler webpack configs

* ~~Use .jsx file extension when using JSX in that file~~
* Simplify webpack config

## About *example/real-world*

This example is big and contains more code than other examples. I have not fix all the ESLint errors in *example/real-world* for now, so I insert some `eslint-disable` to pass the lint-stage check. I will make another PR to update this example.

---

All examples run correctly in my PC, but it is better to test the examples again in your computers (remember to install the new dependencies 😉 )


examples/async/src/components/Picker.jsx
```diff
@@ -1,6 +1,7 @@
-import React, { Component, PropTypes } from 'react'
+import React from 'react'
```
Andarist: correct (typing wise, and probably more future-proof) is `import * as React from 'react'`, both versions work now, because react is packages solely as CJS package, but this might change in the future

examples/error-demo/src/components/ErrorGenerator.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
```
Andarist: I would prefer using standard `.js`, I don't see much value in `.jsx` extension as JSX is handled just fine in regular `.js` files by all the tools.

package.json
```diff
@@ -45,7 +45,7 @@
     "@babel/cli": "7.0.0-beta.52",
     "@babel/core": "7.0.0-beta.52",
     "@babel/node": "7.0.0-beta.52",
-    "@babel/plugin-proposal-object-rest-spread": "^7.0.0-beta.52",
+    "@babel/plugin-proposal-object-rest-spread": "7.0.0-beta.52",
```
Andarist: 👌

Andarist:
Just some nitpicking around import style & file extension. Other than that - ready to merge.

shinima:
I've added `lerna bootstrap` to postinstall. No longer need to run `lerna bootstrap` manually.

Issue #1515, opened 2018-07-02
[refactor] join(...tasks) & cancel(...tasks) should return join/cancel descriptors


Andarist:
At the moment they are internally wrapped in `all` but actually the effect type should get preserved so internal refactor is needed.

At the same time we might consider changing those signatures to:
```js
join([...tasks])
cancel([...tasks])
```
but I'm still on the fence with that, not sure which one is better - variadic arguments or a single array argument? any fresh thoughts on this one @restrry @shinima ?

shinima:
I'm inclining to single array argument.

The problem with variadic arguments I found is that `join(singleTask)` matches both `join(...tasks)` and `join(task)`, and it is unclear that `yield join(singleTask)` will return `taskResult` or `[ taskResult ]`.

Andarist:
You have convinced me to array argument with simplicity of input arg type to output arg type mapping 👍

restrry:
@shinima would you like to work on it?

shinima:
OK. I'll get it done in a few days.

shinima:
Done in #1534

PR #1527, opened 2018-07-11
Refactor join and cancel (#1515)

shinima:
| Q                       | A         |
| ----------------------- | --------- |
| Fixed Issues?           | #1515     |
| Patch: Bug Fix?         |           |
| Major: Breaking Change? | See below |
| Minor: New Feature?     |           |
| Tests Added + Pass?     | Passed    |
| Documentation           | Updated   |
| Any Dependency Changes? |           |

In this PR, I change the signatures of join/cancel from variadic arguments to a single array argument. This PR ensures that calling `cancel(...)` returns a cancel-effect (before it may return an all-effect), and ensures that calling `join(...)` returns a join-effect.

Deprecation warnings are added to make it easier to migrate to new API.  A small change in *internal/channel.js* is also included in this PR 😄 .

### DIFF
--- a/packages/core/src/internal/channel.js
+++ b/packages/core/src/internal/channel.js
@@ -40,11 +40,10 @@ export function channel(buffer = buffers.expanding()) {
     if (closed) {
       return
     }
-    if (!takers.length) {
+    if (takers.length === 0) {
       return buffer.put(input)
     }
-    const cb = takers[0]
-    takers.splice(0, 1)
+    const cb = takers.shift()
     cb(input)
   }
 
--- a/packages/core/src/internal/io.js
+++ b/packages/core/src/internal/io.js
@@ -151,7 +151,7 @@ export function cancel(taskOrTasks) {
   if (process.env.NODE_ENV === 'development') {
     if (is.array(taskOrTasks)) {
       taskOrTasks.forEach(t => {
-        check(t, is.task, `join([...tasks]): argument ${t} is not a valid Task object ${TEST_HINT}`)
+        check(t, is.task, `cancel([...tasks]): argument ${t} is not a valid Task object ${TEST_HINT}`)
       })
     } else if (is.notUndef(taskOrTasks)) {
       check(taskOrTasks, is.task, `cancel(task): argument ${taskOrTasks} is not a valid Task object ${TEST_HINT}`)
--- a/packages/core/src/internal/io.js
+++ b/packages/core/src/internal/io.js
@@ -129,7 +129,6 @@ export function join(taskOrTasks) {
         check(t, is.task, `join([...tasks]): argument ${t} is not a valid Task object ${TEST_HINT}`)
       })
     } else {
-      check(taskOrTasks, is.notUndef, 'join(task): argument task is undefined')
       check(taskOrTasks, is.task, `join(task): argument ${taskOrTasks} is not a valid Task object ${TEST_HINT}`)
     }
   }
@@ -141,7 +140,7 @@ const printCancelDeprecationWarning = once(() =>
   log('warn', '`cancel(...tasks)` is deprecated. Please use `cancel([...tasks])` to cancel multiple tasks.'),
 )
 
-export function cancel(taskOrTasks) {
+export function cancel(taskOrTasks = SELF_CANCELLATION) {
   if (arguments.length > 1) {
     if (process.env.NODE_ENV === 'development') {
       printCancelDeprecationWarning()
@@ -153,13 +152,12 @@ export function cancel(taskOrTasks) {
       taskOrTasks.forEach(t => {
         check(t, is.task, `cancel([...tasks]): argument ${t} is not a valid Task object ${TEST_HINT}`)
       })
-    } else if (is.notUndef(taskOrTasks)) {
+    } else if (taskOrTasks !== SELF_CANCELLATION && is.notUndef(taskOrTasks)) {
       check(taskOrTasks, is.task, `cancel(task): argument ${taskOrTasks} is not a valid Task object ${TEST_HINT}`)
     }
-    // else taskOrTasks is undefined, which means self cancellation
   }
 
-  return makeEffect(effectTypes.CANCEL, taskOrTasks || SELF_CANCELLATION)
+  return makeEffect(effectTypes.CANCEL, taskOrTasks)
 }
 
 export function select(selector = identity, ...args) {
--- a/packages/core/src/internal/io.js
+++ b/packages/core/src/internal/io.js
@@ -127,19 +127,19 @@ export function join(taskOrTasks) {
   return makeEffect(effectTypes.JOIN, taskOrTasks)
 }
 
-export function cancel(...tasks) {
-  if (tasks.length > 1) {
-    return all(tasks.map(t => cancel(t)))
+export function cancel(taskOrTasks) {
+  if (process.env.NODE_ENV === 'development') {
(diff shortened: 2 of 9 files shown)


packages/core/src/internal/io.js
```diff
@@ -112,34 +112,54 @@ export function spawn(fn, ...args) {
...
+  }
   if (process.env.NODE_ENV === 'development') {
-    check(task, is.notUndef, 'join(task): argument task is undefined')
-    check(task, is.task, `join(task): argument ${task} is not a valid Task object ${TEST_HINT}`)
+    if (is.array(taskOrTasks)) {
+      taskOrTasks.forEach(t => {
+        check(t, is.task, `join([...tasks]): argument ${t} is not a valid Task object ${TEST_HINT}`)
+      })
+    } else {
+      check(taskOrTasks, is.notUndef, 'join(task): argument task is undefined')
```
restrry: unnecessary check, the second check covers that case

packages/core/src/internal/io.js
```diff
@@ -112,34 +112,54 @@ export function spawn(fn, ...args) {
...
+  }
+  if (process.env.NODE_ENV === 'development') {
+    if (is.array(taskOrTasks)) {
+      taskOrTasks.forEach(t => {
+        check(t, is.task, `cancel([...tasks]): argument ${t} is not a valid Task object ${TEST_HINT}`)
+      })
+    } else if (is.notUndef(taskOrTasks)) {
+      check(taskOrTasks, is.task, `cancel(task): argument ${taskOrTasks} is not a valid Task object ${TEST_HINT}`)
+    }
+    // else taskOrTasks is undefined, which means self cancellation
```
restrry: if we declare `cancel` as 
```js
function cancel(taskOrTasks = SELF_CANCELLATION) {
```
that comment is unnecessary

packages/core/src/internal/proc.js
```diff
@@ -534,73 +534,63 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
...
-  function runJoinEffect(t, cb) {
-    if (t.isRunning()) {
+  function runJoinEffect(taskOrTasks, cb) {
+    if (is.array(taskOrTasks)) {
+      if (taskOrTasks.length === 0) {
+        cb([])
+        return
+      }
+
+      const childCallbacks = createAllStyleChildCallbacks(taskOrTasks, cb)
```
restrry: `createAllStyleChildCallbacks` won't work on empty array or object. should we add such a check there?

packages/core/test/typescript/effects.ts
```diff
@@ -520,11 +520,11 @@ function* testJoin(): SagaIterator {
   yield join({});
 
   yield join(task);
-  yield join(task, task);
-  yield join(task, task, task);
```
restrry: you can keep that one
```js
// typings:expect-error
yield join(task, task);
```

packages/core/src/internal/io.js
shinima: ok

packages/core/src/internal/io.js
shinima: Yes.   I just copied the old code.

packages/core/src/internal/proc.js
shinima: Agree.  Though `createAllStyleChildCallbacks` is internal only, it makes sense to add such a check.

packages/core/test/typescript/effects.ts
shinima: 👍

shinima:
@restrry  Updated the code according to your reviews.

packages/core/src/internal/io.js
```diff
@@ -153,13 +152,12 @@ export function cancel(taskOrTasks) {
       taskOrTasks.forEach(t => {
         check(t, is.task, `cancel([...tasks]): argument ${t} is not a valid Task object ${TEST_HINT}`)
       })
-    } else if (is.notUndef(taskOrTasks)) {
+    } else if (taskOrTasks !== SELF_CANCELLATION && is.notUndef(taskOrTasks)) {
```
restrry: it cannot be `undefined`, since there is a default parameter for `taskOrTasks`

packages/core/src/internal/channel.js
```diff
@@ -40,11 +40,10 @@ export function channel(buffer = buffers.expanding()) {
     if (closed) {
       return
     }
-    if (!takers.length) {
+    if (takers.length === 0) {
       return buffer.put(input)
     }
-    const cb = takers[0]
-    takers.splice(0, 1)
+    const cb = takers.shift()
```
Andarist: 👌

packages/core/src/internal/io.js
```diff
@@ -112,34 +112,52 @@ export function spawn(fn, ...args) {
...
-    return all(tasks.map(t => join(t)))
-  }
-
-  const task = tasks[0]
+const printJoinDeprecationWarning = once(() =>
+  log('warn', '`join(...tasks)` is deprecated. Please use `join([...tasks])` to join multiple tasks.'),
+)
 
+export function join(taskOrTasks) {
+  if (arguments.length > 1) {
```
Andarist: IMHO we are at perfect place for making breaking changes, let's just drop support for `join(...tasks)` & `cancel(...tasks)`

contextual breaking change warning is nice though! we could move that env check into `if (arguments.length > 1)` & drop `taskOrTasks = array.from(arguments)` (and ofc change the deprecation warning to throw error so the developers can quickly fix their call sites)

packages/core/src/internal/proc.js
```diff
@@ -534,73 +534,63 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
...
-      taskToCancel = task
+  function runCancelEffect(taskOrTasks, cb) {
+    if (taskOrTasks === SELF_CANCELLATION) {
+      cancelSingleTask(task)
+    } else if (is.array(taskOrTasks)) {
+      taskOrTasks.forEach(cancelSingleTask)
+    } else {
+      cancelSingleTask(taskOrTasks)
     }
+    cb()
```
Andarist: this got me thinking - what happens when we cancel a task? it calls `it.return()`, right? what happens if `finally` block throws? is that case handled? 🤔

shinima: *cancelled* or *aborted* are two mutual-exclusive state for a task. So if we cancel a task (task's state becomes *cancelled*), all the subsequent errors from the task iterator should not bubble to the parent (or else task's state becomes *aborted*).

Code for handing error during cancellation is [here](https://github.com/redux-saga/redux-saga/blob/29fbd1b50d7128d3b9e4ff4105ae3b84ed2a2905/packages/core/src/internal/proc.js#L274-L276) as follows:
```js
// function next(arg, isErr)
try {
  // other code
} catch (error) {
  if (mainTask._isCancelled) {
    env.logError(error) // currently, we simply log the error if the task is already cancelled
  }
  // maybe we could add `else` to wrap the following two lines of code
  mainTask._isRunning = false
  mainTask.cont(error, true) // if `mainTask._isCancelled` true, here `mainTask.cont` is noop
}
```

packages/core/src/internal/io.js
shinima: 👍

packages/core/src/internal/proc.js
Andarist: so it's handled - thanks for checking!

PR #1528, opened 2018-07-11
Convert stdChannel to an enhanceable version of multicast channel.

shinima:
| Major: Breaking Change? | See below      |
| Minor: New Feature?     |    👍         |
| Tests Added + Pass?     | 👍  |
| Documentation           | Need to update |

In this PR, stdChannels become enhanceable version of multicast channels. A enhanceable channel has three extra methods (one public, and two private/internal):

- `enhanceable.enhancePut(fn)`: fn is a function that accepts the original put, and return a new put. The new put will replace the original put of the channel. Method `enhancePut` is used to handle monkey-patching of `channel.put`.
- `enhanceable._clone()`: `enhancePut(fn)` mutates channel object in-place. Calling `_clone()` will return a cloned enhanceable, and enhancement to the cloned will not affect the original.
- `enhanceable._connect(dispatch)`: connect channel to external I/O. This method returns a connected-channel. `dispatch` is used as **the put handler** for the connected-channel.

`env.dispatch` is removed in this PR, and all input/output of running saga is done via `env.channel`. This makes sagas a little purer (requires less knowledge about the outer world).

Advantages that the PR brings:

* a unified way of changing channle.put
* options are more consistent between `sagaMiddlewareFactory` and `runSaga`

## Public changes:

- stdChannel instances have a new public method `enhancePut`. Users could use this method to enhance `channel.put`.
- **BREAKING CHANGE**  `options.emitter` of `sagaMiddlewareFactory` is removed. Instead, `options.channel` is added, users could create a stdChannel and enhance it with the emitter to achieve the same goal of the old `options.emitter`. This make `sagaMiddlewareFactory()` and `runSaga()` more consistent in options: `runSaga#options` only has two more fields (i.e. dispatch and getState) than `sagaMiddlewareFactory#options`.

### DIFF
--- a/packages/core/src/internal/middleware.js
+++ b/packages/core/src/internal/middleware.js
@@ -1,11 +1,10 @@
 import * as is from '@redux-saga/is'
 import { check, assignWithSymbols, createSetContextWarning } from './utils'
 import { stdChannel } from './channel'
-import { identity } from './utils'
 import { runSaga } from './runSaga'
 
-export default function sagaMiddlewareFactory({ context = {}, ...options } = {}) {
-  const { sagaMonitor, logger, onError, effectMiddlewares } = options
+export default function sagaMiddlewareFactory(options = {}) {
+  const { context = {}, channel = stdChannel(), sagaMonitor, logger, onError, effectMiddlewares } = options
   let boundRunSaga
 
   if (process.env.NODE_ENV !== 'production') {
@@ -17,15 +16,12 @@ export default function sagaMiddlewareFactory({ context = {}, ...options } = {})
       check(onError, is.func, 'options.onError passed to the Saga middleware is not a function!')
     }
 
-    if (is.notUndef(options.emitter)) {
-      check(options.emitter, is.func, 'options.emitter passed to the Saga middleware is not a function!')
+    if (is.notUndef(channel)) {
+      check(channel, is.channel, 'options.channel passed to the Saga middleware is not a channel')
     }
   }
 
   function sagaMiddleware({ getState, dispatch }) {
-    const channel = stdChannel()
-    channel.put = (options.emitter || identity)(channel.put)
-
     boundRunSaga = runSaga.bind(null, {
       context,
       channel,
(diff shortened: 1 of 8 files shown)


packages/core/src/internal/middleware.js
```diff
@@ -15,15 +14,12 @@ export default function sagaMiddlewareFactory({ context = {}, ...options } = {})
       check(onError, is.func, 'options.onError passed to the Saga middleware is not a function!')
     }
 
-    if (is.notUndef(options.emitter)) {
-      check(options.emitter, is.func, 'options.emitter passed to the Saga middleware is not a function!')
+    if (is.notUndef(channel)) {
```
restrry: minor thing. should we use `is.channel` instead?

shinima: I'm using the specific checks below if options.channel if defined.  I'v found a small bug in `is.channel` below.

shinima:
@Andarist Do you want to have a review?

Andarist:
Yeah, sorry I have been out of touch lately, a lot going on in my life right now. I will try my best to review everything by the end of the week.

shinima:
Rebased upon master

Andarist:
Adding this to v1 milestone - want to review it soon (hopefully for real this time 😅 )

shinima:
Rebased on master, and add some tests to show how does a channel connect to external IO with `enhancePut`.

Andarist:
I've looked through this and it seems to me that it complicates things a little bit - just to allow for enhancing the put method, which could be done easier in userland now (when `emitter` got introduced to the codebase initially it wasnt possible, because we didnt expose `stdChannel` back then, in fact each saga had its own `stdChannel`, so things were quite different).

We could make `sagaMiddlewareFactory` accept a `channel` option which would default to `= stdChannel()`, we would remove `emitter` option entirely and then people could do this

```js
const channel = stdChannel()
const put = channel.put
channel.put = action => {
  if (action.type === 'batch') {
    action.batch.forEach(put)
    return
  }

  put(action)
} 
const sagaMiddleware = createSagaMiddleware({ channel })
```

WDYT?

shinima:
@Andarist You proposed changes seems very nice.  I was thinking that enhance-able channels are somewhat overlap with the functionality of `dispatch`. To enhance a stdChannel, one could use the `enhancePut` method , or put the logic inside the `dispatch` option. Exposing two options with overlapped functionality makes things complex.

I'll change a new PR that only includes 1. removing the `emitter` option from `sagaMiddlewareFactory`,  2. adding a `channel` option.

shinima:
Close this in favor of #1654.

PR #1542, opened 2018-07-25
Changes from the reviews of some other PR

shinima:
| Major: Breaking Change?  |  See below  |

Changes from the reviews of some other PR.  See https://github.com/redux-saga/redux-saga/pull/1527#pullrequestreview-139999445 and https://github.com/redux-saga/redux-saga/pull/1523#discussion_r204892463 for more detail.

### DIFF
--- a/packages/core/src/internal/channel.js
+++ b/packages/core/src/internal/channel.js
@@ -156,10 +156,7 @@ export function multicastChannel() {
     closed = true
     const takers = (currentTakers = nextTakers)
 
-    for (let i = 0; i < takers.length; i++) {
-      const taker = takers[i]
-      taker(END)
-    }
+    takers.forEach(taker => taker(END))
 
     nextTakers = []
   }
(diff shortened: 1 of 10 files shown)


packages/core/src/internal/channel.js
```diff
@@ -156,10 +156,7 @@ export function multicastChannel() {
     closed = true
     const takers = (currentTakers = nextTakers)
 
-    for (let i = 0; i < takers.length; i++) {
-      const taker = takers[i]
-      taker(END)
-    }
+    takers.forEach(taker => taker(END))
```
Andarist: we dont care about return value, so it's better to wrap the body in curly braces
```js
takers.forEach(taker => { taker(END) })
```

Andarist:
Left a minor comment, other than that it's cool!

shinima:
Code updated. 😃

Issue #1478, opened 2018-06-07
Update docs to v1 API


restrry:
Api changes that should be reflected in docs. Also would be nice to add examples, especially for new effect
- [x] put.resolve —> putResolve (#1492)
- [x] task.done —> task.toPromise (#1456)
- [x] take.maybe —> takeMaybe (#1492)
- [x] `delay` as effect (#1492)
- [x] `takeLeading` effect (#1492)
- [x] `retry` effect
- [x] `debounce` effect (#1492)
- [x] eventChannel emitting errors (#1519)
- [x] babel plugin usage example (https://github.com/redux-saga/redux-saga/pull/1546)
- [x] list of library changes from 0.x to 1.0 (https://github.com/redux-saga/redux-saga/pull/1613)
- [x] upgrade path from 0.x to 1.0 (https://github.com/redux-saga/redux-saga/pull/1613)
- [x] update examples in docs to new api (https://github.com/redux-saga/redux-saga/pull/1636)
- [x] remove logError https://github.com/redux-saga/redux-saga/pull/1658 (https://github.com/redux-saga/redux-saga/pull/1700)
- [x] change sagaStack passing https://github.com/redux-saga/redux-saga/pull/1695 (https://github.com/redux-saga/redux-saga/pull/1700)

Slapbox:
Is there an ETA for stable v1.0? A month? 6 months?

Some rough estimate, however rough, would be really helpful. Sorry, I know a little off topic, but I don't want to clutter the issues.

Thanks to everyone who works on this incredible project.

Andarist:
@Slapbox no rough estimate, probably few months (2-3?), we are moving quite slowly, although things got a little bit faster lately

Andarist:
Keep in mind that v1 will be just a small evolution and not a revolution, core APIs are here to stay and released betas should be stable, after all the codebase is tested 😉 I would only advise to pin your redux-saga dep to an exact version and you should be ready to go (after reading our release notes)

kamalkne:
Is anyone working on this issue? I am interested to work on it.

restrry:
actually not. you are welcome, so feel free to get something from the list that doesn't have a link to PR

SeregaSE:
Can i take retry effect?

shinima:
@SeregaSE   Of course.

SeregaSE:
could we say that retry is an effect?

Andarist:
[...]

PR #1546, opened 2018-07-27
Saga stack in error docs

restrry:
| Fixed Issues?            | docs for `babel-plugin` usage example https://github.com/redux-saga/redux-saga/issues/1478 |
| Tests Added + Pass?      | |
| Any Dependency Changes?  | only for error-demo example   |

I'd keep both `logError` and `onError` to separate error logging and high-level error handling (note, not all cases for `logError` and covered with `onError`)


restrry:
need to fix hoisting. after adding `redux-saga-test-plan` tests started failing because the package is hoisted to the root `node_modules` and cannot require `redux-saga` package. probably it could be solved with https://github.com/redux-saga/redux-saga/pull/1557/files

packages/babel-plugin-redux-saga/package.json
```diff
@@ -12,6 +12,9 @@
   "scripts": {
     "test": "jest --env=node"
   },
+  "jest": {
```
Andarist: this shouldnt be needed because we already do this `"test": "jest --env=node"`

restrry: sure, should be deleted. that's an old hack to fix tests locally

packages/core/src/internal/proc.js
```diff
@@ -294,7 +294,8 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
 
         if (env.onError) {
           env.onError(result)
-        } else {
+        }
```
restrry: note: that's actually a change in api.

examples/error-demo/test/sagas.js
```diff
@@ -0,0 +1,58 @@
...
+
+  middleware
+    .run(errorInCallAsyncSaga)
+    .toPromise()
+    .catch((/*error*/) => {
+      const expected = [
+        ['error', 'ReferenceError: undefinedIsNotAFunction is not defined'],
+        [
+          'error',
+          'The above error occurred in task throwAnErrorSaga  src/sagas/index.js?16\n    created by errorInCallAsyncSaga  src/sagas/index.js?25\n',
```
restrry: I'we added one integration test for babel-plugin integration into core. probably that should be in `core` itself, requires some work for defining several babel configs (with and without plugin)

examples/error-demo/package.json
```diff
@@ -3,7 +3,8 @@
   "private": true,
   "version": "1.0.0-beta.1",
   "scripts": {
-    "start": "webpack-dev-server"
+    "start": "webpack-dev-server",
+    "test": "cross-env NODE_ENV=test babel-node test/sagas.js | tap-spec"
```
Andarist: could we use jest for this?

packages/core/src/internal/proc.js
Andarist: seems to me that we dont need both `onError` & `logError`, having them both is confusing

examples/error-demo/package.json
restrry: we can. but then I'd switch all the examples to `jest`, I've added tap only for consistency

packages/core/src/internal/proc.js
restrry: actually use case for them a bit different (at least now) 
1 `onError` - called once, when error bubbles to the root, provides info about error source. usually it's used in user land to get source of error and report them (logger is just another target!)
2 `logError` - can be called several times, called for an error bubbling to the root saga + internal inconsistency errors (when try to cancel already cancelled task).

probably we should add a note about their different nature or remove one of them, but on the whole it seems that they have different use cases

restrry:
@Andarist @shinima ping

shinima:
Overall, it's good change and the error-demo performs well on my PC. One thing I'm not sure is changing the behaviors of `env.logError`, because I have never used it before. How do you think? @Andarist

packages/core/src/internal/proc.js
```diff
@@ -292,7 +292,8 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
...
+        if (env.logError){
```
shinima: `if (env.logError){`  
I think we need a space between `)` and `{`  😃

Andarist:
Sorry, I haven't time yet to review this, we'll take care of this in following days. Just letting you know that I haven't forgotten, just don't have time lately to focus on this 😢

packages/core/src/internal/proc.js
Andarist: > updated: there is another problem with onError we build saga error stack only when error bubbles to the root, so if user handles errors on their own (say try/catch on the upper layer), they don't have normalized saga error stack

But that's OK I think. People shouldnt rely on the stack anyhow - it doesnt matter if we are talking about normal errors or saga errors, stack inspection is shady and unreliable. 

The whole point of saga stack is to improve developer experience because errors caught by saga are otherwise really cryptic. No runtime code should rely on them though.

I think we should get rid of `logError` entirely and just recommend using `onError` for logging purposes. Explaining in the docs that it's role is primarily for logging purposes and it's called only for entirely uncaught errors from which we couldnt recover.

restrry: @Andarist ok, what about saga internal inconsistency errors? don't log them?

Andarist: could we check how other libraries handle such errors? (ie. react)

restrry: react throws errors with https://github.com/facebook/react/blob/master/packages/shared/invariant.js

example with usage error:
https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOMTextarea.js#L44-L47
example with internal error:
https://github.com/facebook/react/blob/master/packages/shared/ReactErrorUtils.js#L120-L124

Andarist: so basically both are handled the same way, i think we can safely assume this is what people expect, so both type of errors should just get handled to `onError` or logged with `console.error` if `onError` is not provided

restrry: there at least 2 kinds of errors in React
- in user component. can be handled by `componentDidCatch`. user can recover from error, report etc.
- internal inconsistency in React. thrown as exception. user cannot handle it.

for me redux-saga `onError` hook is similar to `componentDidCatch`. not quite sure we should mix approaches. I'd throw an exception in case of internal saga error

restrry: ok, let's continue discussion in the related  issue. I'd merge PR

Andarist:
@restrry  could u resolve the conflict and merge this in?

PR #1555, opened 2018-08-10
Refactor effect inspection utils

shinima:
| Major: Breaking Change? | Yes. See below |
| Minor: New Feature?     |                |
| Tests Added + Pass?     |                |

API changes in this PR:

- API `asEffect` is removed
- API `is.effect` is added to test whether an object is an effect
- `effectTypes` is exported by `redux-saga/effects`.

I refactor the saga-monitor-example and split it into several modules. During refactoring the sagaMonitor, I find that there is an internal ROOT type. ROOT effects are used by sagaMonitor to indicate that a root saga starts to run. So I add ROOT as an effect type.

Effects inspection utils in this PR are relatively simple and straightforward. Documentation for effects should be updated and tells users that **Effect is an object that consists of a `type`, a `payload` and a internal symbol `IO`**. So users could import `is` and `effectTypes` to test whether an effect is of an specific type like in [checkers](https://github.com/redux-saga/redux-saga/pull/1555/files#diff-5d397e302f1805e01eb7aeb798443633).

How to inspect the details of the effect payload is still an question. This PR already uses the knowing of payload structure to log effects in a better format. We need to come up with a better way for users to access data in effect payload. Or maybe we can simply state in our doc: **You can log the effects and see the payload structure for debugging purposes, but you should not rely on it since payload structure may change over versions**.


shinima:
By the way, saga-monitor remains the same style of logging output.

image

examples/sagaMonitor/modules/logSaga.js
```diff
@@ -0,0 +1,156 @@
...
+
+  const winnerInd = desc.winner ? (isError ? '✘' : '✓') : ''
+  formatter.addLabel(winnerInd).addLabel(desc.label)
+
+  if (is.iterator(desc.effect)) {
+    formatter.addValue(desc.effect.name).addDescResult(desc, true)
+  } else if (is.promise(desc.effect)) {
+    formatter
+      .addEffectType('promise')
+      .resetStyle()
```
restrry: why we need to call `resetStyle` manually for all the cases except `addLabel`?

examples/sagaMonitor/modules/DescriptorFormatter.js
```diff
@@ -0,0 +1,56 @@
...
+import Formatter from './Formatter'
+import { CANCELLED, PENDING, REJECTED, RESOLVED } from './constants'
+
+const DEFAULT_STYLE = 'color: black'
+const LABEL_STYLE = 'font-weight: bold'
+const EFFECT_TYPE_STYLE = 'color: blue'
+const ERROR_STYLE = 'color: red'
+const CANCEL_STYLE = 'color: #ccc'
+
+export default class DescriptorFormatter extends Formatter {
```
restrry: are we going to reuse `Formatter`? imho implementation with inheritance a bit confusing.

examples/sagaMonitor/modules/logSaga.js
shinima: This is adapted from the old code... Let me find a screenshot for you.

examples/sagaMonitor/modules/DescriptorFormatter.js
shinima: The `Formatter` is a universal formatter which is adapted from the old code. And `DescriptorFormatter` is a special formatter that knows the structure of effect descriptors. So conceptually, `DescriptorFormatter` is a special kind of `Formatter`, so I use inheritance.

examples/sagaMonitor/modules/Manager.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  getRootIds() {
+    return this.rootIds
+  }
+
+  getChildIds(parentEffectId) {
+    return Object.keys(this.map)
+      .filter(effectId => this.map[effectId].parentEffectId === parentEffectId)
+      .map(Number)
```
Andarist: is this necessary? effect ids should be numbers already, right?

examples/sagaMonitor/modules/Manager.js
```diff
@@ -0,0 +1,30 @@
...
+
+  setAsRoot(effectId) {
+    this.rootIds.push(effectId)
+  }
+
+  getRootIds() {
+    return this.rootIds
+  }
+
+  getChildIds(parentEffectId) {
```
Andarist: seems like we should bookkeep the tree structure here too for perf reasons

examples/sagaMonitor/modules/Manager.js
shinima: Keys of object are always strings, and `Object.keys(obj)` returns an array of strings.
image

examples/sagaMonitor/modules/Manager.js
shinima: Yes. Keep an map from parent to children would enhance the performance.

shinima:
Thanks for reviewing. And I'm currently thinking about add another package `@redux-saga/inpection` that exports some utils to allow users to inspect the internals of effects.  `@redux-saga/inpection` is for debugging purposes and `@redux-saga/core` does not depend on it.

I'll update this PR after implementing the `@redux-saga/inpection`.

packages/core/src/internal/io.js
```diff
@@ -7,16 +7,19 @@ const TEST_HINT =
 
 const makeEffect = (type, payload) => ({ [IO]: true, type, payload })
 
-const isForkEffect = eff => eff && eff[IO] && eff.type === 'FORK'
-
 export const detach = eff => {
   if (process.env.NODE_ENV === 'development') {
+    const isForkEffect = eff => is.effect(eff) && eff.type === effectTypes.FORK
```
Andarist: why the change? why to move it to the function?

examples/sagaMonitor/modules/Manager.js
Andarist: oh, yeah - havent quite noticed they are coming here from Object.keys

Andarist:
Could u rebase this PR? 

I'm wondering if we need more inspection utils - they will only abstract simple object property access. I would be good with providing only `is.effect` + effectTypes. WDYT?

packages/core/src/internal/effectTypes.js
```diff
@@ -1,3 +1,4 @@
+export const ROOT = 'ROOT'
```
Andarist: i dont think I quite like this change - seems like `ROOT` is different from other effects because its purely internal

could u describe what do u think about it? what are advantages of having it in your opinion? (remember that I can always be persuaded! 😉 )

packages/core/src/internal/io.js
shinima: `isForkEffect` is used only in development, so I move it inside the if-block to get better code elimination. 

However it turns out that even if `isForkEffect` is placed at top level, the code of `isForkEffect` would be removed when redux-saga builds. 😅

packages/core/src/internal/effectTypes.js
shinima: From the perspective of the people who uses redux-saga, ROOT is a different and should be kept internal only; From the perspective of saga-monitor, ROOT is of no difference compared to ALL/RACE/CALL.

IMO, ROOT is a valid effect type, but it does not need to be exported because users never yield a root effect directly. I'll remove it from *effectTypes.js*.

shinima:
> I'm wondering if we need more inspection utils - they will only abstract simple object property access. I would be good with providing only is.effect + effectTypes. WDYT?

Actually I think **only is.effect + effectTypes** is fine. And as the effect payload shape is stable, I'm inclining to let people inspect the payload directly for (and only for) debugging purposes (we may add some tips into our docs about what an effect looks like).

shinima:
Rebased on master. And add `manager.childIdsMap` for bookkeeping a map from parent-id to child-id-array, also removes ROOT from effectTypes.js.

packages/core/src/internal/io.js
Andarist: yeah, in general function declarations and expressions are one of the easiest things to dead code eliminate, it's better to hoist this to the top level scope to avoid recreating this function inside each detach call

packages/core/src/internal/effectTypes.js
Andarist: if we decide to keep it then it should stay in effectTypes, it shouldnt be exported from effects (as in `import {root} from 'redux-saga/effects'`

there is also a perspective of redux-saga itself - this effect is not interpreted like other effects, its executed eagerly on the startup

shinima: Actually users can not do `import {root} from 'redux-saga/effects'` as `root` is not re-exported in core/src/effects.js.

I'm a little bit of confused about what to do now... Should we add ROOT as a normal effect like ALL/RACE? Or should we remove ROOT from effectTypes, and keep `root` as internal? Or do you have any nicer ideas?

shinima:
> it's better to hoist this to the top level scope to avoid recreating this function inside each detach call

done.

packages/core/src/internal/effectTypes.js
Andarist: im quite sure it shouldnt be exported, im still unsure if it should be an effect at all 🤔 this only seems to satisfy monitor needs, from core's perspective it still seems to be weird imho - @restrry  wdyt?

Andarist: OTOH it is super minor thing, and was classified as an effect already when passing info about it to the monitor

restrry: imho that's not an effect per se, as it  has other lifecycle for end user (import - call - processing - return a result) and more over, users don't know about it, as it internal detail.

Andarist: it could be kept as purely internal effect - not user facing, but I think I like `rootSagaStarted` hook proposal by @shinima better, it's more explicit and less confusing about what effects are

restrry: +1

shinima:
`rootSagaStarted` hook added.

packages/core/effects.d.ts
```diff
@@ -706,3 +706,21 @@ export function debounce<T, T1, T2, T3, T4, T5, T6>(
   arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5, arg6: T6,
   ...rest: any[]
 ): ForkEffect;
+
+export const effectTypes: {
```
Andarist: I'd probably like to move it to separate module - but gonna handle it in separate PR

packages/core/src/utils.js
```diff
@@ -1,4 +1,2 @@
 export { SAGA_ACTION, TASK } from '@redux-saga/symbols'
```
Andarist: the `redux-saga/utils` entry is starting to look useless, we should get rid of it later in favour of modular packages

Issue #1533, opened 2018-07-20
`delay` is confusing


shinima:
In v1.0.0-beta.1, users could import `delay` from `'redux-saga/effects'`. Here `delay` is an effect creator that means users could only use it like `yield delay(2000)`.

But users could still import `delay` from `'redux-saga/utils'`, and here `delay` is a function that returns a Promise. Users could use this `delay` either in `yield delay(2000)` form or in `yield call(delay, 2000)` form.

These two `delay`s are confusing, they have the same name and they both come from redux-saga, but has different meanings and usages.

restrry:
imho we shouldn't export delay from `utils` to avoid confusing and implementing it in user land (if needed) shouldn't be a problem

shinima:
I agree to remove one of them.  But I'm inclining to remove `delay` from `'redux-saga/effects'`.

Because we can use delay-function to create delay-effects, but we can not use delay-effect-creator to create a delay-Promise. In other word, delay function is more general.

And effect-creators from `'redux-saga/effects'` often create effect with corresponding type -- `call()` creates call-effect, `all()` creates all-effects, `fork()` creates fork-effects, etc.. But `delay()` create call-effects, which is not consistent with other effect-creators.

restrry:
helpers create not corresponding effects - takeLatest returns `fork`, for example.
I think the main reason to introduce that change was a bit confusing nature of delay for newcomers. before you have import 2 different functions from saga and use them in right order, you need to understand that `delay` returns promise, that saga can handle promise correctly etc.
```js
import {delay} from 'redux-saga'
import {call} from 'redux-saga/effects'
// then in your code
function* (){
    yield call(delay, 2000)
    ...
}
```

Andarist:
> imho we shouldn't export delay from utils to avoid confusing and implementing it in user land (if needed) shouldn't be a problem

I second that.

> before you have import 2 different functions from saga and use them in right order, you need to understand that delay returns a promise, that saga can handle promise correctly etc.

That was exactly the reason to introduce `delay` as effect. Things like `delay`, `throttle`, `takeEvery` & `takeLatest` are not technically effects, but keeping them separate was quite confusing for many people - so I've decided to just move them to effects for simplicity.

You don't have to worry about it, I'm going to move this `delay` util to separate package (`@redux-saga/delay-promise` or something) or move it out of this repository entirely and just make it a dependency of `redux-saga`.

restrry:
[...]

PR #1557, opened 2018-08-10
More monorepo packages

Andarist:
It's still WIP, probably need to fix tests (havent check them out yet)

Related to https://github.com/redux-saga/redux-saga/issues/1213 , fixes https://github.com/redux-saga/redux-saga/issues/1533


packages/symbols/rollup.config.js
```diff
@@ -0,0 +1,30 @@
+import babel from 'rollup-plugin-babel'
+import pkg from './package.json'
+
+const makeExternalPredicate = externalArr => {
```
restrry: should we extract a function into another build-helper file?

restrry:
what `npm` version do you use? cannot install deps in this branch

packages/symbols/rollup.config.js
Andarist: ideally I'd like to just drop low-level rollup for microbundle (its a rollup wrapper) in the future, so wouldnt bother with that, just need to find some time to contribute to microbundle more to cover more advanced use cases 😅

Andarist:
6.1.0, dont have any issues with installing deps. I'm migrating to jest from tape for the core package right now, after I do that I think this PR will be ready to review (CI should be green too)

prettier.config.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+  printWidth: 120,
+  singleQuote: true,
+  semi: false,
+  excludeFiles: 'fixtures/**/*-expected.js',
```
shinima: `excludeFiles` is only allowed in `overrides` sections.

> [Configuration Overrides](https://prettier.io/docs/en/configuration.html#configuration-overrides) files is required for each override, and may be a string or array of strings. excludeFiles may be optionally provided to exclude files for a given rule, and may also be a string or array of strings.

prettier.config.js
```diff
@@ -0,0 +1,14 @@
...
+  overrides: [
+    {
+      files: ['packages/**/src/**/*.js', 'examples/**/src/**/*.js'],
+      options: {
+        trailingComma: 'all',
```
shinima: I think `trailingComma: 'all'` should move to the top level. ESLint starts to complains some configuration files (e.g. rollup.config.js) if I use prettier to format these files.

shinima:
Should we name our es-modules as `*.mjs`?  `.mjs` seems a better extension for me. [Node supports ECMAScript Modules as *.mjs files](https://nodejs.org/api/esm.html#esm_enabling), and [webpack supports *.mjs out of box](https://webpack.js.org/configuration/resolve/#resolve-extensions) too.

packages/symbols/rollup.config.js
restrry: should we create appropriate issue?

Andarist: its not rly actionable right now, i would only like to move to microbundle if it will be able to tackle all the required tasks with simpler config (so config doesnt have to be updated in multiple places, which I tend to do a lot :/)

prettier.config.js
Andarist: 😢 do u know how to setup this correctly then? we need to have those files untouched by prettier and I find it easier to do this in overrides-like section of the config than in a single glob matching which files have to be prettified (like before)

prettier.config.js
Andarist: good point, it's supported in trailing commas in functions are allowed since node@8 so we should be able to use it in all files

packages/core/codemod.js
```diff
@@ -0,0 +1,31 @@
+module.exports = ({ types: t, template }) => {
```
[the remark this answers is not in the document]
Andarist: oh, this is leftover, has to be removed - I've used it to migrate 😅 but the migration is complete now (at least in core package)

Andarist:
> Should we name our es-modules as *.mjs?

Personally I'm on the fence, would prefer to keep .js for now.

> .mjs seems a better extension for me. Node supports ECMAScript Modules as *.mjs files

It doesnt bring much benefit to node users for now as we distribute CJS format either way.

> and webpack supports *.mjs out of box too.

This also doesnt matter much at the moment, because we distribute `"module"` entry in package.json.

packages/symbols/rollup.config.js
restrry: ok. let's then move the helper in one place. later we can migrate to other builder

Andarist:
Also, sorry for the big PR! 😅 I hoped to do less in it, but kinda went down the rabbit hole

packages/core/jest.config.js
```diff
@@ -0,0 +1,9 @@
+const lernaAliases = require('lerna-alias').jest()
+
+module.exports = {
+  testEnvironment: 'node',
+  moduleNameMapper: lernaAliases,
+  transform: {
```
Andarist: this is kinda a temporary hack for now, we should be able to just use `babel-jest`, but we'd have to install https://github.com/babel/babel-bridge and cant do it right now because we test babel plugin with both babel@6 and babel@7

prettier.config.js
shinima: Oh, I find that we could [add a .prettierignore file](https://prettier.io/docs/en/ignore.html#ignoring-files)

> To exclude files from formatting, add entries to a .prettierignore file in the project root or set the --ignore-path CLI option.

packages/symbols/rollup.config.js
Andarist: not sure how to share such a helper appropriately between packages without making it a package (and i dont think its worth to make it a package), and imho we shouldnt reach out to anything outside a single package directory - that would break package encapsulation

Andarist:
@shinima would love your review on this one :)

shinima:
Okay, I'll do a review now.

packages/core/rollup.config.js
```diff
@@ -1,41 +1,102 @@
...
+  external: makeExternalPredicate(external === 'peers' ? peerDeps : deps.concat(peerDeps)),
   plugins: [
+    alias(lernaAlias()),
     nodeResolve({
       jsnext: true,
     }),
     babel({
       exclude: 'node_modules/**',
-      externalHelpersWhitelist: babelHelpersList.filter(helperName => helperName !== 'AsyncGenerator'),
+      babelrcRoots: path.resolve(__dirname, '../*'),
```
shinima: `path.resolve(__dirname, '../*')` points to the `packages/` folder. Is that the correct folder? 😕

shinima:
I have trouble in installing deps with npm (using yarn is okay), and I fails to run build script on my Windows 😂. However tests are passing in CI, maybe I need to switch a Mac.

packages/core/rollup.config.js
Andarist: yeah, that was the intended directory

babel@7 stopped using babelrc files outside of current working directory (which is packages/core here) - one have to opt in to using them with `babelrcRoots` option. 

The intention of this is to always build (& test etc) from our source directories instead of some build artifacts which might be outdated at the time of running the script

Andarist:
> I have trouble in installing deps with npm (using yarn is okay), and I fails to run build script on my Windows 😂. However tests are passing in CI, maybe I need to switch to a Mac.

That's not good, could u paste in the error? Ideally it should work fine on any system.

restrry:
@shinima I had a problem as well. then I switched to node@10.1.0 and npm@5.6.0 and it resolved the problem. forgot to add a comment here and investigate - too busy on work now

shinima:
node version: v10.7.0
npm version: 6.1.0
OS version: win 10

Logs when I run `npm install`:

```log
$ npm i

> undefined postinstall D:\workspace\redux-saga
> npm run bootstrap


> @ bootstrap D:\workspace\redux-saga
> lerna bootstrap --ignore-scripts

lerna info version 3.0.6
lerna info Bootstrapping 14 packages
lerna info Installing external dependencies
lerna info hoist Installing hoisted dependencies into root
lerna info hoist Pruning hoisted dependencies
lerna info hoist Finished pruning hoisted dependencies
lerna ERR! npm install exited 4294963238 in 'redux-saga'
lerna ERR! npm install stderr:
npm WARN redux-devtools@3.4.1 requires a peer of redux@^3.5.2 but none was installed.

npm ERR! path D:\workspace\redux-saga\node_modules\npm\node_modules\update-notifier
npm ERR! code ENOENT
npm ERR! errno -4058
npm ERR! syscall rename
npm ERR! enoent ENOENT: no such file or directory, rename 'D:\workspace\redux-saga\node_modules\npm\node_modules\update-notifier' -> 'D:\workspace\redux-saga\node_modules\npm\node_modul
es\.update-notifier.DELETE'
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent
```

If I set `npmClient` to `'yarn'` in lerna.json and use yarn to install deps, then the installation is okay. But yarn does not respect `--ignore-scripts` and postbootstrap script will be executed.

--- 

Logs when I run `npm run build` in the core package:

```log
$ npm run build

> redux-saga@1.0.0-beta.1 prebuild D:\workspace\redux-saga\packages\core
> npm run clean


> redux-saga@1.0.0-beta.1 clean D:\workspace\redux-saga\packages\core
> rimraf dist


> redux-saga@1.0.0-beta.1 build D:\workspace\redux-saga\packages\core
> rollup -c


src/index.js, src/effects.js, src/utils.js → dist...
[!] (babel plugin) SyntaxError: D:\workspace\redux-saga\packages\core\src\internal\middleware.js: Support for the experimental syntax 'objectRestSpread' isn't currently enabled (7:63):

   5 | import { runSaga } from './runSaga'
   6 | 
>  7 | export default function sagaMiddlewareFactory({ context = {}, ...options } = {}) {
     |                                                               ^
   8 |   const { sagaMonitor, logger, onError, effectMiddlewares } = options
   9 |   let boundRunSaga
  10 | 

Add @babel/plugin-proposal-object-rest-spread (https://git.io/vb4Ss) to the 'plugins' section of your Babel config to enable transformation.
src\internal\middleware.js (7:62)
SyntaxError: D:\workspace\redux-saga\packages\core\src\internal\middleware.js: Support for the experimental syntax 'objectRestSpread' isn't currently enabled (7:63):

   5 | import { runSaga } from './runSaga'
   6 |
>  7 | export default function sagaMiddlewareFactory({ context = {}, ...options } = {}) {
     |                                                               ^
   8 |   const { sagaMonitor, logger, onError, effectMiddlewares } = options
   9 |   let boundRunSaga
  10 |

Add @babel/plugin-proposal-object-rest-spread (https://git.io/vb4Ss) to the 'plugins' section of your Babel config to enable transformation.
    at Parser.raise (D:\workspace\redux-saga\packages\core\node_modules\@babel\parser\lib\index.js:3893:15)
    at Parser.expectPlugin (D:\workspace\redux-saga\packages\core\node_modules\@babel\parser\lib\index.js:5227:18)
    at Parser.parseObj (D:\workspace\redux-saga\packages\core\node_modules\@babel\parser\lib\index.js:6632:14)
    at Parser.parseBindingAtom (D:\workspace\redux-saga\packages\core\node_modules\@babel\parser\lib\index.js:5479:21)
    at Parser.parseMaybeDefault (D:\workspace\redux-saga\packages\core\node_modules\@babel\parser\lib\index.js:5542:25)
    at Parser.parseAssignableListItem (D:\workspace\redux-saga\packages\core\node_modules\@babel\parser\lib\index.js:5524:21)
    at Parser.parseBindingList (D:\workspace\redux-saga\packages\core\node_modules\@babel\parser\lib\index.js:5516:24)
    at Parser.parseFunctionParams (D:\workspace\redux-saga\packages\core\node_modules\@babel\parser\lib\index.js:7803:24)
    at Parser.parseFunction (D:\workspace\redux-saga\packages\core\node_modules\@babel\parser\lib\index.js:7790:10)
    at Parser.parseExportDefaultExpression (D:\workspace\redux-saga\packages\core\node_modules\@babel\parser\lib\index.js:8165:19)

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! redux-saga@1.0.0-beta.1 build: `rollup -c`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the redux-saga@1.0.0-beta.1 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

This seems like an incorrect setup of babel/rollup for the core package.

packages/core/jest.config.js
restrry: after tomorrow's babel stable release we can start supporting only 7 version

restrry:
switched back to node 10.8.0 and got this error https://github.com/npm/npm/issues/17444
removing package-lock.json and `npm i` fixed the problem. 
@shinima could you try to do the same?

Andarist:
I've deleted lock file for the time being, its gains seems to be questionable for libraries anyway.

restrry:
I think we should add it to `.gitignore` file then

Andarist:
Sure, can do that, but it shouldnt be created anymore now with current `.npmrc` anyway.

restrry:
do we have `.npmrc`?

Andarist:
Oh, lol, I haven't committed it - I've created it while removing the lock itself. Gonna push it in a moment

PR #1566, opened 2018-08-23
Add default type to PutEffect

screendriver:
| Fixed Issues?            | - |
| Patch: Bug Fix?          |  No  |

Just a small change to the TypeScript definition of the `PutEffect`. With the default type to it's generic argument you don't have to provide `PutEffect<any>` all the time (when you want to ignore the action completely) and just can use `PutEffect`.


Andarist:
Thanks for the change (@screendriver ) and for the review (@aikoven )!

screendriver:
Uh that was fast 😱 👍

PR #1568, opened 2018-08-24
Remove support for join(...tasks) and cancel(...tasks)

shinima:
| Major: Breaking Change?  |  See below  |

Remove support for join(...tasks) and cancel(...tasks). See https://github.com/redux-saga/redux-saga/pull/1527#pullrequestreview-139999445 .

### DIFF
--- a/packages/core/src/internal/io.js
+++ b/packages/core/src/internal/io.js
@@ -1,7 +1,7 @@
 import delayP from '@redux-saga/delay-p'
 import * as is from '@redux-saga/is'
 import { IO, SELF_CANCELLATION } from '@redux-saga/symbols'
-import { log, once, array, identity, check, createSetContextWarning } from './utils'
+import { identity, check, createSetContextWarning } from './utils'
 import * as effectTypes from './effectTypes'
 
 const TEST_HINT =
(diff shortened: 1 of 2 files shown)


packages/core/src/internal/io.js
```diff
@@ -114,18 +114,11 @@ export function spawn(fn, ...args) {
...
 export function join(taskOrTasks) {
-  if (arguments.length > 1) {
-    if (process.env.NODE_ENV === 'development') {
-      printJoinDeprecationWarning()
-    }
-    taskOrTasks = array.from(arguments)
-  }
   if (process.env.NODE_ENV === 'development') {
+    if (arguments.length > 1) {
+      throw new Error('join(...tasks) is not supported any more. Please use join([...tasks]) to join multiple tasks.')
```
restrry: `Please use join(tasks)`?

shinima: Should I rephrase this error message?

Andarist: Hm, not sure which one is better - `join([...tasks])` seems to indicate that the argument type is an array, `join(tasks)` does not hold any information regarding this (except pluralized word)

restrry: that's not a big deal. we can keep it as is. I meant, probably it could be confusing - for example, why we require to make a copy of array before passing to effect?

Andarist: let's keep it then, I doubt people will actually make those copies, at least not most commonly

Issue #1254, opened 2017-11-07
Allow using frozen actions


Alxandr:
I just updated redux-saga, to have everything break because we're using frozen actions. The `wrapSagaDispatch` fails when trying to expand our actions that are frozen.

Also, they make the actions non FSA compliant. This should really be optional.

Andarist:
Hi there! This high-level thing here cannot be optional, because it helps us in distinguishing saga-induces actions from the others and that helps with some edge cases in scheduling.

However I was looking at this 'hack' recently and Im very interested in changing it, because it's a little bit gross. Ideally I would just use `Set` for storing those actions temporarily, but I do not want to require people loading polyfills to support old environments. Any ideas how this could be solved? There is always possibility to store them in an array, but Im wondering if some better, maybe more performant, solution could be found.

Alxandr:
If the action is extensible or have a meta property that is extensible, add it there. If not conclude that it can't be added, because it will break things?

Or, instead of using `Object.defineProperty`, you could use `Object.create(originalAction, { saga-props })`

restrry:
`Object.create` resonates with redux action nature and it's official requirement to use plain object without inheritance https://github.com/reduxjs/redux/blob/master/src/createStore.js#L166
@Andarist we could use `WeakSet` or `WeakMap` and degrade to Array for old envs

restrry:
@shinima

shinima:
Could we wrap the actual channel payload into an object, and unwrap the object when we want to use payload?

 That means we change our `wrapSagaDispatch` as follows:

```diff
- export const wrapSagaDispatch = dispatch => action =>
-  dispatch(Object.defineProperty(action, SAGA_ACTION, { value: true }))

+ export const wrapSagaDispatch = dispatch => action => dispatch({ [SAGA_ACTION]: true, action })
```

stdChannel also needs changed:
```diff
export function stdChannel() {
  const chan = multicastChannel()
  const { put } = chan
  chan.put = input => {
    if (input[SAGA_ACTION]) {
-      put(input)
+     put(input.action)
      return
    }
-    asap(() => put(input))
+ // I'm not sure at the moment that `input.action` or `input` should be used here.
+    asap(() => put(input.action))
  }
  return chan
}
```

This approach seems to work and has less magic than  WeakSet/WeakMap.
How do you think about it ? @Andarist @restrry

Andarist:
[...]

PR #1578, opened 2018-09-03
Warn users about putting frozen actions

Andarist:
related to #1254


patrick-mcdougle:
This error message also occurs if you accidentally put a string (like a redux type constant) instead of a redux action (object with type property). Not really anything actionable, just saving other users from making the same mistake I did by posting here.

PR #1579, opened 2018-09-03
Issue#1478 retry

SeregaSE:
Add `retry` effect doc
(https://github.com/redux-saga/redux-saga/issues/1478)[related issue]

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -897,21 +897,23 @@ const debounce = (ms, pattern, task, ...args) => fork(function*() {
 You can also handel a channel as argument and the behaviour is the same as [`debounce(ms, pattern, saga, ..args)`](#debouncems-pattern-saga-args)
 
 ### `retry(maxTries, delay, fn, ...args)`
- Creates an Effect description that instructs the middleware to call the function `fn` with `args` as arguments while `fn` does not return the result or `fn` calls count less then `maxTries`. `Delay` instructs `retry` to make a pause between `fn` calls.
- - `maxTries: Number` - maximum calls count.
- 
+
+Creates an Effect description that instructs the middleware to call the function `fn` with `args` as arguments while `fn` does not return the result or `fn` calls count less then `maxTries`. `Delay` instructs `retry` to make a pause between `fn` calls.
+
+- `maxTries: Number` - maximum calls count.
 - `delay: Number` - length of a time window in milliseconds between `fn` calls.
- - `fn: Function` - A Generator function, or normal function which either returns a Promise as result, or any other value.
- - `args: Array<any>` - An array of values to be passed as arguments to `fn`
+- `fn: Function` - A Generator function, or normal function which either returns a Promise as result, or any other value.
+- `args: Array<any>` - An array of values to be passed as arguments to `fn`
 
- #### Example
- 
- In the following example, we create a basic task `retrySaga`. We use `retry` to try fetch our API 3 times with 10 second interval. If `request` fails first time then `retry` will call `request` one more time while calls count less than 3.
+#### Example
 
- ```javascript
+In the following example, we create a basic task `retrySaga`. We use `retry` to try fetch our API 3 times with 10 second interval. If `request` fails first time then `retry` will call `request` one more time while calls count less than 3.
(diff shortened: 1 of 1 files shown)


docs/api/README.md
```diff
@@ -895,6 +896,36 @@ const debounce = (ms, pattern, task, ...args) => fork(function*() {
...
+Creates an Effect description that instructs the middleware to call the function `fn` with `args` as arguments while `fn` does not return the result or `fn` calls count less then `maxTries`. `Delay` instructs `retry` to make a pause between `fn` calls.
+
+- `maxTries: Number` - maximum calls count.
+- `delay: Number` - length of a time window in milliseconds between `fn` calls.
+- `fn: Function` - A Generator function, or normal function which either returns a Promise as result, or any other value.
+- `args: Array<any>` - An array of values to be passed as arguments to `fn`
+
+#### Example
+
+In the following example, we create a basic task `retrySaga`. We use `retry` to try fetch our API 3 times with 10 second interval. If `request` fails first time then `retry` will call `request` one more time while calls count less than 3.
```
restrry: `to try` --> `to try to`

docs/api/README.md
```diff
@@ -895,6 +896,36 @@ const debounce = (ms, pattern, task, ...args) => fork(function*() {
...
+
+```javascript
+import { put, retry } from 'redux-saga/effects'
+import { request } from 'some-api';
+
+function* retrySaga(data) {
+  try {
+    const response = yield retry(3, 1000 * 10, request, data)
```
restrry: in sake of readability
const SECOND = 1000
const response = yield retry(3, 10 * SECOND, request, data)

docs/api/README.md
```diff
@@ -895,6 +896,36 @@ const debounce = (ms, pattern, task, ...args) => fork(function*() {
 ### `debounce(ms, channel, saga, ...args)`
 You can also handel a channel as argument and the behaviour is the same as [`debounce(ms, pattern, saga, ..args)`](#debouncems-pattern-saga-args)
 
+### `retry(maxTries, delay, fn, ...args)`
+
+Creates an Effect description that instructs the middleware to call the function `fn` with `args` as arguments while `fn` does not return the result or `fn` calls count less then `maxTries`. `Delay` instructs `retry` to make a pause between `fn` calls.
```
restrry: result or --> result of
then --> than

docs/api/README.md
```diff
@@ -895,6 +896,36 @@ const debounce = (ms, pattern, task, ...args) => fork(function*() {
...
+
+- `maxTries: Number` - maximum calls count.
+- `delay: Number` - length of a time window in milliseconds between `fn` calls.
+- `fn: Function` - A Generator function, or normal function which either returns a Promise as result, or any other value.
```
restrry: `as result,` --> `as a result`

docs/api/README.md
restrry: probably nitpick - from the first look is not clear that effect try to call `fn` again if the previous call threw an error

SeregaSE:
I will update PR tomorrow, thanks)

docs/api/README.md
```diff
@@ -895,6 +896,36 @@ const debounce = (ms, pattern, task, ...args) => fork(function*() {
 ### `debounce(ms, channel, saga, ...args)`
 You can also handel a channel as argument and the behaviour is the same as [`debounce(ms, pattern, saga, ..args)`](#debouncems-pattern-saga-args)
 
+### `retry(maxTries, delay, fn, ...args)`
+Creates an Effect description that instructs the middleware to call the function `fn` with `args` as arguments while `fn` does not return the result of `fn` calls count less than `maxTries`. `Delay` instructs `retry` to make a pause between `fn` calls.
```
restrry: we could simplify a bit
```
Creates an Effect description that instructs the middleware to call the function `fn` with `args` as arguments.
In case of failure will try to make another call after `delay` milliseconds, if number of attempts < `maxTries`
```
WDYT?

SeregaSE: Yeap, i think nice, ty.

PR #1582, opened 2018-09-04
Make `detach` immutable

shinima:
| Major: Breaking Change?  |  Yes, but it is unlikely to break things |

Make `detach` immutable that returns a new effect rather than modifying the argument. Since we are treating effects as immutable objects, and we encourage users to treat effects as opaque objects, so it is better not to modify effects.

### DIFF
--- a/packages/core/src/internal/io.js
+++ b/packages/core/src/internal/io.js
@@ -1,7 +1,7 @@
 import delayP from '@redux-saga/delay-p'
 import * as is from '@redux-saga/is'
 import { IO, SELF_CANCELLATION } from '@redux-saga/symbols'
-import { identity, check, createSetContextWarning } from './utils'
+import { check, createSetContextWarning, identity } from './utils'
 import * as effectTypes from './effectTypes'
 
 const TEST_HINT =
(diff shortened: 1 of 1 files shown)


shinima:
I changed the `object.assign` a bit and make it more align with the native `Object.assign`. However the bundle size gets 36B larger 😓 . Should I revert `object.assign`?

packages/core/src/internal/io.js
```diff
@@ -1,22 +1,22 @@
...
+import { identity, check, createSetContextWarning, object } from './utils'
 import * as effectTypes from './effectTypes'
 
 const TEST_HINT =
   '\n(HINT: if you are getting this errors in tests, consider using createMockTask from redux-saga/utils)'
 
 const makeEffect = (type, payload) => ({ [IO]: true, type, payload })
 
-const isForkEffect = eff => eff && eff[IO] && eff.type === 'FORK'
+const isForkEffect = eff => eff && eff[IO] && eff.type === effectTypes.FORK
```
Andarist: 👍

packages/core/src/internal/io.js
```diff
@@ -1,22 +1,22 @@
...
 
 export const detach = eff => {
   if (process.env.NODE_ENV === 'development') {
     check(eff, isForkEffect, 'detach(eff): argument must be a fork effect')
   }
-  eff.payload.detached = true
-  return eff
+  const detachedPayload = object.assign({}, eff.payload, { detached: true })
```
Andarist: this could be just `const payload = { ...eff.payload, detached: true }`

shinima:
> this could be just const payload = { ...eff.payload, detached: true }
Done.

Issue #1512, opened 2018-06-28
rootSagas what is the best way


tastafur:
### rootSagas use all effect simple or use all + fork, which is the best option?

thanks for the help, I've been reading and reading and it's not clear to me

zgavin1:
Hey @tastafur I just actually wrote some docs to help explain this here: https://github.com/redux-saga/redux-saga/blob/master/docs/advanced/RootSaga.md. _But_ I realized that I never added a link to  this page to the Table of Contents. I will make a PR for that asap.

Andarist:
Need to publish them manually for now, I'm not sure though if current docs are not mix of v0.16 and v1. We are heading towards making v1 stable, so I would prefer holding for a little bit longer with deploying the docs.

tastafur:
@zgavin1 I have read your documentation and what I understand in this way 
```javascript
export default function* rootSaga() {
  yield all([
    helloSaga(),
    watchIncrementAsync()
  ])
}
```
you initiate non-blocking parallel effects but in this way also
```javascript
export default function* root() {
  yield all([
    fork(saga1),
    fork(saga2),
    fork(saga3)
  ]);
}
```
I do not understand the difference between these two ways of implementing rootSagas

shinima:
I'm thinking that `yield all([ fork(saga1), fork(saga2), fork(saga3) ])` is not a good pattern to implement sagas.  It may **mislead** people to think that the fork-effects are managed by the call-effect, and call-effect will be resolved after all the forked tasks completes. 
But actually, the fork-effects are always managed by the current task through the underlying forkQueue. Call-effect will be resolved immediately since all fork-effects are non-blocking. Errors from the forked tasks bubble to the current task rather than to the call-effect.

So it is better to design root saga as follows:

```diff
- yield all([ fork(saga1), fork(saga2) ])

// using multiple fork-effects
+ yield fork(saga1)
+ yield fork(saga2)

// or ...
+ yield all([ call(saga1), call(saga2)])
```

By the way, fork-effects in race-effect is  most likely a bug since fork-effects are non-blocking and they will always win the race immediately.

``` javascript
// DO NOT DO THIS. The fork-effect always win the race immediately.
yield race([
  fork(saga1), 
  take('ACTION-1'),
  somePromise,
])
```

The rule of thumb is that use fork-effect directly in your saga, do not nest it in other effects.

zgavin1:
[...]

PR #1583, opened 2018-09-04
docs: Update RootSaga.md

shinima:
I find some part of RootSaga.md may be misleading so I update the doc. And I add some more doc to 
RootSaga.md.

Update list:
* Remove doc for all(fork, fork, fork) pattern;
* Add explanation for difference between one big all effect and multiple fork effects;
* Add warning for nesting fork in all/race.

The update contains the content from [this comment](https://github.com/redux-saga/redux-saga/issues/1512#issuecomment-401539948).

### DIFF
--- a/docs/advanced/RootSaga.md
+++ b/docs/advanced/RootSaga.md
@@ -8,35 +8,50 @@ export default function* rootSaga() {
     helloSaga(),
     watchIncrementAsync()
   ])
+  // code after all-effect
 }
 ```
 
-This is really just one of a few ways to implement your root. Here, the `all` effect is used with an array of your sagas in a non-blocking fashion. Other root implementations may help you better handle errors and more complex data flow.
+This is one of a few ways to implement your root. Here, the `all` effect is used with an array and your sagas will be executed in parallel. Other root implementations may help you better handle errors and more complex data flow.
 
-Contributor @slorber mentioned in issue #760 several other common root implementations. To start, there are some other implementations you may see that will behave similarly to the  tutorial root saga behavior:
-
-```javascript
-export default function* root() {
-  yield all([
-    fork(saga1),
-    fork(saga2),
-    fork(saga3)
-  ]);
-}
-```
-or 
+Contributor @slorber mentioned in [issue#760](https://github.com/redux-saga/redux-saga/issues/760) several other common root implementations. To start, there is one popular implementation that behaves similarly to the tutorial root saga behavior:
 
 ```javascript
 export default function* root() {
   yield fork(saga1)
   yield fork(saga2)
   yield fork(saga3)
+  // code after fork-effect
 }
 ```
 
-`all (fork fork fork)` will return a single effect, while using three unique `yield fork` will give back a fork effect three times. Ultimately the resulting behavior in your app is the same: all of your sub-sagas are started and executed in the same order. `fork` is non-blocking and so allows the `rootSaga` in these two cases to finish while the child sagas are kept running and blocked by their internal effects.
+Using three unique `yield fork` will give back a task descriptor three times. The resulting behavior in your app is that all of your sub-sagas are started and executed in the same order. `fork` is non-blocking and so allows the `rootSaga` in these two cases to finish while the child sagas are kept running and blocked by their internal effects.
 
-Error handling in all three implementations is the same. Any error will terminate the root saga and subsequently all other children. (`fork` effects are still connected to their parent.)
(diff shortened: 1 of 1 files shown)


docs/advanced/RootSaga.md
```diff
@@ -8,35 +8,50 @@ export default function* rootSaga() {
...
+
+The difference between one big all effect and several fork effects are that all effect is blocking, so *code after all-effect* (see comments in above code) is executed after all the children sagas completes, while fork effects are non-blocking and *code after fork-effect* gets executed right after yielding fork effects. Another difference is that you can get task descriptors when using fork effects, so in the subsequent code you can cancel/join the forked task via task descriptors.
+
+## Avoid nesting fork effects in all effect
+
+Nesting fork effects in all effect is not a good pattern to implement sagas. It may mislead us to think that the fork effects are managed by the call effect, and the call effect will be resolved after all the forked tasks completes. But actually, the fork effects are always connected to the parent task through the underlying forkQueue. Call effect will be resolved immediately since all fork effects are non-blocking. Errors from the forked tasks bubble to the parent task rather than to the call effect.
```
restrry: imho not clear how call is related to fork here. you mean that ppl could think that `all` effect behaves as `call`?

shinima: My mistake. It should be *All effect*.

docs/advanced/RootSaga.md
```diff
@@ -8,35 +8,50 @@ export default function* rootSaga() {
...
+Using three unique `yield fork` will give back a task descriptor three times. The resulting behavior in your app is that all of your sub-sagas are started and executed in the same order. `fork` is non-blocking and so allows the `rootSaga` in these two cases to finish while the child sagas are kept running and blocked by their internal effects.
+
+The difference between one big all effect and several fork effects are that all effect is blocking, so *code after all-effect* (see comments in above code) is executed after all the children sagas completes, while fork effects are non-blocking and *code after fork-effect* gets executed right after yielding fork effects. Another difference is that you can get task descriptors when using fork effects, so in the subsequent code you can cancel/join the forked task via task descriptors.
```
Andarist: all those things apply to the `yield all([ fork(saga1), fork(saga2), fork(saga3) ])` case - by yielding that the code after this all effect should be executed **immediately** because each fork effect is synchronously returning something (a task descriptor) - and ofc those descriptors would be very much usable in the calling saga, like this:
```js
const [task1, task2, task3] = yield all([ fork(saga1), fork(saga2), fork(saga3) ])
```

docs/advanced/RootSaga.md
```diff
@@ -8,35 +8,50 @@ export default function* rootSaga() {
...
+  fork(someSaga),
+  take('SOME-ACTION'),
+  somePromise,
+])
+```
+
+On the other hand, fork effects in a race effect is most likely a bug. In the following code, since fork effects are non-blocking, they will always win the race immediately.
 
-Error handling in all three implementations is the same. Any error will terminate the root saga and subsequently all other children. (`fork` effects are still connected to their parent.)
+The rule of thumb is that **yield fork effects directly from your generator**, do not nest it in all/race effects.
```
Andarist: that's true for race, using fork in a race is pointless, but imho `all([fork, fork])` is very much OK for the reasons mentioned in the previous comment

docs/advanced/RootSaga.md
```diff
@@ -8,35 +8,50 @@ export default function* rootSaga() {
...
+The difference between one big all effect and several fork effects are that all effect is blocking, so *code after all-effect* (see comments in above code) is executed after all the children sagas completes, while fork effects are non-blocking and *code after fork-effect* gets executed right after yielding fork effects. Another difference is that you can get task descriptors when using fork effects, so in the subsequent code you can cancel/join the forked task via task descriptors.
+
+## Avoid nesting fork effects in all effect
+
+```javascript
+// DO NOT DO THIS. Fork effects in all effect do not make sense.
+yield all([ fork(saga1), fork(saga2), fork(saga3) ])
+```
+
+Nesting fork effects in all effect is not a good pattern to implement sagas. It may mislead us to think that the fork effects are managed by the all effect, and the all effect will be resolved after all the forked tasks completes. But actually, the fork effects are always connected to the parent task through the underlying forkQueue. All effect will be resolved immediately since all fork effects are non-blocking. Errors from the forked tasks bubble to the parent task rather than to the all effect.
```
Andarist: this is a good explanation overall, but imho should be purely informative, without advising not do this, because the pattern itself is OK

shinima: ok, then I'll make some changes in another PR

PR #1584, opened 2018-09-04
redux-saga/symbols: use String as symbols

### LINKED ISSUE #1559: What is the purpose of using Symbol instead of String in v1.0?
### Description

Hello,

When I mock a module with Jest, the effects created in this module are not recognized by redux-saga and failed all the tests.

You can checkout the second test in this repo:
https://github.com/dangh/test-jest-saga-repro

I trace down to this line https://github.com/redux-saga/redux-saga/blob/29fbd1b50d7128d3b9e4ff4105ae3b84ed2a2905/packages/core/src/internal/io.js#L217

and figured out that I have to re-require all my modules as well as redux-saga so `runSaga()` can recognize `IO` symbol.

It makes the tests become verbose with lots of redundant requires. Is there's any reason we use Symbol instead of string?

shinima:
| Q                        | A   |
| Fixed Issues?            |  fix #1559   |

Use String as symbols in redux-saga/symbols per https://github.com/redux-saga/redux-saga/issues/1559#issuecomment-418195608

I'm on the fence between using String or Symbol. If we are going to stick with Symbol, feel free to close this PR.


Andarist:
The babel plugin needs to be updated as well, cc @restrry - what do you think about it? seems like strings should be "just fine" for the `SAGA_LOCATION` too, right?

restrry:
@Andarist Unfortunately, no. And we didn't have tests for it. I'll add some integration tests. Where should we keep them? In `core` or `babel-plugin` package?

Andarist:
I would say integration tests for this should be kept in the core package - theoretically the babel plugin only aids the metadata being generated for us, but those things could come from any source. The more important thing is if the core package can read those than if babel plugin generates them correctly (the latter being covered by snapshot tests in the plugin package)

PR #1590, opened 2018-09-06
Update deps

### LINKED ISSUE #1459: getContext doesn't work with Symbol

Andarist:
fixes #1462 fixes #1459

While upgrading the deps I've also started using babel runtime helpers - it allows for deduplication of those helpers across all packages that use them. 

I've also used existing `_extends` helper to replace our `object.assign`, as it was only used for setContext effect and similar I've replaced it altogether with `assignWithSymbols` which reuses babel's `_extends`, but additionally handles also symbols.

Because I refer to `_extends` manually in the code I had to add a small babel plugin to replace requested path for cjs output.


Andarist:
I actually want `--ignore-scripts` behaviour when bootstrapping - otherwise lerna will run i.e. prepare scripts in each package.

PR #1597, opened 2018-09-09
Use a proxy cjs file to require dev & prod bundles conditionally as p…

Andarist:
This is what react does 😉 As saga might be used for SSRing it's better to read process.env single time

### DIFF
--- a/packages/core/rollup.config.js
+++ b/packages/core/rollup.config.js
@@ -81,22 +81,41 @@ const createConfig = ({ input, output, external, env, min = false, useESModules 
   ].filter(Boolean),
 })
 
+const multiInput = {
+  core: 'src/index.js',
+  effects: 'src/effects.js',
+  utils: 'src/utils.js',
+}
+
 export default [
-  ...['esm', 'cjs'].map(format =>
-    createConfig({
-      input: {
-        core: 'src/index.js',
-        effects: 'src/effects.js',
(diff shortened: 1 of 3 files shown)


packages/core/rollup.config.js
```diff
@@ -81,22 +81,41 @@ const createConfig = ({ input, output, external, env, min = false, useESModules
...
+    },
+  }),
+  createConfig({
+    input: multiInput,
+    output: {
+      dir: 'dist',
+      format: 'cjs',
+      entryFileNames: 'redux-saga-[name].prod.[format].js',
+    },
+    useESModules: false,
```
shinima: How about setting the default value of `useESModules` to `false`, or just use `const useESModules = output.format !== 'cjs'` in `createConfig()`?  Something it is a little confusing to make a boolean option default to true.

Andarist: My logic behind using `true` as default is that it's easier to catch a mistake with it (in this case). 

If we use `useESModules: true` by mistake in cjs environment it will fail loudly (even if after releasing a broken version) and will give us a chance to fix it quickly.

If we use `useESModules: false` in a wrong way we might not know that at all, because most commonly ESM environments (webpack, rollup) are configured to consume CJS as well.

`useESModules = output.format !== 'cjs'` - we can do that though

PR #1605, opened 2018-09-14
Unify how getFnCallDescriptor is handled

Andarist:


packages/core/src/internal/io.js
```diff
@@ -71,46 +71,51 @@ export function race(effects) {
...
   if (process.env.NODE_ENV !== 'production') {
-    check(fn, is.notUndef, `${meth}: argument fn is undefined`)
+    check(fnDescriptor, is.notUndef, `${callFnEffects.join('/')}: argument fn is undefined`)
   }
 
   let context = null
-  if (is.array(fn)) {
-    ;[context, fn] = fn
-  } else if (fn.fn) {
-    ;({ context, fn } = fn)
```
shinima: Are we going to drop the support for `{ context, fn }`?

packages/core/src/internal/io.js
```diff
@@ -71,46 +71,51 @@ export function race(effects) {
   return makeEffect(effectTypes.RACE, effects)
 }
 
-function getFnCallDesc(meth, fn, args) {
+const callFnEffects = ['call', 'apply', 'cps', 'fork', 'spawn']
+
+function getFnCallDescriptor(fnDescriptor, args) {
   if (process.env.NODE_ENV !== 'production') {
-    check(fn, is.notUndef, `${meth}: argument fn is undefined`)
+    check(fnDescriptor, is.notUndef, `${callFnEffects.join('/')}: argument fn is undefined`)
```
restrry: imho message  `call/apply/cps/fork/spawn: argument fn is undefined` seems confusing as user calls a method, not all of them

Andarist: right, gonna move validation to the effect creators from get `getFnCallDescriptor` itself

packages/core/src/internal/io.js
Andarist: Oh, this is a bummer 😅 had no idea this is publicly supported, gonna have to restore it & write a test for this

Andarist:
I've addressed both your concerns

PR #1606, opened 2018-09-15
Fixed memory leak for closed actionChannels

Andarist:
I don't think this can be tested easily - from the tests we can't check if there is a leak or not as it is observable only internally.

### DIFF
--- a/packages/core/src/internal/channel.js
+++ b/packages/core/src/internal/channel.js
@@ -8,6 +8,7 @@ import * as matchers from './matcher'
 export const END = { type: CHANNEL_END_TYPE }
 export const isEnd = a => a && a.type === CHANNEL_END_TYPE
 
+const CLOSED_CHANNEL_WITH_TAKERS = 'Cannot have a closed channel with pending takers'
 const INVALID_BUFFER = 'invalid buffer passed to channel factory function'
 const UNDEFINED_INPUT_ERROR = `Saga or channel was provided with an undefined action
 Hints:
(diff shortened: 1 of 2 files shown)


packages/core/src/internal/channel.js
```diff
@@ -183,14 +194,20 @@ export function multicastChannel() {
       }
 
       const takers = (currentTakers = nextTakers)
-      takers.forEach(taker => {
+
+      for (let i = 0, len = takers.length; i < len; i++) {
```
Andarist: this is a hot path, so a for loop should be better here, perf-wise

shinima: ok, then let's add a comment that the for-loop is intentional

packages/core/src/internal/channel.js
```diff
@@ -116,15 +121,15 @@ export function eventChannel(subscribe, buffer = buffers.none()) {
 
   unsubscribe = subscribe(input => {
     if (isEnd(input)) {
-      close()
       closed = true
```
shinima: How about moving this statement (`closed = true`) inside `close()`?  

`multicastChannel()` and `channel()` have this statement inside their `close()` too.

packages/core/src/internal/channel.js
```diff
@@ -80,17 +80,22 @@ export function channel(buffer = buffers.expanding()) {
...
-          const taker = arr[i]
-          taker(END)
-        }
-      }
+    if (process.env.NODE_ENV !== 'production') {
+      checkForbiddenStates()
+    }
+
+    if (closed) {
+      return
```
shinima: 👍  returning early makes code clearer

restrry:
@Andarist I thought that jest could detect memory leaks. doesn't it?

Andarist:
no idea, can't find it in their docs, not sure how reliable the feature would be - especially for case like this, this was rather small memory leak

PR #1613, opened 2018-09-25
docs: add list of changes for v1

restrry:
probably that shouldn't be a separate document in repo, but a part of release notes.


Andarist:
I guess this can be CHANGELOG.md that we would add to the docs

restrry:
@Andarist @shinima ping

PR #1616, opened 2018-09-27
update docs/api: yield take(...) return an action object

### LINKED ISSUE #1552: [docs] enhance `take` effect doc
### Docs update suggestions

I think, it would be useful to **add explicit statement about the return value of `take` effects family**.

Same as for, say, `fork` effect is stated `returns a Task object.`

Though, it is not difficult to perceive, that `take` returns an action dispatched. It will make docs more complete if we state it explicitly, imho.

Now, it follows from code example, e.g.:

```
while (true) {
    const action = yield take(patternOrChannel)
    yield fork(saga, ...args.concat(action))
  }
```

but still is kind of implicit.

ggarek:
| Fixed Issues?            | fixes #1552   |
| Patch: Bug Fix?          |  only update docs  |
| Minor: New Feature?      |   no |
| Tests Added + Pass?      | No + Yes |

Describe what yielding `take` returns, details in the issue.

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -171,6 +171,8 @@ In the case a Saga is cancelled (either manually or using the provided Effects),
 Creates an Effect description that instructs the middleware to wait for a specified action on the Store.
 The Generator is suspended until an action that matches `pattern` is dispatched.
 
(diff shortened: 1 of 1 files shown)


Andarist:
Thanks!

PR #1628, opened 2018-10-07
Suspend the scheduler when starting a root sagas

### LINKED ISSUE #1623: Suspend before running root
root should be considered a regular task (its just being run from outside instead of being forked inside) and thus normal scheduling should apply for it.

So we should suspend the scheduler before running, I can dig up a test case for this as I had it somewhere

Andarist:
fixes #1623

### DIFF
--- a/packages/core/__tests__/interpreter/takeSync.js
+++ b/packages/core/__tests__/interpreter/takeSync.js
@@ -1,7 +1,7 @@
 /* eslint-disable no-unused-vars, no-constant-condition */
 import { createStore, applyMiddleware } from 'redux'
 import sagaMiddleware, { END } from '../../src'
-import { take, put, fork, join, call, all, race, cancel, takeEvery } from '../../src/effects'
+import { take, put, fork, join, call, all, race, cancel, takeEvery, delay } from '../../src/effects'
 import { channel, buffers } from '../../src'
 
 test('synchronous sequential takes', () => {
@@ -679,4 +679,29 @@ test('action dispatched in root saga should get scheduled and taken by a "siblin
   return Promise.resolve().then(() => {
     expect(store.getState()).toEqual(['FIRST', 'SECOND'])
   })
+})
+
+test('action dispatched synchronously in forked task should be taken a following sync take', () => {
+  const actual = []
+  const reducer = (state, action) => action.type
+
+  const middleware = sagaMiddleware()
+  const store = createStore(reducer, applyMiddleware(middleware))
+
+  function* root() {
+    // force async, otherwise sync root startup prevents this from being tested appropriately
+    // as the scheduler is in suspended state because of it
+    yield delay(10)
+    yield fork(function*() {
+      yield put({ type: 'A', payload: 'foo' })
+    })
+    actual.push(yield take('A'))
+  }
+
+  return middleware
+    .run(root)
+    .toPromise()
+    .then(() => {
+      expect(actual).toEqual([{ type: 'A', payload: 'foo' }])
+    })
 })
--- a/packages/core/__tests__/interpreter/takeSync.js
+++ b/packages/core/__tests__/interpreter/takeSync.js
@@ -3,6 +3,7 @@ import { createStore, applyMiddleware } from 'redux'
 import sagaMiddleware, { END } from '../../src'
 import { take, put, fork, join, call, all, race, cancel, takeEvery } from '../../src/effects'
 import { channel, buffers } from '../../src'
+
 test('synchronous sequential takes', () => {
   const actual = []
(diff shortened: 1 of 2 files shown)


packages/core/__tests__/interpreter/takeSync.js
```diff
@@ -643,3 +654,29 @@ test('put causing sync dispatch response in store subscriber', () => {
...
+    return state.concat(action.type)
+  }
+
+  const middleware = sagaMiddleware()
+  const store = createStore(reducer, applyMiddleware(middleware))
+
+  function* root() {
+    yield all([
+      put({ type: 'FIRST' }),
+      takeEvery('FIRST', function*() {
```
restrry: what about such case, where we use use `proc` internally for iterable
```js
function* root() {
    yield all([
      put({ type: 'FIRST' }),
      function*() {
        yield take('FIRST')
        yield put({ type: 'SECOND' })
      },
    ])
  }

```

Andarist: aren't they the same? `takeEvery` is also using proc internally, as it's yielding `fork` which is wrapped with proc.

packages/core/src/internal/runSaga.js
```diff
@@ -85,11 +86,14 @@ export function runSaga(options, saga, ...args) {
     finalizeRunEffect,
   }
 
+  suspend()
   const task = proc(env, iterator, context, effectId, getMetaInfo(saga), null)
```
shinima: Could we move `suspend()` and `flush()` into `proc` as follows?

```javascript
export default function proc(env, iterator, parentContext, parentEffectId, meta, cont) {
  suspend()
  // other code...
  flush()
  return task
```

If we do this, we just need write `suspend()` and `flush()` **one time**. Not sure this idea would work, but tests are passing in my PC.

shinima: Further more, if we do like this, we may export a function `wrap` (or some other better name) from *packages/core/src/internal/scheduler.js* and use it as a decorator on `proc()`.

```javascript
export function wrap(fn) {
  return (...args) => {
    try {
      suspend()
      return fn(...args)
    } finally {
      flush()
    }
  }
}
```

packages/core/__tests__/interpreter/takeSync.js
restrry: there is no `takeEvery`, if you add this test you will see that it fails

Andarist: good catch! should be handled as well

packages/core/src/internal/runSaga.js
Andarist: would prefer avoiding additional closures in the saga code, so Im reluctant to using decorator approach

and Im not sure if we can move this safely into the proc, as `runForkEffect` would *slightly* change the order of how things happen (when the cb callback gets invoked) - I will experiment with it though

Andarist: It seems that tests are passing if we move this to proc, I'll try to do further experimentation to see if I can break some case

packages/core/__tests__/interpreter/takeSync.js
Andarist: actually ur given snippet doesnt work, as it yields a generator and not the iterator, even if we correct it to:
```js
function* root() {
    yield all([
      put({ type: 'FIRST' }),
      (function*() {
        yield take('FIRST')
        yield put({ type: 'SECOND' })
      })(),
    ])
  }
```

the second yielded item in all is equivalent to `call` effect (kinda) and thus scheduler doesn't apply to it

or did I miss something?

Andarist:
I've added a test and wrapped root saga's proc in try/finally - in the same way runForkEffect is already, to keep the scheduler's state more safe.

@shinima can't move suspend+flush into proc, because proc is also used for call effects which are not scheduler and because flush has to happen in runForkEffect AFTER the the task descriptor gets returned to the caller and not before that

packages/core/__tests__/interpreter/takeSync.js
```diff
@@ -643,3 +654,54 @@ test('put causing sync dispatch response in store subscriber', () => {
...
+  const actual = []
+  const reducer = (state, action) => action.type
+
+  const middleware = sagaMiddleware()
+  const store = createStore(reducer, applyMiddleware(middleware))
+
+  function* root() {
+    // force async, otherwise sync root startup prevents this from being tested appropriately
+    // as the scheduler is in suspended state because of it
+    yield delay(10)
```
shinima: We can replace it with `yield Promise.resolve()` for faster execution.

Issue #1592, opened 2018-09-06
Saga stopped inside loop without throwing any error


joes-code:
### Steps to reproduce
I've created a small repo and updated readme with expected and actual outcome.
https://github.com/joes-code/redux-saga-beginner-tutorial

#### Description of the bug/issue
Inside the saga function, I'm simply looping and adding numbers. Depends on the loop size and the browser you use, saga might finish the process and work as expected, or it might just stop half way and that's it. 

#### Steps to reproduce the bug/issue

0. Use Google Chrome
1. npm install
2. npm start
3. Follow https://github.com/joes-code/redux-saga-beginner-tutorial#issue

#### Example
https://github.com/joes-code/redux-saga-beginner-tutorial

### Actual results
Hello Sagas!
loop # 0
0 "-" 405450
loop # 1
1 "-" 405450
loop # 2

### The Expected results
Hello Sagas!
loop # 0 
0 "-" 405450 
loop # 1 
1 "-" 405450 
loop # 2 
2 "-" 405450 
loop # 3 
3 "-" 405450 
loop # 4 
4 "-" 405450  

### Environment information

- latest redux-saga version (please see package.json from my sample repo for more information)
- Google Chrome 68.0.3440.106 (Official Build) (64-bit)
- Mac

Andarist:
Thanks for the report, I'm planning to look into the issue - to see if we can avoid stack overflow when processing a large amount of sync consecutive effects.

joes-code:
Thanks @shinima @Andarist for confirming the exception. Yes it would be great if the exception didn't get swallowed or even avoid it altogether. In my real project, I was able to get around this by pretty much removing the yield inside loop, which might freeze the UI if the loop is big, but shouldn't be a problem in my case.

restrry:
I checked our source code and it seems that saga catches an error [here](https://github.com/redux-saga/redux-saga/blob/master/packages/core/src/internal/proc.js#L262-L266)
but because it handles an error in another function call https://github.com/redux-saga/redux-saga/blob/master/packages/core/src/internal/proc.js#L266 , browser engine (Chrome) ignores that expression, so error remains unhandled 
how can we fix that?
- we cannot make the whole process async, as it breaks current contract
- TCO is supported only is Safari
- switch from recursion to an iteration. that could require significant code changes

Andarist:
I would like to explore this:
> switch from recursion to an iteration - probably, but that could require significant code changes

[...]

PR #1629, opened 2018-10-07
Remove try/finally from runForkEffect

Andarist:
I'm not sure if we should merge it - it's just an idea. 

Tests seem to pass, I wonder if there is any downside of removing this. Obviously previous version (with finally) was swallowing errors, but I wonder if there is even any situation when this part of the code could throw (except stack exceeded errors) - I would assume errors are caught elsewhere & handled through continuations.


Andarist:
TIL, the issue with stack limit exceeded is weird then as the error gets swallowed somewhere, but I had hard time debugging this because devtools became quite unresponsive with such deep call stacks :s

restrry:
I saw similar problem debugging https://github.com/redux-saga/redux-saga/issues/1592
probably it's not the same thing, but - when `max error call stack...` error bubbles to catch statement, it was caught, **but** if I called another function to handle an error **sometimes** it was ignored (depending on code sample).

PR #1636, opened 2018-10-15
update docs for v1 api

restrry:
review pls, if I missed or misinterpreted something
@shinima, @Andarist

### DIFF
--- a/docs/advanced/UsingRunSaga.md
+++ b/docs/advanced/UsingRunSaga.md
@@ -7,20 +7,29 @@ When a Saga is started (either at startup or later dynamically), the middleware 
 `redux-saga` provides a way to run a Saga outside of the Redux middleware environment and connect it to a custom Input/Output.
 
 ```javascript
-import { runSaga } from 'redux-saga'
+import { runSaga, stdChannel } from 'redux-saga'
 
-function* saga() { ... }
+const emitter = new EventEmitter();
+const channel = stdChannel();
+emitter.on("action", channel.put);
 
 const myIO = {
-  subscribe: ..., // this will be used to resolve take Effects
-  dispatch: ...,  // this will be used to resolve put Effects
-  getState: ...,  // this will be used to resolve select Effects
+  // this will be used to orchestrate take and put Effects
+  channel,
+  // this will be used to resolve put Effects
+  dispatch(output) {
+    emitter.emit("action", output);
+  },
+  // this will be used to resolve select Effects
+  getState() {
+    return state
+  }
 }
 
 runSaga(
   myIO,
-  saga,
+  function* saga() { ... },
 )
 ```
 
-For more info, see the [API docs](https://redux-saga.js.org/docs/api/index.html##runsagaoptions-saga-args).
+For more info, see the [API docs](https://redux-saga.js.org/docs/api/index.html##runsagaoptions-saga-args), [Channels](./Channels.md), [demo](https://codesandbox.io/s/1yq1lx77jq)
--- a/docs/advanced/UsingRunSaga.md
+++ b/docs/advanced/UsingRunSaga.md
@@ -6,19 +6,19 @@ When a Saga is started (either at startup or later dynamically), the middleware 
 
 `redux-saga` provides a way to run a Saga outside of the Redux middleware environment and connect it to a custom Input/Output.
 
-```javascript
+```js
 import { runSaga, stdChannel } from 'redux-saga'
 
-const emitter = new EventEmitter();
-const channel = stdChannel();
-emitter.on("action", channel.put);
+const emitter = new EventEmitter()
+const channel = stdChannel()
+emitter.on("action", channel.put)
 
 const myIO = {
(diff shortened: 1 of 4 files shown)


docs/recipes/README.md
```diff
@@ -19,7 +19,9 @@ function* watchInput() {
 By using this helper the `watchInput` won't start a new `handleInput` task for 500ms, but in the same time it will still be accepting the latest `INPUT_CHANGED` actions into its underlaying `buffer`, so it'll miss all `INPUT_CHANGED` actions happening in-between. This ensures that the Saga will take at most one `INPUT_CHANGED` action during each period of 500ms and still be able to process trailing action.
 
 ## Debouncing
+From redux-saga@v1 [debounce](../api/README.md#debouncems-pattern-saga-args) is built-in effect.
 
+For `redux-saga@0.x`:
```
shinima: Should we mention v0.x here? 😕 

I think the target readers are who uses v1.0, because other parts of  documentation is reflecting the v1.0 version and does not mention v0.x at all.

docs/advanced/UsingRunSaga.md
```diff
@@ -7,20 +7,29 @@ When a Saga is started (either at startup or later dynamically), the middleware
 `redux-saga` provides a way to run a Saga outside of the Redux middleware environment and connect it to a custom Input/Output.
 
 ```javascript
-import { runSaga } from 'redux-saga'
+import { runSaga, stdChannel } from 'redux-saga'
 
-function* saga() { ... }
+const emitter = new EventEmitter();
```
shinima: let's remove the semicolons here 😃 . We've set `semi: false` in our prettier config.

docs/recipes/README.md
restrry: sound legit

docs/api/README.md
```diff
@@ -78,31 +78,16 @@
...
-        emit(action);
-       }
-     });
-     ```
-
   - `logger` : Function -  defines a custom logger for the middleware. By default, the middleware logs all errors and
 warnings to the console. This option tells the middleware to send errors/warnings to the provided logger instead. The logger is called with the params `(level, ...args)`. The 1st indicates the level of the log ('info', 'warning' or 'error'). The rest corresponds to the following arguments (You can use `args.join(' ')` to concatenate all args into a single String).
 
   - `onError` : Function - if provided, the middleware will call it with uncaught errors from Sagas. useful for sending uncaught exceptions to error tracking services.
+  - `effectMiddlewares` : Function [] - allows you to intercept any effect, resolve it on your own and pass to the next middleware.
```
shinima: Here we could add a link to the testing docs which contains a detailed example of `effectMiddlewares`.

We could add this line: `See [this section](/docs/advanced/Testing.md#effectmiddlwares) for a detailed example.`

docs/recipes/README.md
restrry: should I remove the whole description how `retry` is implemented under-the-hood? imho it's quite useful for understanding the saga principles on the examples

docs/advanced/UsingRunSaga.md
Andarist: we should tweak out lintstaged config to format .md files too

docs/advanced/UsingRunSaga.md
```diff
@@ -7,20 +7,29 @@ When a Saga is started (either at startup or later dynamically), the middleware
...
 
 runSaga(
   myIO,
-  saga,
+  function* saga() { ... },
 )
 ```
 
-For more info, see the [API docs](https://redux-saga.js.org/docs/api/index.html##runsagaoptions-saga-args).
+For more info, see the [API docs](https://redux-saga.js.org/docs/api/index.html##runsagaoptions-saga-args), [Channels](./Channels.md), [demo](https://codesandbox.io/s/1yq1lx77jq)
```
Andarist: codesandbox example is cool, we could make it an official example in the example directory and then we could point to it here with url like this:
https://codesandbox.io/s/github/redux-saga/redux-saga/tree/master/examples/run-saga-demo

docs/recipes/README.md
Andarist: > should I remove the whole description how retry is implemented under-the-hood? 

no, i think (as you) that its helpful for the readers

docs/advanced/UsingRunSaga.md
restrry: 🚀 will experiment with it

restrry: I's suggest to make in another PR.
I made it work with [changes](https://github.com/redux-saga/redux-saga/compare/update-docs...add-codesanbox-examples), just need to resolve some problems:
1 codesanbox doesn't install `devDependencies` https://github.com/CompuIves/codesandbox-client/issues/924 so I had to move all deps under `dependencies` field. 
2 `sagaMonitor` package is not published. we can switch to https://github.com/redux-saga/redux-saga-devtools or publish one from the example folder

restrry:
addressed comments, pls have a look again

PR #1637, opened 2018-10-15
Use task.root to track whether a task is a root task or a detached task.

shinima:
| Tests Added + Pass?      | |

We used `task.cont == null` to judge whether a task is root/detached (then we build the sagaStack for root/detached tasks when these task aborts). This was a little confusing, so in this PR, I add a new field `isRoot` in `Task` object, and `task.cont` won't be null any more.

### DIFF
--- a/packages/core/src/internal/proc.js
+++ b/packages/core/src/internal/proc.js
@@ -420,7 +420,7 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
   }
 
   function resolveIterator(iterator, effectId, meta, cb) {
-    proc(env, iterator, taskContext, effectId, meta, false, cb)
+    proc(env, iterator, taskContext, effectId, meta, /* isRoot */ false, cb)
   }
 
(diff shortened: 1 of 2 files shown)


packages/core/src/internal/proc.js
```diff
@@ -420,7 +420,7 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
   }
 
   function resolveIterator(iterator, effectId, meta, cb) {
-    proc(env, iterator, taskContext, effectId, meta, cb)
+    proc(env, iterator, taskContext, effectId, meta, false, cb)
```
restrry: imho already hard to read. could we extract in a separate variable?
```js
const isRoot = false
proc(env, iterator, taskContext, effectId, meta, isRoot, cb)
```

shinima: okay, I'll add an inline comment here.

```javascript
proc(env, iterator, taskContext, effectId, meta, /* isRoot */ false, cb)
```

PR #1639, opened 2018-10-15
Issue#1638

### LINKED ISSUE #1638: Delay type in api.md not equal real function
### Steps to reproduce

1) Show [delay type](https://github.com/redux-saga/redux-saga/tree/master/docs/api#delayms-val) effect
2) And compare with `delayP` function
https://github.com/redux-saga/redux-saga/blob/a1c465d18780bc69450569b76f43e4e35baac5c7/packages/delay-p/src/index.js#L3-L14
3) It is always return true

Where is the error in the documentation or in the implementation of the function `delayP` ?

#### Example
```js
const test = yield delay(1, "myValue"); // true
```
### Actual results
`test === true`

### The Expected results
```js
test === "myValue" // if the documentation is the source of truth
```

### Environment information
- redux-saga 1.0.0-beta.2
- babel

retyui:
| Fixed Issues?            | Fixes #1638|
| Minor: New Feature?      |   No  |

### DIFF
--- a/packages/delay-p/src/index.js
+++ b/packages/delay-p/src/index.js
@@ -1,9 +1,9 @@
 import { CANCEL } from '@redux-saga/symbols'
 
-export default function delayP(ms) {
+export default function delayP(ms, returnValue = true) {
   let timeoutId
   const promise = new Promise(resolve => {
-    timeoutId = setTimeout(resolve, ms, true)
+    timeoutId = setTimeout(resolve, ms, returnValue)
   })
 
   promise[CANCEL] = () => {
--- a/packages/core/__tests__/sagaHelpers/delay.js
+++ b/packages/core/__tests__/sagaHelpers/delay.js
@@ -3,7 +3,7 @@ import { createStore, applyMiddleware } from 'redux'
 import { delay } from '../../src/effects'
 import delayP from '@redux-saga/delay-p'
 
-test('delay', () => {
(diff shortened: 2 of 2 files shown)


packages/delay-p/src/index.js
```diff
@@ -1,9 +1,9 @@
 import { CANCEL } from '@redux-saga/symbols'
 
-export default function delayP(ms) {
+export default function delayP(ms, returnValue = true) {
```
restrry: imho shouldn't have default value

shinima: I think the default value is fine. When we do not care the return value, we can omit the `returnValue`. And the default value (`true`) is useful when we use `delayP` in a race:

```javascript
const result= race({
  timeout: call(delayP, 2000),
  response: call(fetch, url),
})
// result.timeout is true if delayP resolves first
if (result.timeout) { /* ... */ }
else { /* ... */ }
```

shinima: The behavior now is the same as in v0.16.1. I think this is a small mistake when separating internal modules. 
https://github.com/redux-saga/redux-saga/blob/600802fe2c91be57c6d0c659424b9844739b5379/src/internal/utils.js#L94-L103

retyui:
@restrry Is it ready to merge or do you have any question?

PR #1648, opened 2018-10-26
Add helpful dev warning about yielding call(delay, ms) incorrectly

Andarist:


packages/core/src/internal/io.js
```diff
@@ -122,8 +122,17 @@ function getFnCallDescriptor(fnDescriptor, args) {
...
+const isNotDelayEffect = fn => fn !== delay
+
 export function call(fnDescriptor, ...args) {
   if (process.env.NODE_ENV !== 'production') {
+    check(
+      fnDescriptor,
+      isNotDelayEffect,
+      `instead of writing \`yield call(delay, ${
+        args[0]
+      })\` where delay is an effect from \`redux-saga/effects\` you should write \`yield delay(${args[0]})\``,
```
restrry: it's nice, but most likely linter of build system will let you know that you try to import delay which is not exported by effect module

Andarist: actually `delay` is an effect now (its just bound call, but still)

PR #1654, opened 2018-10-28
Refactor signature of sagaMiddleware options.

shinima:
| Major: Breaking Change? | See below      |
| Minor: New Feature?     |             |
| Tests Added + Pass?     | 👍  |
| Documentation           | Need to update |

This PR is a subset of #1528. `options.emitter` is removed from `sagaMiddlewareFactory`, instead `options.channel` is added, so users can enhance a channel and pass the channel to the middleware to get the same effect as the `options.emitter`. Note that it is a breaking change.

### DIFF
--- a/packages/core/__tests__/middleware.js
+++ b/packages/core/__tests__/middleware.js
@@ -94,24 +94,57 @@ test('enhance channel.put with an emitter', () => {
   }
 
   function* saga() {
-    yield takeEvery(ac => ac.type.length === 1, function*(ac) {
-      actual.push(ac.type)
-      yield put({ type: `put_${ac.type}` })
+    yield takeEvery(ac => ac.from !== 'saga', function*({ type }) {
+      actual.push({ saga: true, got: type })
+      yield put({ type: `pong_${type}`, from: 'saga' })
     })
+    yield takeEvery(
+      ac => ac.from === 'saga',
+      ({ type }) => {
+        actual.push({ saga: true, got: type })
+      },
+    )
+  }
+
+  let pastStoreCreation = false
+  const rootReducer = (state, { type }) => {
+    if (pastStoreCreation) {
+      actual.push({ reducer: true, got: type })
+    }
+
+    return {}
   }
 
   const middleware = sagaMiddleware({ channel })
-  const store = createStore(() => {}, applyMiddleware(middleware))
+  const store = createStore(rootReducer, {}, applyMiddleware(middleware))
+  pastStoreCreation = true
+
   middleware.run(saga)
   store.dispatch({ type: 'a' })
   store.dispatch({
     type: 'batch',
-    batch: [{ type: 'b' }, { type: 'c' }, { type: 'd' }],
+    batch: [{ type: 'b' }, { type: 'c' }],
   })
-  store.dispatch({ type: 'e' })
+  store.dispatch({ type: 'd' })
 
   // saga must be able to take actions emitted by middleware's custom emitter
-  const expected = ['a', 'put_a', 'b', 'put_b', 'c', 'put_c', 'd', 'put_d', 'e', 'put_e']
+  const expected = [
+    { reducer: true, got: 'a' },
+    { saga: true, got: 'a' },
+    { reducer: true, got: 'pong_a' },
+    { saga: true, got: 'pong_a' },
+    { reducer: true, got: 'batch' },
+    { saga: true, got: 'b' },
+    { reducer: true, got: 'pong_b' },
+    { saga: true, got: 'pong_b' },
+    { saga: true, got: 'c' },
+    { reducer: true, got: 'pong_c' },
+    { saga: true, got: 'pong_c' },
+    { reducer: true, got: 'd' },
+    { saga: true, got: 'd' },
+    { reducer: true, got: 'pong_d' },
+    { saga: true, got: 'pong_d' },
+  ]
   expect(actual).toEqual(expected)
 })
 
--- a/packages/core/__tests__/middleware.js
+++ b/packages/core/__tests__/middleware.js
@@ -1,7 +1,8 @@
 import { createStore, applyMiddleware } from 'redux'
-import sagaMiddleware from '../src'
+import sagaMiddleware, { stdChannel } from '../src'
 import * as is from '@redux-saga/is'
 import { takeEvery } from '../src/effects'
+
 test('middleware output', () => {
   const middleware = sagaMiddleware() // middleware factory must return a function to handle {getState, dispatch}
 
@@ -19,12 +20,14 @@ test('middleware output', () => {
 
   expect(actionHandler.length).toBe(1)
 })
+
 test("middleware's action handler output", () => {
   const action = {}
   const actionHandler = sagaMiddleware()({})(action => action) // action handler must return the result of the next argument
 
   expect(actionHandler(action)).toBe(action)
 })
+
 test('middleware.run', () => {
   let actual
 
@@ -49,6 +52,7 @@ test('middleware.run', () => {
 
   expect(actual).toEqual(expected)
 })
+
 test('middleware options', () => {
   try {
     sagaMiddleware({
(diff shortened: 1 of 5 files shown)


packages/core/src/internal/runSaga.js
```diff
@@ -77,8 +85,7 @@ export function runSaga(options, saga, ...args) {
   }
 
   const env = {
-    stdChannel: channel,
-    dispatch: wrapSagaDispatch(dispatch),
+    channel: { ...channel, put: wrapSagaDispatch(dispatch || channel.put) },
```
shinima: Swap `channel.put` here so that `runPutEffect` just need to call `channel.put` (no need to know the `dispatch` any more).

`dispatch || channel.put` means that we are making the `dispatch` **optional**, when uses do not supply a `dispatch`, users can still use put/take through the underlying `env.channel`.

packages/core/src/internal/io.js
```diff
@@ -52,7 +52,7 @@ export function put(channel, action) {
   }
   if (is.undef(action)) {
     action = channel
-    channel = null
+    channel = undefined
```
Andarist: a short comment above this line that we assign undefined and not null to make default parameter work at later point would be helpful

packages/core/src/internal/middleware.js
```diff
@@ -1,11 +1,10 @@
 import * as is from '@redux-saga/is'
 import { check, assignWithSymbols, createSetContextWarning } from './utils'
 import { stdChannel } from './channel'
-import { identity } from './utils'
 import { runSaga } from './runSaga'
 
-export default function sagaMiddlewareFactory({ context = {}, ...options } = {}) {
-  const { sagaMonitor, logger, onError, effectMiddlewares } = options
+export default function sagaMiddlewareFactory(options = {}) {
+  const { context = {}, channel = stdChannel(), sagaMonitor, logger, onError, effectMiddlewares } = options
```
Andarist: I would inline this destructuring into `sagaMiddlewareFactory` declaration, without it I'm thinking that `options` might be used later on and search for references to it
```js
export default function sagaMiddlewareFactory({
	channel = stdChannel(),
	context = {},
	effectMiddlewares
	logger,
	onError,
	sagaMonitor,
} = {}) {
	// ...
}
```

packages/core/src/internal/runSaga.js
Andarist: doesnt it mean that if somebody would use custom `channel` through the option then his/her action wouldnt get dispatched automatically?

packages/core/__tests__/middleware.js
```diff
@@ -76,47 +80,35 @@ test('middleware options', () => {
...
+  channel.put = action => {
+    if (action.type === 'batch') {
+      action.batch.forEach(rawPut)
+      return
+    }
+    rawPut(action)
+  }
 
   function* saga() {
     yield takeEvery('*', ac => actual.push(ac.type))
```
Andarist: regarding my comment here - https://github.com/redux-saga/redux-saga/pull/1654/files#r228758007 

```js
yield takeEvery('*', function*(ac) {
  actual.push(ac.type)
  yield put({ type: 'saga_action' })
})
```

because of making `dispatch` "optional" you have made this put here not hitting reducers automatically

packages/core/src/internal/runSaga.js
shinima: Let me describe it clearer:

`sagaMiddlewareFactory.options` has now 6 fields (namely `context, channel, sagaMonitor, logger, onError, effectMiddlewares`), and all of them is optional. When connect to a redux store, `sagaMiddlewareFactory` can get `getState` and `dispatch`, and then pass them to `runSaga` along with the above 6 fields.

In cases that users are using `sagaMiddlewareFactory`, `options.dispatch` in `runSaga` is always not null, `put` effects will always hit reducers.

`runSaga.options` has now 8 fields, 2 fields more than `sagaMiddlewareFactory.options`, namely `dispatch` and `getState`.

In cases that users are directly using `runSaga`, `options.dispatch` behaves like a put-handler. **In the past**, users must pass the `dispatch` option, otherwise calls to the `dispatch` will throw an exception in [`runPutEffect`](https://github.com/redux-saga/redux-saga/blob/master/packages/core/src/internal/proc.js#L456). **Now**, `dispatch || channel.put` means that we are making the `dispatch` in `runSaga.options` optional, when uses do not supply a `dispatch`, put/take still works, and users can use it for inter-saga communication.

Andarist: `dispatch` is used in saga for emitting actions outside of its runtime
`channel.put` is used for emitting **into** saga runtime

Having them separate makes a nice distinction, although might be a little bit confusing.

Current solution has other problem though, by making this:
```js
channel: { ...channel, put: wrapSagaDispatch(dispatch || channel.put) },
```

> In cases that users are using sagaMiddlewareFactory, options.dispatch in runSaga is always not null, put effects will always hit reducers.

I've pushed out a change to the test to showcase the problem, you can check out that simply removing `dispatch` is causing a problem there.

packages/core/__tests__/middleware.js
```diff
@@ -108,7 +111,7 @@ test('enhance channel.put with an emitter', () => {
   store.dispatch({ type: 'e' })
 
   // saga must be able to take actions emitted by middleware's custom emitter
-  const expected = ['a', 'b', 'c', 'd', 'e']
+  const expected = ['a', 'put_a', 'b', 'put_b', 'c', 'put_c', 'd', 'put_d', 'e', 'put_e']
```
shinima: 😕  I think the expected result should be `['a', 'b', 'c', 'd', 'e']`, because we are calling `actual.push(ac.type)` only when `ac.type.length === 1`.

If I change the `saga` into...
```javascript
function* saga() {
  yield takeEvery('*', ac => actual.push(ac.type))
  yield takeEvery(ac => ac.type.length === 1, function*(ac) {
    yield put({ type: `put_${ac.type}` })
  })
}
```
..., then the expected result should be `['a', 'put_a', 'b', 'put_b', 'c', 'put_c', 'd', 'put_d', 'e', 'put_e']`.

packages/core/src/internal/runSaga.js
Andarist: > 😕 Looks the expected result is wrong to me. Can you double check that?

You are right, I've made a silly mistake there 😅 

I think there might be edge cases around current solution, but Im not saying it's a wrong direction - we only might need to tweak it a little bit. I need to give this a little bit more thought, gonna experiment today with how it works exactly now and what can and what cannot be done to see if we can follow that direction.

shinima:
Seems current problem is with the following line of code.

```javascript
const env = {
  channel: { ...channel, put: wrapSagaDispatch(dispatch || channel.put) },
  // ...
}
```

Let me try to explain how it works. The following is the influenced code:

```javascript
// previous code
// in env setup
env.channel = channel
// in runPutEffect
function runPutEffect({ channel, action, resolve }, cb) {
  asap(() => {
    let result
    try {
      result = (channel ? channel.put : env.dispatch)(action)
    } catch (error) {
      cb(error, true)
      return
    }
    // ......
  })
}

// current code
// in env setup
env.channel = { ...channel, put: wrapSagaDispatch(dispatch || channel.put) }
// in runPutEffect
function runPutEffect({ channel = env.channel, action, resolve }, cb) {
  asap(() => {
    let result
    try {
      result = channel.put(action)
    } catch (error) {
      cb(error, true)
      return
    }
    // ......
  })
}
```

Image what would happend in the following different cases...

1. When we use `yield put(ch, ac)`, the `channel` passed to `runPutEffect()` will be `ch`, both previous and current code will execute `ch.put(ac)`

2. When we use `yield put(ac)`...
   - In previous code, `channel` passed to `runPutEffect()` is `undefined`, so the code will execute `env.dispatch(ac)`.
   - In current code, it will execute as `env.channel.put(ac)`, but since we have swapped the `env.channel.put` and it refers to `dispatch`, the actual executed code is the same as `dispatch(ac)`.

3. When we call `channel.put(ac)` outside the saga... In current code, `env.channel = { ...channel, put: ... }` does not mutate the channel inplace (it creates a new channel with a new put method), so `channel.put(ac)` will have the same behavior as in the previous code. In both code, `ac` will go through the channel, but it will **NOT** hit reducers.

4. When we `store.dispatch(ac)`... after hitting reducers, `ac` will be captured by sagaMiddleware and it will be send to channel, so the case becomes case-3...

Hope this explanation is readily comprehensible.

---

The idea behind this change is that, by swapping the `channel.put`, we move our logic into the channel and keep the saga runtime simple.  Hope this can be understood and adopted by you. For me, it seems better than the old code, thought actually I'm okay with the old code.

shinima:
@Andarist  any ideas?  If we are unsure about removing `env.dispatch`,  I could keep `env.dispatch` in this PR and then move the change into another PR.

Andarist:
Thanks - you can create a PR with your dispatch -> channel change so we can discuss further, keep in mind though Im not convinced yet that we should follow this path.

PR #1655, opened 2018-10-28
Remove redux-saga/utils

Andarist:
`redux-saga/utils` had now only 4 exports, 2 were symbols, other 2 were testing-related so I've moved them to separate package (thanks @restrry for a suggestion)

I've cleaned up a little bit situation with TS typings too, there are 2 unanswered questions regarding that have to be answered before merging.


packages/testing-utils/src/index.js
```diff
@@ -0,0 +1,36 @@
+import { TASK } from '@redux-saga/symbols'
+
+export const cloneableGenerator = generatorFunc => (...args) => {
```
restrry: we continue keeping tests in the core package?

restrry: should we update docs as well? 
```js
import { cloneableGenerator } from '@redux-saga/testing-utils'
```

packages/is/index.d.ts
```diff
@@ -0,0 +1,21 @@
+// TODO: some of those guards need import types, they would have to be imported from `redux-saga` module
```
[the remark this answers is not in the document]
aikoven: Totally fine, circular imports of types are not a problem.

It's also ok to just copy them

packages/testing-utils/index.d.ts
```diff
@@ -1,51 +1,5 @@
...
-export function arrayOfDeferred<T>(length: number): Deferred<T>[];
-
-interface MockTask extends Task {
-  setRunning(running: boolean): void;
-  setResult(result: any): void;
-  setError(error: any): void;
-}
-
-export function createMockTask(): MockTask;
+// TODO: how this dependency should be expressed? should it be dep? peerDep? or can it be skipped?
```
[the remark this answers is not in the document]
aikoven: Not sure actually. Probably `optional`, since this package can be used without TS

packages/testing-utils/src/index.js
Andarist: no, need to move them too, thanks for reminding about it

Andarist: yeah, gonna do it as part of this PR

Andarist:
@aikoven I've tweaked this PR - decided to create `@redux-saga/types` package to avoid cyclic dependency (i.e. lerna is giving warning when it encounters cycles).

I had to tweak types a little bit - Effect is generic now, old Effect got renamed to SagaEffect and I've introduced AnyEffect for usage in some places. Could u validate that this OK? Unfortunately for some reason it broke all & race tests - I would really appreciate you taking look at this, because I'm unsure why it broke and how to fix it.

I'm also having a small problem with `END` - I think the problem is that it was both a exported as type and as value, I've moved it to types package and tried to just reexport it but it seems that it's not equivalent to the previous situation - not sure how to proceed with this one either.

Andarist:
> So to fix this, you should change SagaIterator from IterableIterator<AnyEffect> to IterableIterator<SagaEffect>.

Damn, I think I understand this. Would it be possible to create something like `EffectCombinator` to account for all & race and define `AnyEffect` differently - so the TS could know that if EffectCombinator is used it should use current `Effect<any, any>` (that would be only a fallback for non-combinator case)? Maybe that's possible with conditional types?

aikoven:
This might be possible.

I can see only two possibilities: define a union of all accepted types, or handle particular cases of `all` and `race` in a special way (via conditional types).

Otherwise, there's no way for TS to know that e.g. yielding a promise inside `all` should fail with strict typings: it will only affect the type of `Effect.payload` and the resulting payload type might be perfectly valid for another effect.

To make it work, we could make a condition on the `T` of `Effect<T, P>`, and if `T extends 'ALL' | 'RACE'`, set a more narrow type for `payload`. Or, if you're ok with changing runtime stuff for the sake of typings, add another field to distinguish combinator from a regular effect.

Andarist:
> To make it work, we could make a condition on the T of Effect<T, P>, and if T extends 'ALL' | 'RACE', set a more narrow type for payload. Or, if you're ok with changing runtime stuff for the sake of typings, add another field to distinguish combinator from a regular effect.

Seems like an additional runtime field wouldnt be that bad of a solution - I would rather do that instead of hardcoding ALL & RACE in "generic" EffectCombinator type.

I'm toying with this in TS online playground but cannot make it work - I'd love some help with that if possible. Im currently stuck on a fact that `SagaIterator` is not parametrized and I need to have a type parameter to use conditional types. This leads me to the conclusion that I might not understand how this all should be tied together in the TS world 🤔

After a second thought - by introducing an additional runtime property we shouldnt be forced to using conditional types, going to try make it work without them. I think I havent fully understood how conditional types work - but from what I understand now they have to be used in combination with generics and I've wanted to use them in non-generic situation

aikoven:
Uh, I'm sorry, I accidentally pushed to this repo, not my fork. I'll explain the fix in the next comment.

aikoven:
So we're trying to distinguish between a valid and invalid effect in types. We've already split `Effect` and `CombinatorEffect`:

```ts
interface Effect<T, P> {
  '@@redux-saga/IO': true;
  combinator: false;
  type: T;
  payload: P;
}

export interface CombinatorEffect<T> {
  '@@redux-saga/IO': true;
  combinator: true;
  type: T;
  payload: CombinedEffects;
}
```

And we assume that everything that's assignable to `Effect` is a valid effect (because we don't want to have a union of all known effects anymore). 

Now `CombinatorEffect` is valid if it combines valid effects. To type check what's inside combinator effect, I've added a type parameter to `CombinedEffects`:

```ts
type CombinedEffects<E> = {[key: string]: E} | E[]
```

Note that this doesn't mean that all combined effects would be of the same type. E.g.:

```ts
const effect = all([call(...), fork(...)]);
// effect.payload would have type:
type P = CombinedEffects<CallEffect | ForkEffect>;
```

So a combinator effect is valid if its payload is assignable to `CombinedEffects<ValidEffect>`, which gives us the type:

```ts
interface AnyCombinatorEffect extends CombinatorEffect<any, ValidEffect> {}
```

The `interface` is needed to allow type recursion, which is impossible with `type` declarations (TS limitation).

Andarist:
THIS IS GREAT ❤️ Thank you so much! Have you also reviewed the rest of this PR? Does those changes make sense to you? 

You should probably also add yourself to the contributors list of the types package (and the core package too) if you want. Your help on those things is invaluable

aikoven:
I have concerns about the naming: the presence of `Effect`, `AnyEffect` and `ValidEffect` might be confusing. As a user I would expect a type named `Effect` to reflect everything that `redux-saga` considers an effect, not just "simple" effects (non-combinators). How about this:

```ts
interface SimpleEffect<T, P> {
  '@@redux-saga/IO': true;
  combinator: false;
  type: T;
  payload: P;
}

interface CombinatorEffect<T, E> {
  '@@redux-saga/IO': true;
  combinator: true;
  type: T;
  payload: CombinedEffects<E>;
}

// default type parameter lets us avoid extra `AnyEffect` type
type Effect<T = any> = SimpleEffect<T, any> | CombinatorEffect<T, any>;
```

For strict checks we could go with:

```ts
interface StrictCombinatorEffect<T> extends CombinatorEffect<T, StrictEffect<T>> {}

type StrictEffect<T = any> = Effect<T> | StrictCombinatorEffect<T>;
```

Also, `CombinedEffects` could be renamed to `CombinatorEffectDescriptor` to be consistent with other effects.

aikoven:
> Does those changes make sense to you?

Everything else looks good to me.

Off-topic: I would like to refactor the typings a bit before major release, but I've been too busy lately. Could you please ping me before it's final? 

Particularly I wanted to copy documentation to the typings, as we did for `redux`, so it's visible in IDE quick help popups. For VS Code users this would even work with non-TS code! 

I also wanted to use generic rest parameters (TS 3.0 feature) for stuff like `call` effects, where we now have multiple overloaded signatures.

Andarist:
I was concerned about naming too - so definitely something that I'd like to improve. This PR in regards to typings was rather a quick hack on my side, so I wasn't paying much attention to the naming as I couldn't figure out better names quickly.

> Off-topic: I would like to refactor the typings a bit before major release, but I've been too busy lately. Could you please ping me before it's final?

Sure, this is one of the last PRs I was going to make before releasing rc, but we got into some other refactors lately which we would like to finish before releasing it - it shouldnt take much time though. Anyway - I think we can still change TS typings during RC time as TS typings are just additional thing that we support, but its not part of the strict core. We can treat them as DefinitelyTyped types which only happens to live in this repo.

Issue #1621, opened 2018-10-02
Unify logError/onError


Andarist:
Having them both is confusing, we should only have one of them.

restrry:
discussable https://github.com/redux-saga/redux-saga/pull/1546#discussion_r222030376

PR #1658, opened 2018-10-30
Remove logError parameter

restrry:
| Fixed Issues?            | fix https://github.com/redux-saga/redux-saga/issues/1621 |
| Major: Breaking Change?  |  Yes   |

1 don't ignore error during cancellation, allow it to bubble
2 remove `logError` parameter


packages/core/src/internal/proc.js
```diff
@@ -263,9 +263,6 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
         mainTask.cont(result.value)
       }
     } catch (error) {
-      if (mainTask._isCancelled) {
```
restrry: jfyi was added https://github.com/redux-saga/redux-saga/commit/2e53f31d40d79eed2d776c17a1a1a4b53be0fe7d

packages/core/__tests__/interpreter/cancellation.js
```diff
@@ -738,3 +738,29 @@ test('cancel should support for self cancellation', () => {
...
+  function* worker() {
+    const taskA = yield io.fork(child);
+    yield io.delay(100)
+    yield io.cancel(taskA);
+  }
+
+  return middleware.run(worker)
+    .toPromise()
+    .catch((error) => {
+      expect(error).toEqual(expectedError)
```
shinima: Actually this promise will resolves to `undefined`, so the catch-callback does not run and the test passes.

If you change this to ...
```javascript
const promise = middleware.run(worker).toPromise()
return expect(promise).rejects.toBe(expectedError)
```
... then the test will fail.

packages/core/src/internal/proc.js
```diff
@@ -263,9 +263,6 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
...
-        env.logError(error)
```
shinima: This should not be removed, or else the error will be swallowed...
If you log `mainTask.cont` in this if-statement, it will print as `noop`, so the error object will be sent to `noop`.  
JYI, `mainTask.cont` is set to `noop` by `forkQueue` when we cancel the task.

restrry: yes, I was going to throw here, but forgot as tests 'passed'

shinima:
Let me think aloud the current behavior of exception during cancellation as far as I understood. In the following code, the parent is going to cancel the child, and child would throw an exception during the cancellation. The behavior is that we will re-throw the exception in already-cancelled tasks, so exception will bubble to parent. When parent catches the exception, it will run ...

https://github.com/redux-saga/redux-saga/blob/309d6be7507a22e2443e7843e61ef8555ea25eba/packages/core/src/internal/proc.js#L269-L270

... so the parent will be aborted with the bubbled exception.

```javascript
function* parent() {
  // ... other code ...
  yield cancel(child) // child would throw during cancellation
}
```

My concern is that whether it is the desired behavior of cancellation. Because cancelling a task means that we don't care that child anymore, exception from a cancelled task should NOT abort our parent task. However, if we want to have `env.onError` called with the exception, we must accept this behavior first. 😕

restrry:
logic that could throw in during cancellation controls the resource management, so we can leave a system in unpredictable state. basically there are 2 options:
- swallow an error and probably log it. using `onError` seems to be a bit awkward, as end user needs to distinguish between  error bubbling to the root and other 'not critical' errors.
- re-throw an error. breaking changes, but forces users to have safe `finally` block.

shinima:
> but forces users to have safe `finally` block.

I like this idea. By the way, if we adopt the re-throw approach, maybe we should warn users that an exception during cancellation occurs

restrry:
I'm on vacation for next 3 weeks, so feel free to make any changes required for v1 release

shinima:
Overall, I like this PR, it makes thing easier.  @Andarist what do you think?  If we like it, I will do a more thorough review.

packages/core/src/internal/middleware.js
```diff
@@ -5,18 +5,10 @@ import { identity } from './utils'
...
-  const { sagaMonitor, logger, onError, effectMiddlewares } = options
+  const { sagaMonitor, onError, effectMiddlewares } = options
   let boundRunSaga
 
   if (process.env.NODE_ENV !== 'production') {
-    if (is.notUndef(logger)) {
-      check(logger, is.func, 'options.logger passed to the Saga middleware is not a function!')
-    }
-
-    if (is.notUndef(onError)) {
```
Andarist: `onError` validation should be kept

packages/core/src/internal/runSaga.js
```diff
@@ -51,17 +56,8 @@ export function runSaga(options, saga, ...args) {
...
-    }
-  }
-
-  const log = logger || _log
-  const logError = err => {
-    log('error', err)
-    if (err && err.sagaStack) {
-      log('error', err.sagaStack)
-    }
+    check(onError, is.func, 'onError passed to the redux-saga is not a function!')
```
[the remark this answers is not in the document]
Andarist: sure it is weird, those checks are dev-only though and meant to help people not using TS/flow

packages/core/src/internal/middleware.js
Andarist: Oh, I see you have moved it entirely to runSaga file 👌

PR #1659, opened 2018-10-30
Move all effect runners to a separated file

shinima:
This PR is a subset of #1524, hope this could be merged more easily.

In this PR, effect-runners have a unified function interface which can be seen in the new implementation of `runEffect` (the interface may need further discussion and refinement):

```javascript
  if (is.promise(effect)) {
    resolveIterator(effect, effectId, meta, currCb)
    const executingContext = {
      effectId,
      task,
      taskContext,
      mainTask,
      taskQueue,
      digestEffect,
      resolvePromise,
      resolveIterator,
    }
    const effectRunner = effectRunnerMap[effect.type]
    effectRunner(env, effect.payload, currCb, executingContext)
  } else {
    // anything else returned as is
    currCb(effect)
  }
}
```

Unlike #1524, in this PR, I just move all the effect runners into a separated file. This PR has NO new features or breaking changes, it is just an internal refactor.

### DIFF
--- a/packages/core/src/internal/effectRunnerMap.js
+++ b/packages/core/src/internal/effectRunnerMap.js
@@ -124,10 +124,18 @@ export function runCPSEffect(env, { context, fn, args }, cb) {
 
   // catch synchronous failures; see #152
   try {
-    const cpsCb = (err, res) => (is.undef(err) ? cb(res) : cb(err, true))
+    const cpsCb = (err, res) => {
+      if (is.undef(err)) {
+        cb(res)
+      } else {
+        cb(err, true)
+      }
+    }
+
     fn.apply(context, args.concat(cpsCb))
+
     if (cpsCb.cancel) {
-      cb.cancel = () => cpsCb.cancel()
+      cb.cancel = cpsCb.cancel
     }
   } catch (error) {
     cb(error, true)
--- a/packages/core/src/internal/effectRunnerMap.js
+++ b/packages/core/src/internal/effectRunnerMap.js
@@ -2,6 +2,8 @@ import { SELF_CANCELLATION, TERMINATE } from '@redux-saga/symbols'
 import * as is from '@redux-saga/is'
 import * as effectTypes from './effectTypes'
 import { channel, isEnd } from './channel'
+// usage of proc here makes internal circular dependency
+// this works fine, but it is a little bit unfortunate
 import proc from './proc'
 import resolvePromise from './resolvePromise'
 import matcher from './matcher'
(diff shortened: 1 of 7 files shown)


shinima:
This PR seems to be an great start to do internal refactor,  I will investigate further potential refinement that can be done based on this PR.

packages/core/src/internal/proc.js
```diff
@@ -327,23 +280,18 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
...
-      else if (type === effectTypes.FORK) runForkEffect(payload, effectId, currCb)
-      else if (type === effectTypes.JOIN) runJoinEffect(payload, currCb)
-      else if (type === effectTypes.CANCEL) runCancelEffect(payload, currCb)
-      else if (type === effectTypes.SELECT) runSelectEffect(payload, currCb)
-      else if (type === effectTypes.ACTION_CHANNEL) runChannelEffect(payload, currCb)
-      else if (type === effectTypes.FLUSH) runFlushEffect(payload, currCb)
-      else if (type === effectTypes.CANCELLED) runCancelledEffect(payload, currCb)
-      else if (type === effectTypes.GET_CONTEXT) runGetContextEffect(payload, currCb)
-      else if (type === effectTypes.SET_CONTEXT) runSetContextEffect(payload, currCb)
-      else currCb(effect)
```
restrry: so we don't need this logic branch. do we?

packages/core/src/internal/effectRunnerMap.js
```diff
@@ -0,0 +1,323 @@
...
+    fn.apply(context, args.concat(cpsCb))
+    if (cpsCb.cancel) {
+      cb.cancel = () => cpsCb.cancel()
+    }
+  } catch (error) {
+    cb(error, true)
+  }
+}
+
+export function runForkEffect(env, { context, fn, args, detached }, cb, { effectId, taskContext, taskQueue }) {
```
restrry: the same question: is it okay to have a `cb` as not last argument?

packages/core/src/internal/proc.js
shinima: We do not need these if-else anymore, the branch logic is in this single statement: `const effectRunner = effectRunnerMap[effect.type]`.

packages/core/src/internal/effectRunnerMap.js
shinima: I think it is just fine. You can think of `cb` as required (every effect runner will use it), and think of the arguments after `cb` as optional (only a few runners will use them), and required argument comes before optional arguments.

shinima: We may refine the order if we have better alternatives 😄 .

shinima:
@Andarist Have a look at this PR?  It should be an PR that we would like to merge it.

Andarist:
I've made some changes - gonna probably merge this in in a moment but you can always comment it when you will have time and we can revisit/correct things.

PR #1660, opened 2018-10-30
Extract forkQueue and newTask to separated files

shinima:
Summary:

- internal refactor only, no public changes
- `forkQueue()` and `newTask()` get moved to new separated files
- The interface of `Task` and `effectRunner` gets changed

---

In this PR, I extract some of the code in `proc()` into two new files: _forkQueue.js_ and _newTask.js_. This PR is an internal refactor.

The biggest change in this PR is the interface of `Task` object:

- `task.mainTask` is added
- `task.context` is added, it is the same as the previous `taskContext`
- `task.queue` is added, it is the same as the previous `taskQueue`
- `task.cancel()` is added, it is the same as the previous `cancel()` which was defined in the `proc()`
- `task.end(result, isErr)` is added, it is the same as the previous `end()` which was defined in the `proc()`
- `task.crashedEffect` is added, which was defined in `proc()`
- `task._isRunning / task._isCancelled / task._isAborted / task._result / task._error / task._deferredEnd` gets removed, they are now defined as local variable in `newTask()`
- `cancelledDueToErrorTasks` is moved into `newTask()`

Because we can access the task-context and task-queue via the task object, the interface of effect runners can be simplified. It is now as follows:

```javascript
const executingContext = { task, digestEffect };

const effectRunner = effectRunnerMap[effect.type];
effectRunner(env, effect.payload, currCb, executingContext);
```

Seems that `effectRunner(env, task, effect.payload, currCb, digestEffect)` is a more proper interface. How do you think?

The idea behind this PR is that task-context or task-queue actually belongs to the task, so it is very natural that `.context` and `.queue` are fields of a task object. At the same time, `cancel()` and `end()` are operations to a task, so it is reasonable to write code such as `task.cancel()` or `task.end(arg, isErr)`.

### DIFF
--- a/packages/core/src/internal/forkQueue.js
+++ b/packages/core/src/internal/forkQueue.js
@@ -1,0 +1,75 @@
+import { noop, remove } from './utils'
+
+/**
+ Used to track a parent task and its forks
+ In the fork model, forked tasks are attached by default to their parent
+ We model this using the concept of Parent task && main Task
+ main task is the main flow of the current Generator, the parent tasks is the
+ aggregation of the main tasks + all its forked tasks.
+ Thus the whole model represents an execution tree with multiple branches (vs the
+ linear execution tree in sequential (non parallel) programming)
+
+ A parent tasks has the following semantics
+ - It completes if all its forks either complete or all cancelled
+ - If it's cancelled, all forks are cancelled as well
+ - It aborts if any uncaught error bubbles up from forks
+ - If it completes, the return value is the one returned by the main task
+ **/
+export default function forkQueue(mainTask, onAbort, cont) {
+  let tasks = []
+  let result
+  let completed = false
+
+  addTask(mainTask)
+  const getTasks = () => tasks
+  const getTaskNames = () => tasks.map(t => t.meta.name)
+
+  function abort(err) {
+    onAbort()
+    cancelAll()
+    cont(err, true)
+  }
+
+  function addTask(task) {
+    tasks.push(task)
+    task.cont = (res, isErr) => {
+      if (completed) {
+        return
+      }
+
+      remove(tasks, task)
+      task.cont = noop
+      if (isErr) {
+        abort(res)
+      } else {
+        if (task === mainTask) {
+          result = res
+        }
+        if (!tasks.length) {
+          completed = true
+          cont(result)
+        }
+      }
+    }
+  }
+
+  function cancelAll() {
+    if (completed) {
+      return
+    }
+    completed = true
+    tasks.forEach(t => {
+      t.cont = noop
+      t.cancel()
+    })
+    tasks = []
+  }
+
+  return {
+    addTask,
+    cancelAll,
+    abort,
+    getTasks,
+    getTaskNames,
+  }
+}
--- a/packages/core/src/internal/newTask.js
+++ b/packages/core/src/internal/newTask.js
@@ -1,0 +1,128 @@
+import deferred from '@redux-saga/deferred'
+import * as is from '@redux-saga/is'
+import { TASK, TASK_CANCEL } from '@redux-saga/symbols'
+import { assignWithSymbols, check, createSetContextWarning } from './utils'
+import { addSagaStack, sagaStackToString } from './error-utils'
+import forkQueue from './forkQueue'
+
+export default function newTask(env, mainTask, parentContext, parentEffectId, meta, isRoot, cont) {
+  let running = true
+  let cancelled = false
+  let aborted = false
+  let taskResult
+  let taskError
+  let deferredEnd = null
+
+  const cancelledDueToErrorTasks = []
+
+  const context = Object.create(parentContext)
+  const queue = forkQueue(
+    mainTask,
+    function onAbort() {
+      cancelledDueToErrorTasks.push(...queue.getTaskNames())
+    },
+    end,
+  )
+
+  /**
+   This may be called by a parent generator to trigger/propagate cancellation
+   cancel all pending tasks (including the main task), then end the current task.
+
+   Cancellation propagates down to the whole execution tree holded by this Parent task
+   It's also propagated to all joiners of this task and their execution tree/joiners
+
+   Cancellation is noop for terminated/Cancelled tasks tasks
+   **/
+  function cancel() {
+    // We need to check both Running and Cancelled status
+    // Tasks can be Cancelled but still Running
+    if (running && !cancelled) {
+      cancelled = true
+      queue.cancelAll()
+      // Ending with a TASK_CANCEL will propagate the Cancellation to all joiners
+      end(TASK_CANCEL)
+    }
+  }
+
+  function end(result, isErr) {
+    running = false
+
+    if (!isErr) {
+      taskResult = result
+      deferredEnd && deferredEnd.resolve(result)
+    } else {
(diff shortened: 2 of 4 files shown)


shinima:
As `_isRunning` / `_isCancelled` / `_isAborted` are defined under the scope of `newTask`, there is no need to have a underline as the name start, so I've change/simplify some field names of task/mainTask object in 7018761.

packages/core/src/internal/newTask.js
```diff
@@ -0,0 +1,135 @@
...
+    meta,
+    isRoot,
+    context,
+    joiners: [],
+    queue,
+    crashedEffect: null,
+
+    // methods
+    cancel,
+    cont,
```
restrry: should we rename it to something like `callback` or `continue`? `cont` sounds not super obvious.
we pass `cb` inside https://github.com/redux-saga/redux-saga/pull/1660/commits/5cf3e27fe770c9a36576e4d4bc3c88cfce1adc72#diff-ecefd26543e2e37544ae6d726d23b588R210

packages/core/src/internal/forkQueue.js
```diff
@@ -0,0 +1,74 @@
+import { noop, remove } from './utils'
+
+/**
+ Used to track a parent task and its forks
+ In the new fork model, forked tasks are attached by default to their parent
```
restrry: seems `new` can be deleted :)

packages/core/src/internal/newTask.js
shinima: I'm okay with both names. It seems that our convention is `cont` for task/mainTask, and `cb` for the rest  (runXXXEffect, digestEffect, etc.).

Andarist:
Is this PR/branch supposed to target effectRunnerMap branch and not the master?

shinima:
Updated the commits and the description of this PR. @Andarist @restrry

packages/core/src/internal/newTask.js
```diff
@@ -0,0 +1,134 @@
+import deferred from '@redux-saga/deferred'
+import * as is from '@redux-saga/is'
+import { TASK, TASK_CANCEL } from '@redux-saga/symbols'
+import { assignWithSymbols, check, createSetContextWarning } from './utils'
+import { addSagaStack, sagaStackToString } from './error-utils'
+import forkQueue from './forkQueue'
+
+export default function newTask(env, mainTask, parentContext, parentEffectId, meta, isRoot, cont) {
```
shinima: The interface of `newTask()` is very much like `proc()`.

`proc(env, iterator, parentContext, parentEffectId, meta, isRoot, cont)`

packages/core/src/internal/proc.js
```diff
@@ -289,7 +134,7 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
       resolvePromise(effect, currCb)
     } else if (is.iterator(effect)) {
       // resolve iterator
-      proc(env, effect, taskContext, effectId, meta, /* isRoot */ false, currCb)
+      proc(env, effect, task.context, effectId, meta, /* isRoot */ false, currCb)
     } else if (effect && effect[IO]) {
       const effectRunner = effectRunnerMap[effect.type]
       effectRunner(env, effect.payload, currCb, executingContext)
```
shinima: what about `effectRunner(env, task, effect.payload, currCb, digestEffect)`? 🤔 

The first argument is constant during the whole saga execution.
The second argument is constant during the executing of the task.
The third argument changes every time we yield a new effect.
The fourth is the callback, and the last is an argument we rarely use.

packages/core/src/internal/newTask.js
Andarist: Just posting both here for convenience:
```js
function newTask(env, mainTask, parentContext, parentEffectId, meta, isRoot, cont)
function proc(env, iterator, parentContext, parentEffectId, meta, isRoot, cont)
```

They indeed seem to relate to each other really closely, wondering if we can do smth about it.

packages/core/src/internal/newTask.js
```diff
@@ -0,0 +1,134 @@
...
+  let running = true
+  let cancelled = false
+  let aborted = false
```
Andarist: seems like it would be better to make those a single enum, i believe only one should ever be true, smth like
```js
const RUNNING = 0
const CANCELLED = 1
const ABORTED = 2
const DONE = 3

let status = RUNNING
```
should make the trick

packages/core/src/internal/newTask.js
```diff
@@ -0,0 +1,134 @@
...
+    queue,
+    crashedEffect: null,
+
+    // methods
+    cancel,
+    cont,
+    end,
+    setContext,
+    toPromise,
+    isRunning: () => running,
```
Andarist: this is slightly weird for me that task uses getter methods but mainTask uses an instance properties (so checking those things in the code is different, depending on what kind of task we check)

Andarist:
Could you also rebase this?

packages/core/src/internal/newTask.js
shinima: The problem is that "Tasks can be Cancelled but still Running", so it is impossible that we just use a variable 😞 

FYI, line 155, 156:
https://github.com/redux-saga/redux-saga/blob/a619f15cf46ea463528af86c18fb346e9f275985/packages/core/src/internal/proc.js#L153-L166

Andarist: Not sure when this can happen, maybe this comment is outdated/inaccurate?

packages/core/src/internal/newTask.js
shinima: Yes... a little weird.  I refactored `mainTask` in 586c375c70a65e483861c35198c3249124737c6b, how do you think about it?

shinima:
Rebased.

packages/core/src/internal/newTask.js
shinima: A task is *running and cancelled* when the task is performing the cleanup logic (code in the finally block).

Andarist: We dont support async cleanups though, so the finally block should return synchronously, therefore it should in theory make it safe to transition from running to cancelled either before making calling into cancellation logic or just after that (preferably former). This could require some minor code changes, but sounds like a reasonable assumption to me that it should be doable, right?

shinima: Even though there's no async cleanups, a cancelled task can still yield put/select /cancelled effects in their finally block, I'm not sure whether it is okay we just can transit the status from running to cancelled, especially when the put effects in cleanup code maybe delayed due to saga scheduling. 

I'm not saying it is impossible, I just mean that I'm not sure about it. Using a single status variable is attractive, but I think I need more thought on it .

shinima:
I'll try **using a single variable `status` to representing the status of a task** and **adding `task.digestEffect()`** in other PRs.

Is there anything actionable in this PR? or can I merge this PR now?

Andarist:
I still dont like that mainTask is using properties and task is using getter methods - but we can postpone changing this until we explore merging statuses into a single variable.

Also - I would explore trying to hide mainTask as implementation detail of the task, or the forkQueue. It's slightly confusing thing at maybe it could be hidden? Im not sure about it - just a thought I had right now.

PR #1667, opened 2018-11-06
Fixed real world example

garethbradley:
There were a couple of issues with the real-world example.

Use of lodash/array and lodash/object has changed and results in module not found error

Change to Github API? Keys are id numbers, not login names. Updated UserPage and RepoPage to find  IDs


examples/real-world/yarn.lock
```diff
@@ -0,0 +1,3805 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
Andarist: this file should be removed, while you are at it you could add `yarn.lock` to `.gitignore`d files

examples/real-world/containers/RepoPage.js
```diff
@@ -6,6 +6,7 @@ import Repo from '../components/Repo'
 import User from '../components/User'
 import List from '../components/List'
 import PropTypes from 'prop-types'
+import object from 'lodash/object'
```
Andarist: could we just switch to using `lodash` entry directly in those places? like `import { findKey } from 'lodash'`

garethbradley: I'm no expert - I'd have to try it. I'm guessing the syntax would look like:

`import { findKey } from 'lodash/object';`

Andarist: actually `lodash` exports everything from the "main entry" too, u dont have to "scope" it under `object/lodash`

if you dont feel confident in doing this, let me know and I'll merge this as is

garethbradley: There we go - that should be sorted.

Cheers

Andarist:
Thanks!

PR #1678, opened 2018-11-21
Using a single variable to track the status of Task

shinima:
Using a single variable to track task status per https://github.com/redux-saga/redux-saga/pull/1660#discussion_r234586922.  Feel free to add commits.

### DIFF
--- a/packages/core/__tests__/taskToPromise.js
+++ b/packages/core/__tests__/taskToPromise.js
@@ -1,0 +1,80 @@
+import { TASK_CANCEL } from '@redux-saga/symbols'
+import { runSaga, stdChannel } from '../src'
+import { cancel, delay, fork } from '../src/effects'
+import { noop } from '../src/internal/utils'
+
+function simpleRunSaga(saga) {
+  const channel = stdChannel()
+  return runSaga({ channel, dispatch: channel.put, onError: noop }, saga)
+}
+
+test('calling toPromise() of an already completed task', () => {
+  const result = 'result-of-saga'
+
+  const task = simpleRunSaga(function* saga() {
+    return result
+  })
+  expect(task.isRunning()).toBe(false)
+  return expect(task.toPromise()).resolves.toBe(result)
+})
+
+test('calling toPromise() before a task completes', () => {
+  const result = 'result-of-saga'
+
+  const task = simpleRunSaga(function* saga() {
+    yield delay(10)
+    return result
+  })
+  expect(task.isRunning()).toBe(true)
+  return expect(task.toPromise()).resolves.toBe(result)
+})
+
+test('calling toPromise() of an already aborted task', () => {
+  const error = new Error('test-error')
+
+  const task = simpleRunSaga(function* saga() {
+    throw error
+  })
+  expect(task.isRunning()).toBe(false)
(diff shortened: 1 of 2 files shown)


packages/core/src/internal/newTask.js
```diff
@@ -28,29 +31,33 @@ export default function newTask(env, mainTask, parentContext, parentEffectId, me
...
       // Ending with a TASK_CANCEL will propagate the Cancellation to all joiners
       end(TASK_CANCEL)
     }
   }
 
   function end(result, isErr) {
-    running = false
-
     if (!isErr) {
+      // The status here maybe RUNNING or CANCELLED
```
Andarist: "may be"

packages/core/src/internal/newTask.js
```diff
@@ -88,12 +94,10 @@ export default function newTask(env, mainTask, parentContext, parentEffectId, me
...
 
-    if (!running) {
-      if (aborted) {
-        deferredEnd.reject(taskError)
-      } else {
-        deferredEnd.resolve(taskResult)
-      }
+    if (status === ABORTED) {
+      deferredEnd.reject(taskError)
+    } else if (status === DONE) {
```
Andarist: CANCELLED should be resolved somehow too - in the Promise worls it seems like it should be treated as rejection (unfortunately) and that we should inject some artificial cancellation error here

shinima: To resolve with TASK_CANCEL is okay for me.  Because when we join a task inside a try-block, the code would not jump into the catch-block if the task is cancelled, while an error would let the iterator jump to catch-block.

Andarist: seems reasonable 👌

Andarist:
👌

packages/core/src/internal/newTask.js
restrry: ok for me

packages/core/__tests__/taskToPromise.js
```diff
@@ -0,0 +1,80 @@
...
+
+  simpleRunSaga(function* saga() {
+    child = yield fork(function* child() {
+      yield delay(10000)
+    })
+    yield cancel(child)
+  })
+
+  expect(child.isRunning()).toBe(false)
+  return expect(child.toPromise()).resolves.toBe(TASK_CANCEL)
```
restrry: should we cover `isCancelled/isAborted` in those tests as well?

packages/core/src/internal/newTask.js
```diff
@@ -5,10 +5,13 @@ import { assignWithSymbols, check, createSetContextWarning } from './utils'
 import { addSagaStack, sagaStackToString } from './error-utils'
 import forkQueue from './forkQueue'
 
+const RUNNING = 0
+const CANCELLED = 1
+const ABORTED = 2
+const DONE = 3
```
restrry: should mainTask have `status` filed as well for consistency?

packages/core/__tests__/taskToPromise.js
shinima: Seems testing `isCancelled / isAborted`  is a little trivial, but ofc you could rename this file to `__tests__/task.js` and add some tests for other methods on task object.

PR #1679, opened 2018-11-21
Make digestEffect accessible via task.digestEffect

shinima:
Make `digestEffect` accessible via `task.digestEffect` per https://github.com/redux-saga/redux-saga/pull/1660#discussion_r234583861, and change the interface of effectRunner to  `effectRunner(env, task, effect.payload, currCb)`

I simply pass the `digestEffect` when calling `newTask()`, there maybe better alternatives to implement the `task.digestEffect`. Just post it here so we can discuss it, feel free to change the implementation.

### DIFF
--- a/packages/core/src/internal/newTask.js
+++ b/packages/core/src/internal/newTask.js
@@ -5,7 +5,7 @@ import { assignWithSymbols, check, createSetContextWarning } from './utils'
 import { addSagaStack, sagaStackToString } from './error-utils'
 import forkQueue from './forkQueue'
 
-export default function newTask(env, mainTask, parentContext, parentEffectId, meta, isRoot, cont) {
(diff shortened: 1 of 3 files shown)


packages/core/src/internal/newTask.js
```diff
@@ -114,6 +114,7 @@ export default function newTask(env, mainTask, parentContext, parentEffectId, me
     // methods
     cancel,
     cont,
+    digestEffect,
```
Andarist: this is somewhat leaky (but the problem is not exclusive to the digestEffect) - because we give those tasks to users and things like - `digestEffect` should never be called by them (except in effect runners)

We don't necessarily have to fix it, but its a pity nevertheless. Not sure how we should split those in public and kinda protected interfaces

shinima: Oh, I forgot that the task can be accessed by the users, adding too much fields/methods to the task is a bad idea.

PR #1687, opened 2018-11-25
Use jest.useFakeTimer instead of lolex.

shinima:

Just remove lolex and use the built-in `jest.useFakeTimer`.


packages/core/__tests__/sagaHelpers/throttle.js
```diff
@@ -58,17 +59,11 @@ test('throttle', () => {
...
-      },
-      err => {
-        clock.uninstall()
-        throw err
-      },
-    )
 })
+
 test('throttle: pattern END', () => {
+  jest.useRealTimers()
```
restrry: cleanup should be done in the same test when you setup a  fake timer (as it's done before)

Issue #1684, opened 2018-11-24
sagaStack refactor


Andarist:
instead of using a `.sagaStack` property we should just wrap errors in a tuple or smth (i.e. `[caught, sagaStack]`), propagate it through the code and unwrap when needed - that would allow us to collect `sagaStack` for thrown non-objects

onError should be called like this - `onError(caught, { sagaStack })`

Andarist:
Done in #1695

PR #1695, opened 2018-11-29
Sagastack refactoring

restrry:
| Fixed Issues?            | https://github.com/redux-saga/redux-saga/issues/1684 |
| Major: Breaking Change?  |  Yes  |

Pass ErrorStack object, which is a wrapper around `Error`, which collects all related runtime info for the exception & unwrap error before passing into user land

### DIFF
--- a/packages/core/src/internal/proc.js
+++ b/packages/core/src/internal/proc.js
@@ -5,8 +5,8 @@ import effectRunnerMap from './effectRunnerMap'
 import resolvePromise from './resolvePromise'
 import nextEffectId from './uid'
 import { asyncIteratorSymbol, noop, shouldCancel, shouldTerminate } from './utils'
-import { SagaError } from './error-utils'
 import newTask from './newTask'
+import * as sagaError from './sagaError'
 
 export default function proc(env, iterator, parentContext, parentEffectId, meta, isRoot, cont) {
   if (process.env.NODE_ENV !== 'production' && iterator[asyncIteratorSymbol]) {
@@ -62,16 +62,16 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
    * This is the generator driver
    * It's a recursive async/continuation function which calls itself
    * until the generator terminates or throws
-   * @param {internal commands(TASK_CANCEL | TERMINATE) | SagaError | any} arg - value, generator will be resumed with.
+   * @param {internal commands(TASK_CANCEL | TERMINATE) | any} arg - value, generator will be resumed with.
    * @param {boolean} isErr - the flag shows if effect finished with an error
    *
-   * receives either (command | effect result, false) or (SagaError, true)
+   * receives either (command | effect result, false) or (any thrown thing, true)
    */
   function next(arg, isErr) {
     try {
       let result
       if (isErr) {
-        result = iterator.throw(arg.getError())
+        result = iterator.throw(arg)
       } else if (shouldCancel(arg)) {
         /**
           getting TASK_CANCEL automatically cancels the main task
@@ -108,13 +108,13 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
         }
         mainTask.cont(result.value)
       }
-    } catch (err) {
+    } catch (error) {
       if (mainTask.status === CANCELLED) {
-        throw err
+        throw error
       }
       mainTask.status = ABORTED
 
-      mainTask.cont(isErr ? arg : new SagaError(err), true)
+      mainTask.cont(error, true)
     }
   }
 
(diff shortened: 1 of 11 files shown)


packages/core/src/internal/proc.js
```diff
@@ -163,15 +169,18 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
       cb.cancel = noop // defensive measure
       if (env.sagaMonitor) {
         if (isErr) {
-          env.sagaMonitor.effectRejected(effectId, res)
+          env.sagaMonitor.effectRejected(effectId, SagaError.isErrorAlreadyWrapped(res) ? res.getError() : res)
```
Andarist: when this can be not wrapped? shouldnt the wrapping happen right away after catching the error? so it should be already wrapped when it reaches this line?

packages/core/src/internal/proc.js
```diff
@@ -163,15 +169,18 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
...
         } else {
           env.sagaMonitor.effectResolved(effectId, res)
         }
       }
+
       if (isErr) {
-        task.crashedEffect = effect
```
Andarist: I really like removing this from the task 👌

packages/core/src/internal/proc.js
```diff
@@ -103,12 +108,13 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
...
-    } catch (error) {
+    } catch (err) {
       if (mainTask.status === CANCELLED) {
-        throw error
+        throw err
       }
       mainTask.status = ABORTED
-      mainTask.cont(error, true)
+
+      mainTask.cont(isErr ? arg : new SagaError(err), true)
```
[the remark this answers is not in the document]
Andarist: this changes how it behave previously - could u explain how it behave and how it will behave after this change?

previously locally caught error was passed to `mainTask.cont`, currently `arg` gets reused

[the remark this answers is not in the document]
Andarist: ok, after a thought i think i know how it works - previously arg and locally caught error were actually the same, right?

packages/core/src/internal/proc.js
TranChungThanh: Chungtran6655@gmail.com

Andarist: ok, so it seems this gets actually called before wrapping as `currCb` in which we are here is actually the first thing to be called by effect runners 😅

Andarist:
I've added a refactoring commit - please take a look.

I'm trying to bookkeep the sagaStack separately from passing the real error around. This allows me to just pass the thrown thing without converting it back to the real thrown error before passing it to user's code.

This also has a nice benefit of saving over 100 bytes 😃 We might be able to sit under 6.5kb now.

Tests are failing now, because I don't reset the sagaStack - Im wondering where is the best place to do that - seems like we have to do it at 2 places:
- after user handles the error
- when we stringify the sagaStack (its passed to onError then and we know it wont be used anymore)

Gonna try with that ^ in a moment.

Andarist:
Tests are failing, but it seems that they should actually pass (expected results are wrong?). Those failing tests have no info about crashed effect in the sagaStack - not sure how exactly I have affected this, but it seems that current behaviour is more correct, right?

packages/core/src/internal/sagaError.js
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export const clear = () => {
+  crashedEffect = UNIQUE
+  sagaStack.length = 0
+}
+
+export const setCrashedEffect = effect => {
+  // track only first effect (the one which has caused the crash)
+  if (crashedEffect !== UNIQUE) {
```
Andarist: after a thought - falsy values cannot crash saga, so the whole `UNIQUE` thing can be removed, and a simple truthy check should suffice

packages/core/src/internal/proc.js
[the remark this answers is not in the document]
restrry: yes, I must add the comment about it. 
right now there are 2 source of errors for `next`
1 an exception passed as an argument. we will throw it into iterator and will catch it again. to remain stack trace we should reuse `arg`.
2 an exception raised while resuming iterator, so we wrap it

restrry:
@Andarist actually tests are correct. let's look at the next example:
```js
function* sagaA(){
  yield call(sagaB);
}

function* sagaB(){
  undefinedIsNotAFunction()
}
```
here error happens during `sagaB` execution, but current code says that error happens in `sagaB` during effect invocation `call(sagaB)`.
so we should log info about crashed effect *only* for first error stack
here we solved it with checking if error is already wrapped
https://github.com/redux-saga/redux-saga/pull/1695/commits/be2e862bc0df84ccf3a05f689a8bf0fca866f7a0#diff-ecefd26543e2e37544ae6d726d23b588R179

Andarist:
Makes sense. Before going further with this - WDYT about current implementation?

restrry:
@Andarist my only concern is singleton usage, as it's another source of state, which we have to control manually. but seems that approach simplifies code and does the trick, so go ahead 👍

Andarist:
I agree that in general this is a weird pattern - but sometimes it comes really handy. Fun fact is that React uses such things heavily 😉 

I've fixed tests by stretching the singleton usage even further and I'm not that happy about it, but couldn't figure out quickly how to fix it in a better way. I've added an explanatory comment about it to help the future readers, if anybody comes to better solution - I would be more than happy to merge it in. It's an implementation detail though so I'm going to merge in this PR in a moment.

This means that we are ready to release release candidate version this week ❤️ Gonna wait for improved TS typings before releasing stable v1 but there is generally nothing else to be done here - implementation-wise. The only thing we need to do before releasing v1 is figuring out how to deal with README file

PR #1700, opened 2018-12-05
Update docs for v1

restrry:
| Fixed Issues?            | update docs |

1. could you check that I didn't miss something?
1. as I understand all task internal changes didn't affect public API. did it?
we removed `__is*` flags from `task` interface, but it doesn't mean to be a part of public interface.
1. also we added `mainTask` to `task` interface. seems it's a detail of internal implementation should we expose it at all?

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -118,7 +118,7 @@ export interface SagaMiddlewareOptions<C extends object = {}> {
    * If provided, the middleware will call it with uncaught errors from Sagas.
    * useful for sending uncaught exceptions to error tracking services.
    */
-  onError?(error: Error): void
+  onError?(error: Error, errorInfo: ErrorInfo): void
   /**
    * Allows you to intercept any effect, resolve it on your own and pass to the
    * next middleware.
@@ -188,6 +188,10 @@ export function runSaga<Action, State, S extends Saga>(
   ...args: Parameters<S>
 ): Task
 
+interface ErrorInfo {
+  sagaStack: string
+}
+
 /**
  * The `{subscribe, dispatch}` is used to fulfill `take` and `put` Effects. This
  * defines the Input/Output interface of the Saga.
@@ -224,7 +228,7 @@ export interface RunSagaOptions<A, S> {
   /**
    * See docs for `createSagaMiddleware(options)`
    */
-  onError?(error: Error): void
+  onError?(error: Error, errorInfo: ErrorInfo): void
   /**
    * See docs for `createSagaMiddleware(options)`
    */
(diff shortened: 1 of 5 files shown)


shinima:
We still need to use `task.mainTask.status` in `runCancelledEffect()`.  

A task may have its iterator consumed up, but still wait for a long-running child task. In such cases, `mainTask.status` is `DONE` and `task.status` is `CANCELLED`, we need `mainTask.status` to determine whether a saga is cancelled or not.

```javascript
function* saga() {
  yield fork(longRunningSaga)
  // mainTask.status is DONE here
  // but if the task is cancelled at this point, the task.status will be CANCELLED
}
```

shinima:
3. The refactor to `task` and `mainTask` should be internal only. I did not change the public interface of Task. so we don't need to document the changes.

Andarist:
`mainTask` is special anyway so maybe we should just transition it to CANCELLED state when the task gets cancelled?

shinima:
`mainTask.status` should reflect the fact of running status of the iterator, so I think changing mainTask.status is not a good idea.

On the other hand, I think maybe using `cb(task.isCancelled())` in `runCancelledEffect()` is enough. `mainTask.status` becoming DONE means the main body of the saga has finished, so the saga will never yield a cancelled effect to query the cancellation status. That means we can simply ignore the cases when `mainTask.status` is `DONE`, and we do not need to care the edge case [that I just described](https://github.com/redux-saga/redux-saga/pull/1700#issuecomment-444439738).

However, I do a quick test against *using `cb(task.isCancelled())` in `runCancelledEffect()`* on my PC, two test cases are failing.

Andarist:
I'm investigating switching to `task.isCancelled()` in `runCancelledEffect`. The problem is in order of things when propagating the cancellation. I'll report back when I will have more concrete findings or a solution for this.

packages/core/index.d.ts
```diff
@@ -58,11 +58,14 @@ export interface SagaMiddleware<C = {}> extends Middleware {
 
   setContext(props: Partial<C>): void;
 }
+interface onErrorRuntime {
```
Andarist: I'd call it something like `ErrorInfo` or something, `onErrorRuntime` suggests to me that this is a handler function

restrry: ok, will fix after https://github.com/redux-saga/redux-saga/pull/1703 merged

Andarist:
Could u rebase?

restrry:
done

Andarist:
❤️ gonna drop RC this week! Thanks for all the work you have put into the project

PR #1703, opened 2018-12-10
Revamp typings

### LINKED ISSUE #1683: TS typings for v1
Need to do this https://github.com/redux-saga/redux-saga/pull/1655#issuecomment-440672034 , also @aikoven has some other things that he'd like to refactor before stable v1.

aikoven:
Closes #1683.

* Replaced N overloads with [Generic rest parameters](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#generic-rest-parameters) for stuff like `call`, `fork`, `select`, `takeEvery` etc.
* Added JSDoc to definitions so that users could see the docs in IDE quick help popups (even if they don't use TS but use VS Code!)
* Reformatted typings and tests according to project's Prettier rules.

Tests are mostly untouched (apart from formatting), so there's backward compatibility to some extent. Still, these changes are certainly breaking.


packages/types/index.d.ts
```diff
@@ -1,84 +1,197 @@
...
+  toPromise<T = any>(): Promise<T>
+  /**
+   * Cancels the task (If it is still running)
+   */
+  cancel(): void
+  setContext<C extends object>(props: Partial<C>): void
+}
+
+// https://github.com/Microsoft/TypeScript/issues/5453#issuecomment-419680547
+export namespace TupleHelpers {
```
restrry: is it used somewhere?

Andarist: Also - shouldnt namespaces be avoided?

packages/types/package.json
```diff
@@ -12,7 +12,8 @@
   "contributors": [
     "Mateusz Burzyński <mateuszburzynski@gmail.com> (https://github.com/Andarist)",
     "Mikhail Shustov <restrry@gmail.com> (https://github.com/restrry)",
-    "Shi Feichao <842351815@qq.com> (https://github.com/shinima)"
+    "Shi Feichao <842351815@qq.com> (https://github.com/shinima)",
+    "Daniel Lytkin <dan.lytkin@gmail.com (https://github.com/aikoven)"
```
Andarist: ❤️

packages/types/index.d.ts
Andarist: it's used in effect.d.ts, in general I'd prefer not using namespaces, you have also pulled in all helpers from the linked issue, but you are using only 3 of them, so probably we should remove unused ones (even if they might come handy in the future) - or you could maybe publish those types as a package and we could depend on it here? :)

aikoven: If you're ok with adding a dependency, there's a package that has similar helpers: https://github.com/ksxnodemodules/typescript-tuple

Andarist: I have no problems with adding dependencies - especially non-runtime ones 😄 Outsource all the things 😉

aikoven: Done!

packages/core/package.json
```diff
@@ -58,7 +58,8 @@
     "@redux-saga/is": "^1.0.0-beta.3",
     "@redux-saga/symbols": "^1.0.0-beta.3",
     "@redux-saga/types": "^1.0.0-beta.3",
-    "redux": ">=0.10 <5"
+    "redux": ">=0.10 <5",
+    "typescript-tuple": "^2.1.0"
```
restrry: shouldn't it be a devDependency?

Andarist:
Thanks for the great work @aikoven  ❤️

packages/core/index.d.ts
```diff
@@ -1,164 +1,358 @@
...
+   * block.
+   *
+   * In the case a Saga is cancelled (either manually or using the provided
+   * Effects), the middleware will invoke `return()` method of the Generator.
+   * This will cause the Generator to skip directly to the finally block.
+   *
+   * @param saga a Generator function
+   * @param args arguments to be provided to `saga`
+   */
+  run<S extends Saga>(saga: S, ...args: Parameters<S>): Task
```
stianjensen: Where does the name `Parameters` come from? I'm getting errors in my code saying `Cannot find name 'Parameters'.` when I try to compile with typescript.

stianjensen: Seems like the issue might have been that I was using typescript 3.0.0.

Will typescript 3.2.0 be a requirement for redux-saga 1.x?

Andarist: Is `Parameters` builtin type in TS@3.2?

stianjensen: I don't know, but the error messages went away when I upgraded \_(ツ)_/¯

Andarist: We definitely don't have manpower to maintain more than one set of typings. From what I see though this is implemented using constructs available before that, so maybe we could just use `Parameters` copy temporarily.

Not sure though what TS version is requirement for the rest of the typings. If we have more typings depending on >3.1 then we'll have to just require that version from our users. cc @aikoven

aikoven: Yeah, I think that we now require TS >= 3.1. We could make it 3.0 by copying `Parameters` if that makes enough sense for someone.

Andarist: Let's sit on this one until final release to see what's the initial feedback.

packages/core/index.d.ts
```diff
@@ -1,164 +1,358 @@
...
+  /**
+   * Allows you to intercept any effect, resolve it on your own and pass to the
+   * next middleware.
+   */
+  effectMiddlewares?: EffectMiddleware[]
 }
 
+export interface SagaMiddleware<C extends object = {}> extends Middleware {
+  /**
+   * Dynamically run `saga`. Can be used to run Sagas **only after** the
```
dorian-marchal: Can't it be rerun after stopping it by dispatching `END`?

PR #1704, opened 2018-12-11
Remove mainTask property from the Task object

Andarist:
related to https://github.com/redux-saga/redux-saga/pull/1700#issuecomment-446185315

### DIFF
--- a/packages/core/src/internal/newTask.js
+++ b/packages/core/src/internal/newTask.js
@@ -18,7 +18,7 @@ export default function newTask(env, mainTask, parentContext, parentEffectId, me
   const queue = forkQueue(
     mainTask,
     function onAbort() {
-      cancelledDueToErrorTasks.push(...queue.getTaskNames())
+      cancelledDueToErrorTasks.push(...queue.getTasks().map(t => t.meta.name))
     },
     end,
   )
@@ -39,7 +39,7 @@ export default function newTask(env, mainTask, parentContext, parentEffectId, me
       status = CANCELLED
       queue.cancelAll()
       // Ending with a TASK_CANCEL will propagate the Cancellation to all joiners
-      end(TASK_CANCEL)
+      end(TASK_CANCEL, false)
     }
   }
 
@@ -103,14 +103,12 @@ export default function newTask(env, mainTask, parentContext, parentEffectId, me
     // fields
     [TASK]: true,
     id: parentEffectId,
-    mainTask,
     meta,
     isRoot,
     context,
     joiners: [],
     queue,
     crashedEffect: null,
-
     // methods
     cancel,
     cont,
@@ -118,7 +116,7 @@ export default function newTask(env, mainTask, parentContext, parentEffectId, me
     setContext,
     toPromise,
     isRunning: () => status === RUNNING,
-    isCancelled: () => status === CANCELLED,
+    isCancelled: () => status === CANCELLED || (status === RUNNING && mainTask.status === CANCELLED),
     isAborted: () => status === ABORTED,
     result: () => taskResult,
     error: () => taskError,
--- a/packages/core/src/internal/effectRunnerMap.js
+++ b/packages/core/src/internal/effectRunnerMap.js
@@ -9,7 +9,6 @@ import resolvePromise from './resolvePromise'
 import matcher from './matcher'
 import { asap, immediately } from './scheduler'
 import { current as currentEffectId } from './uid'
-import { CANCELLED } from './task-status'
 import {
   assignWithSymbols,
   createAllStyleChildCallbacks,
@@ -168,6 +167,22 @@ function runForkEffect(env, { context, fn, args, detached }, cb, { task: parent 
 }
 
 function runJoinEffect(env, taskOrTasks, cb, { task }) {
+  const joinSingleTask = (taskToJoin, cb) => {
+    if (taskToJoin.isRunning()) {
+      const joiner = { task, cb }
+      cb.cancel = () => {
+        remove(taskToJoin.joiners, joiner)
+      }
+      taskToJoin.joiners.push(joiner)
+    } else {
+      if (taskToJoin.isAborted()) {
+        cb(taskToJoin.error(), true)
+      } else {
+        cb(taskToJoin.result())
+      }
+    }
+  }
+
   if (is.array(taskOrTasks)) {
     if (taskOrTasks.length === 0) {
       cb([])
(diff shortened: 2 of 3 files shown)


shinima:
Thanks for clarifying the logic in https://github.com/redux-saga/redux-saga/pull/1700#issuecomment-446185315.

> I believe we should switch task to CANCELLED state in the end when receiving TASK_CANCEL, otherwise it will become DONE which is really confusing - after all the task received TASK_CANCEL and its mainTask is cancelled.

As far as I known, before receiving TASK_CANCEL the task should be already in CANCELLED status, because we have set it in `cancel()`:

https://github.com/redux-saga/redux-saga/blob/1b3ea61b41d322f28d288a32bc0eb931f430cb5e/packages/core/src/internal/newTask.js#L35-L44

I do the following edit...

```diff
// newTask.js
 function end(result, isErr) {
    if (!isErr) {
      // The status here may be RUNNING or CANCELLED
      // If the status is CANCELLED, then we do not need to change it here
-     if (result === TASK_CANCEL) {
-       status = CANCELLED
-     } else if (status !== CANCELLED) {
+     if (status !== CANCELLED) {
        status = DONE
      }
      taskResult = result
      deferredEnd && deferredEnd.resolve(result)
    } else {
```

... and tests are still passing.

packages/core/src/internal/newTask.js
```diff
@@ -39,15 +39,17 @@ export default function newTask(env, mainTask, parentContext, parentEffectId, me
...
     }
   }
 
   function end(result, isErr) {
     if (!isErr) {
       // The status here may be RUNNING or CANCELLED
       // If the status is CANCELLED, then we do not need to change it here
-      if (status !== CANCELLED) {
+      if (result === TASK_CANCEL) {
+        status = CANCELLED
```
shinima: Seems that this statement is unnecessary, see https://github.com/redux-saga/redux-saga/pull/1704#issuecomment-446208451

shinima:
If I am wrong, please point me out. It seems incorrect to me that we do ` task.isCancelled = () => status === CANCELLED || mainTask.status === CANCELLED`. The mainTask/iterator gets cancelled does not mean that the task is cancelled:

image

In the figure above, circles mean iterators and squares mean tasks. It shows an error bubbling from a forked descendants to the root. When the error occurs (left-side in the figure), most of the iterators(mainTasks) are in running status (colored as blue). When the bubbling is done (right-side in the figure), all running iterators go into cancelled status (colored as gray) including the root iterator.

So in this figure, root.mainTask.status is CANCELLED, but root.task.status is ABORTED other than CANCELLED.

shinima:
You can run the following test which will lead to a strange behavior.

```javascript
test('task status', () => {
  const middleware = sagaMiddleware()
  createStore(() => ({}), {}, applyMiddleware(middleware))

  function* genFn() {
    yield io.fork(function*() {
      yield io.delay(10)
      throw new Error('error')
    })
    yield io.delay(10000)
  }

  const task = middleware.run(genFn)
  return task.toPromise().catch(err => {
    console.log(task.isCancelled()) // prints true
    console.log(task.isAborted()) // prints true
    // the task is both cancelled and aborted
  })
})
```

shinima:
🤔  The above figure make me believe that the implementation in the master branch (i.e `cb(task.mainTask.status === CANCELLED)` for `runCancelledEffect`) is correct, because the cancellation status of mainTask and task is not related --  mainTask.status could be CANCELLED while task.status is ABORTED (as the above figure shows), and task.status could be CANCELLED while mainTask.status is DONE (as described in https://github.com/redux-saga/redux-saga/pull/1700#issuecomment-444439738).

packages/core/src/internal/newTask.js
Andarist: Yeah, tests are passing because I havent add a new one (yet). It's true that `cancel` sets the status to `CANCELLED`, but `end` (as task.cont or mainTask.cont, cant remember right now) might be called from the proc too (without calling `task.cancel` here) when propagating the `TASK_CANCEL` through saga tree

Andarist:
Thanks for the thorough comments! ❤️ I'll try to analyze this some time later to give proper feedback back.

packages/core/src/internal/newTask.js
shinima: Yeah, I have thought there may be some similar cases that `task.end(TASK_CANCEL)` is called without calling `task.cancel()`,  but I failed to write a test case to prove the idea. It will be best you can add tests. 😄

Andarist:
Ok, so it seems that a simple tweak to:
```js
task.isCancelled = () => status === CANCELLED || (status === RUNNING && mainTask.status === CANCELLED)
```
should do the trick, right?

packages/core/__tests__/interpreter/cancellation.js
```diff
@@ -760,3 +760,31 @@ test('should bubble an exception thrown during cancellation', () => {
...
+  let task
+  function* worker() {
+    task = yield io.fork(parent)
+  }
+
+  return middleware
+    .run(worker)
+    .toPromise()
+    .then(() => {
+      expect(task.isCancelled()).toBe(true)
```
shinima: This is expected to be false. Cancellation won't bubble from child to parent.  Uncaught errors bubbles from child to parent, and cancellation propagates from parent to children.

shinima:
> Ok, so it seems that a simple tweak to: `task.isCancelled = () => status === CANCELLED || (status === RUNNING && mainTask.status === CANCELLED)` should do the trick, right?

I think it is correct now, but it is so complicated 😓 .   

How about add a method (e.g. `isMainCancelled()`) on Task object, so we can call `task.isMainCancelled()` in `runCancelledEffect()` and then we can remove mainTask property from the Task object?  It should be much more straightforward than the hacking `task.isCancelled()`.

packages/core/__tests__/interpreter/cancellation.js
Andarist: Right, this was a quick draft and I have not finished it (and obviously made a mistake here).

Andarist:
> How about add a method (e.g. isMainCancelled()) on Task object? Then we can call task.isMainCancelled() in runCancelledEffect(), and then we can remove mainTask property from the Task object. It should be much more straightforward than hacking task.isCancelled().

We could do that, it's better than exposing whole `mainTask`. I'm only worried that whole notion of mainTask is an implementation detail and it might be better to hide it entirely. WDYT? cc @restrry , let's reach consensus on that, I'm not overly attached to either option, so let's speak up with your preferences.

restrry:
imho we should isolate as much internals as possible, so I'd go with this sophisticated condition in `cancel` with comment `why` and `what` we check here.

PR #1706, opened 2018-12-14
Fix issue when joining cancelled state didn't cause joiner to land in…

Andarist:


packages/core/__tests__/interpreter/cancellation.js
```diff
@@ -760,3 +760,55 @@ test('should bubble an exception thrown during cancellation', () => {
 
   return expect(middleware.run(worker).toPromise()).rejects.toBe(expectedError)
 })
+
+test('task should end in cancelled state when joining cancelled child', () => {
```
Andarist: this is a new test which didnt work previously

packages/core/__tests__/interpreter/cancellation.js
```diff
@@ -760,3 +760,55 @@ test('should bubble an exception thrown during cancellation', () => {
...
+  const task = middleware.run(parent)
+
+  return task.toPromise().then(() => {
+    expect(task.isCancelled()).toBe(true)
+    expect(task.isRunning()).toBe(false)
+    expect(task.isAborted()).toBe(false)
+  })
+})
+
+test('task should end in cancelled state when parent gets cancelled', () => {
```
Andarist: this has worked but I have added in nevertheless when testing things out

PR #1712, opened 2018-12-21
Move core source to scoped package, use "redux-saga" only as proxy to it

Andarist:
I got to really like lately scoped packages - how they are nicely grouped in `node_modules`. This is a proposal to use `@redux-saga/core` package to hold all the core's source and use old `redux-saga` merely to proxy to it. WDYT?


shinima:
👍   +1 for this.   

If we do this, should we change how effects are imported? I think it would be too cumbersome to use `import { put, take } from '@redux-saga/core/effects'`. `'@redux-saga/effects'` seems more suitable.

Andarist:
The public, recommended way, of importing both redux-saga and effects would still be:
```js
import createSagaMiddleware from 'redux-saga'
```

PR #1717, opened 2018-12-29
Add Rollup configuration to generate Browser-compatible ESM bundle

### LINKED ISSUE #1716: Add ES-browser build
## Description of the bug/issue
Similarly to redux, redux-saga is using the `NODE_ENV` variable for some checks. This variable resides in window.process, which is not defined in some cases. Adding a minified build with the production checks removed will alleviate this issue.

## Example fix
The PR linked below fixes this issue for redux, but now the problem is still present in redux-saga:
https://github.com/reduxjs/redux/pull/3143

Simply adding another js build for redux-saga that removes the production checks and minifies the code will be sufficient.

TimvdLippe:
| Fixed Issues?            | Fixes #1716 |
| Patch: Bug Fix?          | Feature   |
| Any Dependency Changes?  | Yes  |

This should generate browser-compatible ESM bundle. I haven't had the time to test it in my project, but I have inspected the output bundle and it looks correct. This is the general configuration based on https://github.com/reduxjs/redux/pull/3143

I also had to replace uglify with terser, as uglify is out-of-date and does not parse modern ES2015+ statements. Lastly Rollup had to be updated to fix an issue with transform streams breaking.

@martijnjanssen could you verify this works for you? You can `npm pack` the `packages/core` to generate the files.


Andarist:
Im still wondering how do you plan to consume it - do you use any bundler? redux-saga has no peer dependencies but if it would - how would u consume them (they would be referenced as npm modules in the source)?

TimvdLippe:
We dont use any bundler. We npm install redux-saga and then use es modules
in the browser to import the relevant parts. We serve using python
http.server.

Since redux and redux-saga should only be installed and loaded once, they
are peers in node_modules.

Andarist:
But how do u access that es bundle? do you copy the file from node_modules? or do u just serve node_modules content when receiving requests?

Do you use react?

Don't take me wrong - I'm not opposed to merging it (in some form). I just try to understand the problem at hand better.

TimvdLippe:
No problem at all!

At build time, we copy and process only the relevant files to our build folder. This includes node_modules files. This means that all unused (configuration) files are removed when building a production bundle.

During dev, we have a simple server which just returns files. No processing or transformations there.

We do not use React. Just good old JavaScript with some libraries (like yours).

martijnjanssen:
We are using redux-saga just like you are, but not combined with react. We are using https://github.com/Polymer/lit-element for the webcomponents implementation, this is easily combined with redux and saga, similarly to react. This can all be served by a simple webserver.

For the build it is possible to simply find all imports in the components, process (minify, transpile, whatever you want) them and output all the files in some `build` directory. When serving this `build` directory (Nginx or another fileserver) all correct import paths are preserved. Redux-saga will then reside in, for example, `/node_modules/redux-saga/redux-saga.js` or similar.

Implementation in another framework or native javascript is actually the same as using it in react. I think that react making all these abstractions in the build process is not that beneficial, since people are so dependent on the current build system (webpack). There are a lot of ways to achieve the same thing and allow for more tweaking or freedom in implementation.

Andarist:
My question was more about how do u technically use it (how browser knows what file it should request etc) rather how combine saga with other framework. Its really cool that u have found use for it outside of react!

Im nly concerned about understanding ur build setup, seems like u have to manually maintain mapping between node modules names to disk files? And parse imported files to rewrite imports to those (during build)? 

Dont u have currently a waterfall of requests on the startup of ur app?

martijnjanssen:
Yes, there is a mapping directly to the files right now, using the path. This indeed causes a waterfall, but this can be changed using a bundle action (or something similar) in the build step later on. We just don't want a dependency on webpack.

Andarist:
> We just don't want a dependency on webpack.

There are other bundlers out there though, seems like you are going to have to reimplement some existing stuff - I'm not the one to judge though.

package.json
```diff
@@ -86,7 +86,7 @@
     "redux-devtools-log-monitor": "^1.0.0-beta-3",
     "redux-logger": "^2.0.2",
     "rimraf": "^2.4.3",
-    "rollup": "^0.65.2",
+    "rollup": "^0.66.0",
```
Andarist: If we are updating this - we should update to v1.

packages/core/rollup.config.js
```diff
@@ -50,6 +50,8 @@ const createConfig = ({ input, output, external, env, min = false, useESModules
     }),
     babel({
       exclude: 'node_modules/**',
+      // Es modules in browser does not need transpilation
```
Andarist: This is not quite true, I'd prefer to transpile this - just not to the es5. We could leverage https://babeljs.io/docs/en/babel-preset-env#targetsesmodules .

This would make the setup slightly more complicated (how to turn it on & off in babelrc files?), but maybe we could leverage upcoming `.custom` feature of `rollup-plugin-babel` (available in v5.0.0-alpha.0), I don't have time right now to investigate this, but it would look smth like this:
```
babel.custom(() => ({
  config(config) {
    // manipulate the config here (it should contain `@babel/preset-env`)
    // or just return untouched config back for every other build
    return config
  }
}))
```

packages/core/rollup.config.js
```diff
@@ -92,6 +94,17 @@ export default [
       entryFileNames: 'redux-saga-[name].[format].js',
     },
   }),
+  createConfig({
+    input: multiInput,
+    output: {
+      dir: 'dist',
+      format: 'esm',
+      entryFileNames: 'redux-saga-[name].[format].mjs',
```
Andarist: are u aware of the fact that generated chunks (not entry files) contain a hash in their filenames and you will have to handle it? redux-saga update is probably going to change the hash each time

Andarist: I'm also not fond of using .mjs for this, just make it smth like:
```js
entryFileNames: 'redux-saga-[name]. esmodules-browsers.js',
```

Andarist:
This will need a rebase now - please move the new rollup configuration to this file: https://github.com/redux-saga/redux-saga/blob/6d7cb99cf489da0b4de72eee44b7333d3c8f2cba/packages/redux-saga/rollup.config.js

TimvdLippe:
Sadly I haven't found the time to process your review feedback. @martijnjanssen do you have time to go through the feedback and resolve them?

Andarist:
@TimvdLippe @martijnjanssen I've merged with upstream master & tweaked the build to output those bundles correctly. From my POV this is ready to be merged (I'd appreciate second look though).

One question got sort of unanswered - are u aware of the fact that generated chunks (not entry files) contain a hash in their filenames and you will have to handle it? redux-saga update is probably going to change the hash each time

martijnjanssen:
So, I just tested it, and it works flawlessly. I'm using the `redux-saga-core.esmodules-browsers.js` and `redux-saga-effects.esmodules-browsers.js` files when importing saga, I don't see where I would run into issues with the chunks having hashes in their filenames. Am I missing something after all?

Issue #2, opened 2015-12-18
Missing put / delay imports in README examples


quicksnap:
In the README, `put` and `delay` are missing from `import`. Is this intentional? Being new to this library, I was immediately thrown off, and went searching where they came from..

```js
import { take, call } from 'redux-saga'
// sagas/index.js
function* incrementAsync() {

  while(true) {

    // wait for each INCREMENT_ASYNC action  
    const nextAction = yield take(INCREMENT_ASYNC)

    // call delay : Number -> Promise
    yield delay(1000)

    // dispatch INCREMENT_COUNTER
    yield put( increment() )
  }

}

export default [incrementAsync]
```

quicksnap:
I see that `delay` is just an example function, but `put` is part of the API. I'll whip up a PR to adjust the readme

PR #1718, opened 2019-01-03
Add codesanbox examples

TruongNhatVo:


restrry:
branch updated.

examples/async/README.md
```diff
@@ -0,0 +1,5 @@
+# Run locally
+`npm start`
+
+# Open in browser
+[Click here](https://codesandbox.io/s/github/redux-saga/redux-saga/tree/master/examples/async)
```
restrry: to test use `add-codesanbox-examples` branch name: https://codesandbox.io/s/github/redux-saga/redux-saga/tree/add-codesanbox-examples/examples/shopping-cart

examples/real-world/store/configureStore.prod.js
```diff
@@ -1,14 +1,11 @@
 import { createStore, applyMiddleware } from 'redux'
 import createSagaMiddleware, { END } from 'redux-saga'
+import sagaMonitor from 'saga-monitor-example'
```
shinima: Should we change configureStore.dev.js as well?

restrry: for sure, it should.
now when I add it I can see that `saga-monitor-example` is not compiled and not ready to be consumed. need to figure out how to fix it

Andarist: this is somewhat a problem because this example is marked as private now - we can add a build step & publish it, that way codesandbox will know how to consume it

restrry: shouldn't we switch to https://github.com/redux-saga/redux-saga-devtools package?

Andarist: yeah, that would be nice too - i dont know how up to date it is though & don't have time personally to adjust those examples to use it, publishing saga-monitor-example has just seemed like smaller effort

restrry:
would you mind to review again? now only `real-world` uses `saga-monitor-example`. also I had to add build step for saga-monitor. 
as `saga-montor-example` uses other packages I disabled `parallel` build (on my laptop build time increased from 9.6s to 14.4s, more or less is ok).
when we switch to https://github.com/redux-saga/redux-saga-devtools we can get rid of saga-monitor-example

examples/error-demo/README.md
```diff
@@ -0,0 +1,7 @@
+# Run locally
+`npm start`
+
+# Open in browser
+[Click here](https://codesandbox.io/s/github/redux-saga/redux-saga/tree/master/examples/error-demo)
+
+*Note* babel-plugin doesn't work yet on codesanbox.
```
Andarist: do you know what's the problem? i guess codesandbox doesnt support babel customization?

examples/real-world/store/configureStore.prod.js
Andarist: do we plan to publish this?

package.json
```diff
@@ -9,7 +9,7 @@
     "test": "npm run test-only",
     "check": "run-s lint test",
     "clean": "lerna run --parallel clean && lerna clean --yes && rimraf node_modules",
-    "build": "lerna run --parallel build",
```
Andarist: was this causing a problem? 🤔

restrry: when I added  a build step for  `saga-monitor-example` running `npm run build` started failing. from `lerna` [docs](https://github.com/lerna/lerna/tree/master/commands/exec#--parallel)
```
--parallel
...
Similar to --stream, but completely disregards concurrency and topological sorting,
```

examples/real-world/store/configureStore.prod.js
restrry: we can actually, don't see a lot of values tho. most likely ppl want to use something more advanced, like `redux-saga-devtools `

PR #1726, opened 2019-01-14
Add hint about possibility to use babel plugin to improve reported st…

Andarist:


packages/core/__tests__/channel-recipes.js
```diff
@@ -37,7 +37,11 @@ test('action channel', () => {
 })
 
 test('error check when constructing actionChannels', () => {
-  const middleware = sagaMiddleware()
+  const middleware = sagaMiddleware({
+    onError: err => {
+      expect(err.message).toMatchInlineSnapshot(`"actionChannel(pattern,...): argument pattern is not valid"`)
```
restrry: probably we should remove error message assertion below

packages/core/__tests__/interpreter/onerror.js
```diff
@@ -1,10 +1,17 @@
+/* eslint-disable no-console */
 import { createStore, applyMiddleware } from 'redux'
 import sagaMiddleware from '../../src'
 import * as io from '../../src/effects'
+
 test('saga onError is optional', () => {
+  let consoleError = console.error
```
restrry: shouldn't it be a separate test? doesn't look like `saga onError is optional`:)

packages/core/__tests__/channel-recipes.js
Andarist: That's true - it could be removed, but I didnt want to change too much here. The proper way would be to remove latter assertion AND making sure that this one is happening (probably by wrapping the whole thing in `return new Promise(...)` and resolving it in `onError`.

This could be done, but as I've said - I didn't want to change too much, I just finally got annoyed by messy output which has been produced and I've fixed it in the fastest way I could.

packages/core/__tests__/interpreter/onerror.js
Andarist: that's also kinda true - this relation is implicit, maybe we could just add a note to the `test`'s description, smth like:
```js
test('saga onError is optional (the default is console.error)', () => {
  // ...
})
```

packages/core/src/internal/sagaError.js
```diff
@@ -64,10 +69,14 @@ export const setCrashedEffect = effect => {
 export const toString = () => {
   const [firstSaga, ...otherSagas] = sagaStack
   const crashedEffectLocation = firstSaga.crashedEffect ? effectLocationAsString(firstSaga.crashedEffect) : null
-  const errorMessage = `The above error occurred in task ${sagaLocationAsString(firstSaga.meta)}${
+  let errorMessage = `The above error occurred in task ${sagaLocationAsString(firstSaga.meta)}${
     crashedEffectLocation ? ` \n when executing effect ${crashedEffectLocation}` : ''
   }`
 
+  if (process.env.NODE_ENV !== 'production' && !firstSaga.meta.location) {
```
Andarist: @restrry u know that logic better, is this condition a good one for this check? Can it yield false positives?

packages/core/__tests__/interpreter/onerror.js
restrry: 👍

packages/core/src/internal/sagaError.js
restrry: unfortunately it doesn't work as expected. for errors raised in helpers we don't have `location`, as effect is `hidden` in library code. for example, one of such cases:
```js
  yield retry(3, 10, function() {
    undefinedIsNotAFunction()
  })
```
as an alternative solution we can use
```js
sagaStack.some(s => s.meta.location)
```
which is a bit better, but still could lead to false positives.

Andarist: hm, shouldnt "lastSaga" have `location` nearly in all cases then? 🤔

restrry: I suppose so

restrry: @Andarist do you want me to finish work on this branch?

Andarist: That would definitely help, I'll try to dig into the babel location stuff in the meantime

szb512 (commented):
Maybe we should make changes...

szb512 (approved):

szb512 (approved):

szb512:
These changes are already included in #1797 .

anandfresh (approved):

Issue #684, opened 2016-12-11
Different ways to watch actions


schuchowsky:
Hello guys, while reading the documentation and tutorials, I see different ways to watch actions and trigger sagas:

*While take*
```javascript
function* watchTodos() {
  while (yield take('FETCH_TODOS')) {
    yield fork(loadTodos);
  }
}
```

*While true*
```javascript
function* watchFetch() {
  while (true) {
    yield take(FETCH_POSTS)
    yield call(fetchPosts)
  }
}
```

And I am, particularly, using this syntax:
```javascript
export default function* watchFiles() {
  yield [
    takeEvery('DELETE_FILE', deleteFile),
    takeEvery('LIST_FILES', listFiles),
  ]
}
```

So, my doubt is that I don't understand the need of having that while(true) loop, if we have `takeEvery` function. Am I going to miss something using my approach?

kuy:
@schuchowsky  First, I'd like to explain a difference between `while(true)` and `while(yield take(...))`. If you don't need to use an action that is yielded from `take` effect, you may use `while(yield take(...))` pattern. If you want to use an action, please use `while(true)` with `var action = yield take(...)`.

> So, my doubt is that I don't understand the need of having that while(true) loop, if we have takeEvery function.

Yes, you're right. But sometimes I feel it's overkill to always use `takeEvery` instead of while loop in small sagas.

schuchowsky:
@kuy Oh I see... so in the example above, the `while(true)` could be replaced by `while (yield take (FETCH_POSTS)`, because the function yielded from take effect is not being used in that case, right?

Overkill? Hm... What would be the correct use for it then? 

I feel like it's more readable using `takeEvery` because in one function I declare all my watchers.

kuy:
> the while(true) could be replaced by while (yield take (FETCH_POSTS), because the function yielded from take effect is not being used in that case, right?

Yes!

> Overkill? Hm... What would be the correct use for takeEvery then?

To use `takeEvery`, you need to split a saga. This is what I feel it's overkill for small sagas. For example, I use `takeEvery` to avoid dropping dispatched action.

```js
function* watchFetch() {
  while (true) {
    yield take(FETCH_POSTS)
    yield call(fetchPosts)
  }
}
```

[...]

PR #1744, opened 2019-01-25
Add `takeLatestPerKey` and `takeLeadingPerKey` helpers

### LINKED ISSUE #694: tracking unique redux-saga tasks
I am using sagas to track multiple async tasks, but there is one problem I haven't been able to completely solve:

```
function* performTask1() {
    // Some logic here to takeLatest for the relevant component

    // check if component id matches?

    // Only perform API call with the latest
    const { result } = yield takeLatest('doAsync2')
}

function* performTask2() {
    const { result } = yield call(api, args)
    // do something with results (not relevant)
}

function* watchAsyncTasks() {
    yield takeEvery('doAsync2', performTask2)
    yield takeEvery('doAsync1', performTask1)
}

```
1. componentA dispatches doAsync1
2. componentB dispatches doAsync1

3. component C dispatches doAsync2 (for good measure)

4. componentA dispatches doAsync1
5. componentB dispatches doAsync1

How can I use sagas to ensure that only sagas 3, 4, and 5 complete their API call?

### LINKED ISSUE #1751: Scope tasks from `takeLatest` and `takeLeading` by ID
I opened PR https://github.com/redux-saga/redux-saga/pull/1744 to address this idea and to couple that PR to an active issue.

This is somewhat of a bump of https://github.com/redux-saga/redux-saga/issues/694 - but it's been a couple of years since then and I still think there is a good case to be made for an out-of-the-box solution here.

Let's say you have a saga whose primary function is fetching a single user based on an ID - and let's say that you want to fetch multiple users around the same time.

If you want to ensure that a maximum of one fetching task is running per ID, you might want something like `takeLatest` and `takeLeading` but scoped by something that we select from the action that spawns the saga (like an ID).

Examples:

```js
function* fetchUser(action) {
  const { payload: { id } } = action
  const response = yield call(getUser, id)
  // ...
}

function* rootSaga() {
  yield takeLatestPerKey('FETCH_USER', fetchUser, action => {
    const { payload: { id } } = action
    return id
  })
}
```

```js
function* fetchUser(action) {
  const { payload: { id } } = action
  const response = yield call(getUser, id)
  // ...
}

function* rootSaga() {
  yield takeLeadingPerKey('FETCH_USER', fetchUser, action => {
    const { payload: { id } } = action
    return id
  })
}
```

I'd love to hear your thoughts on this issue.

Thank you for this amazing library and all the hard work that goes into maintaining it.

therealparmesh:
| Fixed Issues?            |  Fixes #1751, Fixes #694  |
| Tests Added + Pass?      |  Yes  |

**I created CodeSandbox examples for what I'm currently calling [`takeLatestPerKey`](https://codesandbox.io/s/kwr75ynq7v) and [`takeLeadingPerKey`](https://codesandbox.io/s/j38593lp49).**

Example usage:
`yield takeLatestPerKey("ACTION", worker, function selectId({ payload: { id } }) { return id; });`
`yield takeLeadingPerKey("ACTION", worker,  function selectId({ payload: { id } }) { return id; });`

Awhile ago, I came across the same need as described in https://github.com/redux-saga/redux-saga/issues/694. I also opened a new issue in https://github.com/redux-saga/redux-saga/issues/1751 to accompany this PR.

I borrowed the solution written by @Andarist based on `takeLatest` from the https://github.com/redux-saga/redux-saga/issues/694. It was simple enough to imagine what a similar solution based on `takeLeading` would look like. Both of these helpers take a key selector function - which takes the action that spawns the saga as a parameter and that returns a key. They then effectively scope a `takeLatest`/`takeLeading` to this key.

I'd like to contribute these helpers back to `redux-saga` as they have been very useful in the places I've worked. I hope there's interest in adding these helpers in, as I believe they would benefit many people with similar needs.

### DIFF
--- a/packages/core/__tests__/sagaHelpers/takeLatestPerKey.js
+++ b/packages/core/__tests__/sagaHelpers/takeLatestPerKey.js
@@ -1,0 +1,132 @@
+import sagaMiddleware, { END } from '../../src'
+import { createStore, applyMiddleware } from 'redux'
+import { arrayOfDeferred } from '@redux-saga/deferred'
+import { take, cancel, takeLatestPerKey } from '../../src/effects'
+test('takeLatestPerKey', () => {
+  const defs = arrayOfDeferred(6)
+  const actual = []
+  const middleware = sagaMiddleware()
+  const store = applyMiddleware(middleware)(createStore)(() => {})
+  middleware.run(root)
+
+  function selectKey(action) {
+    return action.payload.id
+  }
+
+  function* root() {
+    const task = yield takeLatestPerKey('ACTION', worker, selectKey, 'a1', 'a2')
+    yield take('CANCEL_WATCHER')
+    yield cancel(task)
+  }
+
+  function* worker(arg1, arg2, action) {
+    const idx = action.payload.data - 1
+    const response = yield defs[idx].promise
+    actual.push([arg1, arg2, response])
+  }
+
+  return Promise.resolve()
+    .then(() => {
+      store.dispatch({
+        type: 'ACTION',
+        payload: {
+          id: 1,
+          data: 1,
+        },
+      })
+    })
+    .then(() => {
+      store.dispatch({
+        type: 'ACTION',
+        payload: {
+          id: 1,
+          data: 2,
+        },
+      })
+    })
+    .then(() => defs[0].resolve('w-1'))
+    .then(() => {
+      store.dispatch({
+        type: 'ACTION',
+        payload: {
+          id: 2,
+          data: 4,
+        },
+      })
+    })
+    .then(() => defs[1].resolve('w-2'))
+    .then(() => {
+      store.dispatch({
+        type: 'ACTION',
+        payload: {
+          id: 2,
+          data: 3,
+        },
+      })
+    })
+    .then(() => defs[3].resolve('w-4'))
+    .then(() => defs[2].resolve('w-3'))
+    .then(() => {
+      store.dispatch({
+        type: 'ACTION',
+        payload: {
+          id: 1,
+          data: 5,
+        },
+      })
+      /*
+        We immediately cancel the watcher after firing the action
+        The watcher should be cancelled after this
+        no further task should be forked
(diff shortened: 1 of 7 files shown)


polomoshnov:
Don't you think that the suggested implementation will miss the actions dispatched between `yield take` and `yield call`, as described [here](https://github.com/redux-saga/redux-saga/issues/684#issuecomment-266286205) and [here](https://github.com/redux-saga/redux-saga/issues/694#issuecomment-412929281).

And also, the `tasks` map in this implementation can grow indefinitely... It may be a good idea to wait for the task to complete (`yield join(task)`) and then to delete that task from the `tasks` map? (As seen [here](https://stackoverflow.com/a/56609932).)

polomoshnov:
Haven't tested this but the above mentioned CodeSandbox examples might be changed as follows:

```
function takeLeadingPerKey(patternOrChannel, worker, keySelector, ...args) {
	return fork(function* () {
		const tasks = {};

		yield takeEvery(patternOrChannel, function* (action) {
			const key = yield call(keySelector, action);

			if (!(tasks[key] && tasks[key].isRunning())) {
				tasks[key] = yield fork(worker, ...args, action);
				
				yield join(tasks[key]);

				if (tasks[key] && !tasks[key].isRunning()) {
					delete tasks[key];
				}
			}
		});
	});
}

function takeLatestPerKey(patternOrChannel, worker, keySelector, ...args) {
	return fork(function* () {
		const tasks = {};

		yield takeEvery(patternOrChannel, function* (action) {
			const key = yield call(keySelector, action);

			if (tasks[key]) {
				yield cancel(tasks[key]);
			}

			tasks[key] = yield fork(worker, ...args, action);

			yield join(tasks[key]);

			if (tasks[key] && !tasks[key].isRunning()) {
				delete tasks[key];
			}
		});
	});
}
```

therealparmesh:
What should we do with this PR?

Andarist:
I'm sorry that you have put work into this PR but I think that in the end, this doesn't belong to the core and it should live in the userland. Hope you understand.

Issue #1742, opened 2019-01-24
babel-plugin-redux-saga cannot read property start of undefined


royipressburger:
#### Description of the bug/issue
Hi, 

I just updated to version `1.0.0` from `1.0.0-rc-0`.
Everything is working good until I added the `babel-plugin` then I'm getting exception:
`TypeError: Cannot read property 'start' of undefined`

here is the stactrace:
```
TypeError: Cannot read property 'start' of undefined
    at calcLocation (/Users/project/node_modules/babel-plugin-redux-saga/src/index.js:57:26)
    at PluginPass.FunctionExpression (/Users/project/node_modules/babel-plugin-redux-saga/src/index.js:116:26)
    at newFn (/Users/project/node_modules/@babel/core/node_modules/@babel/traverse/lib/visitors.js:193:21)
    at NodePath._call (/Users/project/node_modules/@babel/core/node_modules/@babel/traverse/lib/path/context.js:53:20)
    at NodePath.call (/Users/project/node_modules/@babel/core/node_modules/@babel/traverse/lib/path/context.js:40:17)
    at NodePath.visit (/Users/project/node_modules/@babel/core/node_modules/@babel/traverse/lib/path/context.js:88:12)
    at TraversalContext.visitQueue (/Users/project/node_modules/@babel/core/node_modules/@babel/traverse/lib/context.js:118:16)
    at TraversalContext.visitMultiple (/Users/project/node_modules/@babel/core/node_modules/@babel/traverse/lib/context.js:85:17)
    at TraversalContext.visit (/Users/project/node_modules/@babel/core/node_modules/@babel/traverse/lib/context.js:144:19)
    at Function.traverse.node (/Users/project/node_modules/@babel/core/node_modules/@babel/traverse/lib/index.js:94:17)
    at NodePath.visit (/Users/project/node_modules/@babel/core/node_modules/@babel/traverse/lib/path/context.js:95:18)
    at TraversalContext.visitQueue (/Users/project/node_modules/@babel/core/node_modules/@babel/traverse/lib/context.js:118:16)
    at TraversalContext.visitSingle (/Users/project/node_modules/@babel/core/node_modules/@babel/traverse/lib/context.js:90:19)
    at TraversalContext.visit (/Users/project/node_modules/@babel/core/node_modules/@babel/traverse/lib/context.js:146:19)
    at Function.traverse.node (/Users/project/node_modules/@babel/core/node_modules/@babel/traverse/lib/index.js:94:17)
[...]

PR #1745, opened 2019-01-27
Babel-plugin dont handle empty location

restrry:
| Fixed Issues?            | https://github.com/redux-saga/redux-saga/issues/1742 |
| Patch: Bug Fix?          |  bug fix  |

`FunctionExpression` didn't have empty location check (when code is not part of original code, but generated by other babel-plugins) so I added

### DIFF
--- a/packages/babel-plugin-redux-saga/test/fixtures/preset-env/source.js
+++ b/packages/babel-plugin-redux-saga/test/fixtures/preset-env/source.js
@@ -5,3 +5,13 @@ function* test1() {
 function* test2() {
   yield 2
 }
+
+class Component extends React.PureComponent {
+  *getData() {
+    yield 1
+  }
+  render() {
+    const data = [...this.getData()]
+    return data
+  }
+}
--- a/packages/babel-plugin-redux-saga/src/index.js
+++ b/packages/babel-plugin-redux-saga/src/index.js
@@ -91,11 +91,13 @@ module.exports = function(babel) {
      *  })
      */
     FunctionDeclaration(path, state) {
-      if (!isSaga(path)) return
+      var node = path.node
+      if (!node.loc || !isSaga(path)) return
 
-      var functionName = path.node.id.name
+      var functionName = node.id.name
       var filename = getFilename(state.file.opts, state.opts.useAbsolutePath)
-      var locationData = calcLocation(path.node.loc, filename)
+
+      var locationData = calcLocation(node.loc, filename)
 
       const extendedDeclaration = createLocationExtender(t.identifier(functionName), locationData)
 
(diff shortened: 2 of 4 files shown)


packages/babel-plugin-redux-saga/test/fixtures/preset-env/source.js
```diff
@@ -5,3 +5,13 @@ function* test1() {
 function* test2() {
   yield 2
 }
+
+class Component extends React.PureComponent {
```
Andarist: is this testing the reported issue though? the reporter mentions async functions being involved (which makes them trnaspiled to generators)

isnt this code here having location? (its not inserted by babel)

restrry: by whatever reason `async/await` after being processed by `regenerator` has `path.node.generator === true`.
at the same time with `env` preset both `async/await` and `function*` are replaced by regenerator and don't have locations [demo](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEDCD2BbADvAdgUzQFwIzQwA9ssATGAJQzGGwDoAFAVwCcMEV0ttoBvAKGjRIATzTBoAcwzYAImGxgAFAEo-godDbZWaYQHcwASx4MWSIxAx02EeCABuGJWiYgQKjQF8NbNKQwWVXVNaGB0CB5SBTBoAF5oAG06FOwAC0s6aTkY1QBdDSFtXWhoxW9-H35QSBgOVEwcACYCYjJKalpGVnYkBu4Q4QgxCWz5RWCBUOKWPTBDE2gzCysbDDtHZ1d3TyEfIrJAycKwiKiY-KSUunTMsdyVAumZErKwCq8gA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=6.26.0&envVersion=1.6.2)

Andarist: hm, this seems terrible for our use case as preset-env is really popular (as it should be - it's primary use case for babel now)

I'll try look into this as soon as possible to see how babel handle this and if there is anything we can do to overcome this

restrry: @Andarist updated my comment. to test the case reference the original issue

szb512:
I suggest that you go here to find out how to upgrade to the latest version of babel:
https://babeljs.io/docs/en/v7-migration

szb512:
Maybe then you can try again.

restrry:
@szb512 actually we aim to support both babel v6 and v7. later we will remove v6 when most projects drop its support

PR #1754, opened 2019-01-30
Fixes #1753 issue when the first arg of call effect's function is a Symbol

### LINKED ISSUE #1753: call effect breaks when the first argument is a Symbol in v1
### Steps to reproduce

`yield call(fn, Symbol('arg1'))`

When you try to pass a Symbol argument to the function using `call` effect an error is thrown

```
    TypeError: Cannot convert a Symbol value to a string
      at call (node_modules/@redux-saga/core/dist/chunk-e088dd6b.js:511:84)
```

The issue, that in the `call` function there is the following code:

```
    check(fnDescriptor, isNotDelayEffect, "instead of writing `yield call(delay, " + args[0] + ")` where delay is an effect from `redux-saga/effects` you should write `yield delay(" + args[0] + ")`");
```
And when you try to concat a Symbol (and convert it to a string) an error is thrown.

mike1808:
Fixes #1753 issue when the first arg of call effect is a Symbol.

### DIFF
--- a/packages/core/src/internal/io.js
+++ b/packages/core/src/internal/io.js
@@ -141,9 +141,9 @@ export function call(fnDescriptor, ...args) {
     check(
       fnDescriptor,
       isNotDelayEffect,
-      `instead of writing \`yield call(delay, ${
-        args[0]
-      })\` where delay is an effect from \`redux-saga/effects\` you should write \`yield delay(${args[0]})\``,
+      `instead of writing \`yield call(delay, ${String(
+        args[0],
+      )})\` where delay is an effect from \`redux-saga/effects\` you should write \`yield delay(${String(args[0])})\``,
     )
     validateFnDescriptor('call', fnDescriptor)
   }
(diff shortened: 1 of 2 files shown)


Andarist:
Could you describe how did it fail without this lazily evaluation?

packages/core/src/internal/io.js
```diff
@@ -141,9 +141,10 @@ export function call(fnDescriptor, ...args) {
     check(
       fnDescriptor,
       isNotDelayEffect,
-      `instead of writing \`yield call(delay, ${
-        args[0]
-      })\` where delay is an effect from \`redux-saga/effects\` you should write \`yield delay(${args[0]})\``,
+      () =>
+        `instead of writing \`yield call(delay, ${
+          args[0]
```
Andarist: this will still throw unhelpful error for
```js
yield call(delay, Symbol())
```

IMHO better solution would be to just stringify explicitly args[0] - `String(args[0])`

mike1808: Ah, didn't think about that. I'll change.

Andarist:
Thanks!

PR #1757, opened 2019-01-31
Handle undefined values and prevent infinite loops

### LINKED ISSUE #1750: Browser freezes in 1.0.0
### Steps to reproduce
Needs to be investigated.

After upgrading from 0.16.2 to 1.0.0 and running the app, the browser (Chrome) gets stuck in what seems like a `while` loop and ceases to respond. Other than changes to the imports of `delay` from `redux-saga` to `redux-saga/effects` and `task.done` to `task.toPromise()`, no code changes were done. Reverting back to 0.16.2 resolves the issue.

Needs to be investigated.

#### Example
Needs to be created.

### Actual results
The browser becomes unresponsive.

### The Expected results
A responsive browser.

### Environment information
1.0.0

rybon:
Fixes #1750.

This slight change prevents the code from entering an infinite loop and causing the browser to freeze. This behavior is currently triggered by returning effects (Promises) that resolve to `Promise.resolve()` or `Promise.reject()`. The resulting value of these calls is `undefined`. This change accounts for that and terminates the call stack, preventing the infinite loop.

### DIFF
--- a/packages/core/src/internal/proc.js
+++ b/packages/core/src/internal/proc.js
@@ -163,7 +163,7 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
 
     // Completion callback passed to the appropriate effect runner
     function currCb(res, isErr) {
-      if (effectSettled) {
+      if (effectSettled || res === undefined) {
         return
       }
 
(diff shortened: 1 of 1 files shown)


rybon:
Apparently this breaks the test suite.

@Andarist could you have a look at this issue? I really want to prevent this issue from popping up again. Browser freezes should never ever happen.

Andarist:
@rybon could you prepare a simple repro case showcasing the problem?

rybon:
Sure, I can do that.

rybon:
https://codesandbox.io/s/o95y24znnz

`yield () => Promise.resolve()` works, `takeLatest(SOME_TYPE, () => Promise.resolve())` does not.

PR #1758, opened 2019-02-01
Fix endless loop while createTaskIterator gets a function that return…

### LINKED ISSUE #1750: Browser freezes in 1.0.0

shinima:
…s a Promise resolved with undefined.

| Fixed Issues?            | Fixes #1750 |

In the previous implementation, if the function passed to `createTaskIterator` returns a promise which resolves to be undefined, the iterator created by `createTaskIterator` will lead to an endless loop.

In the PR, I use `let resolved = false` to make sure that the promise is interpreted at most once.

### DIFF
--- a/packages/core/src/internal/effectRunnerMap.js
+++ b/packages/core/src/internal/effectRunnerMap.js
@@ -37,10 +37,17 @@ function createTaskIterator({ context, fn, args }) {
       return result
     }
 
-    const next = (value = result) => ({
-      value,
-      done: !is.promise(value),
-    })
+    let resolved = false
+
+    const next = arg => {
+      if (!resolved) {
+        resolved = true
+        // Only promises returned from fork will be interpreted. See #1573
+        return { value: result, done: !is.promise(result) }
+      } else {
+        return { value: arg, done: true }
+      }
+    }
 
     return makeIterator(next)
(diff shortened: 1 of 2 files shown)


packages/core/src/internal/effectRunnerMap.js
```diff
@@ -37,10 +37,17 @@ function createTaskIterator({ context, fn, args }) {
       return result
     }
 
-    const next = (value = result) => ({
```
shinima: If the promise resolves to be `undefined`, next time `next` is called with the resolved value, the default argument `result` is used.

Andarist:
Thanks ❤️

PR #1759, opened 2019-02-01
Bug Fixed ValidEffect in @redux-saga/types

Dai7Igarashi:
| Fixed PR?            | `Fixes #1655`  |
| Tests Added + Pass?      | No |

Variable `ValidEffect` might not be defined in `@redux-saga/types`, so it caused an import error as using TypeScript.  

I wonder if another way, fix `import { Effect } from '@redux-saga/types'` in `is/index.d.ts`, might be better.

### DIFF
--- a/packages/is/index.d.ts
+++ b/packages/is/index.d.ts
@@ -1,9 +1,9 @@
-import { ActionPattern, Buffer, Channel, GuardPredicate, Pattern, Task, ValidEffect } from '@redux-saga/types'
+import { ActionPattern, Buffer, Channel, GuardPredicate, Pattern, Task, Effect } from '@redux-saga/types'
 
 export const array: GuardPredicate<Array<any>>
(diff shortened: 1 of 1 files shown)


packages/is/index.d.ts
```diff
@@ -3,7 +3,7 @@ import { ActionPattern, Buffer, Channel, GuardPredicate, Pattern, Task, ValidEff
 export const array: GuardPredicate<Array<any>>
 export const buffer: GuardPredicate<Buffer<any>>
 export const channel: GuardPredicate<Channel<any>>
-export const effect: GuardPredicate<ValidEffect>
+export const effect: GuardPredicate<ValidEffect<any>>
```
aikoven: I think we can just use `Effect` here, without adding new types.

Dai7Igarashi:
Thank you for reviewing my PR.
I just changed `ValidEffect` to `Effect`.

PR #1769, opened 2019-02-10
typings: better inference of action types from action pattern

### LINKED ISSUE #1740: TS typedefs: take(array) does not work with heterogenous redux-action actions
Was on 0.16 and tried the 1.0.0, and I'm getting this TS error when trying to mix redux-action actions with distinct payload types:

image

It works fine if the payload has same type:

image

It used to work fine on 0.16

Didn't try but afaik a workaround is using toString:

image

Not sure how to fix this, but  `StringableActionCreator<any>` does work in `ActionSubPattern`

aikoven:
Fixes #1740.

This PR adds a new type

```ts
type ActionMatchingPattern<P extends ActionPattern>
```

that calculates matched action types (a union) given action pattern, which may be an array of stringable action creators or predicates.

This type is then used in additional overloads for `takeEvery`, `takeLatest`, `takeLeading` as well as `throttle` and `debounce` that support inferring action types and avoid the need to specify them explicitly.

### DIFF
--- a/packages/core/__tests__/typescript/effects.ts
+++ b/packages/core/__tests__/typescript/effects.ts
@@ -28,6 +28,7 @@ import {
   debounce,
 } from 'redux-saga/effects'
 import { Action, ActionCreator } from 'redux'
+import { StringableActionCreator, ActionMatchingPattern } from '@redux-saga/types'
 
 interface MyAction extends Action {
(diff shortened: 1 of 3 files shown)


Andarist:
Thanks ❤️

Issue #1608, opened 2018-09-19
When try to run the examples getting a compilation error.


anikakalbrun:
### Steps to reproduce
I have cloned the project
went to the async directory: cd /examples/async
installed node_modules: npm install
tried to run the project: npm run start

This is the output I am getting:
--------------------------------------------------------------------------------------------------

^CTNS6237L:async a$ npm run start

> async-example@1.0.0-beta.1 start /Users/a/IdeaProjects/redux-saga/examples/async
> webpack-dev-server

ℹ ｢wds｣: Project is running at http://localhost:3000/
ℹ ｢wds｣: webpack output is served from /
[BABEL] Note: The code generator has deoptimised the styling of /Users/a/IdeaProjects/redux-saga/examples/async/node_modules/react-dom/cjs/react-dom.development.js as it exceeds the max of 500KB.
✖ ｢wdm｣: Hash: 4e8714d1ecdbfb4c7dfc
Version: webpack 4.19.1
Time: 5300ms
Built at: 2018-09-19 12:54:43
     Asset       Size  Chunks             Chunk Names
   main.js   1.64 MiB    main  [emitted]  main
index.html  292 bytes          [emitted]  
Entrypoint main = main.js
[./node_modules/@babel/polyfill/lib/index.js] 893 bytes {main} [built]
[./node_modules/loglevel/lib/loglevel.js] 6.84 KiB {main} [built]
[./node_modules/url/url.js] 22.3 KiB {main} [built]
[./node_modules/webpack-dev-server/client/index.js?http://localhost:3000] (webpack)-dev-server/client?http://localhost:3000 7.79 KiB {main} [built]
[./node_modules/webpack-dev-server/client/overlay.js] (webpack)-dev-server/client/overlay.js 3.58 KiB {main} [built]
[./node_modules/webpack-dev-server/client/socket.js] (webpack)-dev-server/client/socket.js 1.05 KiB {main} [built]
[./node_modules/webpack/hot/dev-server.js] (webpack)/hot/dev-server.js 1.65 KiB {main} [built]
[./node_modules/webpack/hot/emitter.js] (webpack)/hot/emitter.js 75 bytes {main} [built]
[./node_modules/webpack/hot/log-apply-result.js] (webpack)/hot/log-apply-result.js 1.31 KiB {main} [built]
[0] multi (webpack)-dev-server/client?http://localhost:3000 (webpack)/hot/dev-server.js ./src/main.js 52 bytes {main} [built]
[...]

PR #1782, opened 2019-02-20
Fix issue #1608

jpggvilaca:
This fixes issue #1608.

There were 2 devDependencies needed for the examples to work.


Andarist:
Thanks for the PR, I believe though that the issue should be resolved by this https://github.com/redux-saga/redux-saga/commit/71d2caa6a558036dfd526d2aac41734da587022b .

PR #1785, opened 2019-02-21
Prevent cancellation of ended joined tasks

romaingweb:
| Fixed Issues?            |    |

Cancelling a join task caused all nested forked tasks to be cancelled as
well, regardless of their status. This means an finished/ended forked
task was cancelled unnecessarily.

Because the `joiners` property is changed from an array to `null` when the task is ended, the
subsequent cancellation that tries to remove the joiner from `joiners` caused a `Cannot read property 'indexOf' of null` error.

We added a guard in the cancel callback that ensures the task is
running.

We wonder if the joiners property should be reset to `[]` instead of
`null`. It's another way of fixing this issue.

### DIFF
--- a/packages/core/__tests__/interpreter/race.js
+++ b/packages/core/__tests__/interpreter/race.js
@@ -138,3 +138,25 @@ test('saga race between sync effects', () => {
       expect(actual).toEqual(expected)
     })
 })
+test('saga race cancelling joined tasks', () => {
+  const middleware = sagaMiddleware()
+  applyMiddleware(middleware)(createStore)(() => {})
+
+  function* genFn() {
+    yield io.race({
+      join: io.join([
+        yield io.fork(function*() {
+          yield io.delay(10)
+        }),
+        yield io.fork(function*() {
+          yield io.delay(100)
+        }),
+      ]),
+      timeout: io.delay(50),
+    })
+  }
+
+  const task = middleware.run(genFn)
+
+  return Promise.resolve().then(() => task.toPromise())
+})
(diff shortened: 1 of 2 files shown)


packages/core/__tests__/interpreter/race.js
```diff
@@ -138,3 +138,25 @@ test('saga race between sync effects', () => {
       expect(actual).toEqual(expected)
     })
 })
+test('saga race cancelling joined tasks', () => {
+  const middleware = sagaMiddleware()
+  applyMiddleware(middleware)(createStore)(() => {})
+
+  function* genFn() {
```
Andarist: Could you rewrite this test (or add another one) to cover this in more explicit manner? The fact that race cancels losers is an implicit one, would be good to have a straightforward test for this.

packages/core/__tests__/interpreter/race.js
```diff
@@ -138,3 +138,25 @@ test('saga race between sync effects', () => {
...
+          yield io.delay(100)
+        }),
+      ]),
+      timeout: io.delay(50),
+    })
+  }
+
+  const task = middleware.run(genFn)
+
+  return Promise.resolve().then(() => task.toPromise())
```
Andarist: as there is no assertion here it would be at least good to add a comment here that the true assertion is that it doesnt fail - or add
```
.catch(() => {
  throw new Error("This shouldn't happen, this saga shouldn't crash.")
})
```
(or similar)

Andarist:
Thanks for the PR, I left some minor comments to be addressed

PR #1791, opened 2019-02-23
Don't stringify args[0] in checks of io.call().

### LINKED ISSUE #1776: saga logging breaks the app
Development logging cant handle all values

Run the following saga
```
const mySaga = function * (): Saga<void> {
	const data = Object.create(null)
	yield call(console.log, data)
}
```
or just
```String(Object.create(null))```
which is actually beeing evaluated in https://github.com/redux-saga/redux-saga/blob/65034cc8b08ab5a6ff1e7441f3062960615f368d/packages/core/src/internal/io.js#L146 

### Actual results
The logging  that should help the developers breaks the app.
Before this issue I thought that you can put just anything into `String()` or `Boolean()`. But that is not true.

### The Expected results
A working app should not break because of dev logs

### Thoughts toward the solution

1. Fix our app, so that redux-saga works with normal objects. It is the first thing that I did. However, it really took me a while to see where the problem emerged after upgrading to redux-saga@1 - very frustrating. Additionally `Object.create(null)` was sent via action in a library, redux middleware, that we use, so I couldnt actually remove it, just create a workaround for it. The irony is that the app works fine in production but fails in development because of a log that is not actually being logged
2. create a helper function , e.g. `tryString`, with a try-catch clause that would be used to convert values to string.

### Environment information

- redux-saga version 1.0.1

### PR DESCRIPTION
**Pull request by shinima**
| Fixed Issues?            | Fixes #1776 |
| Tests Added + Pass?      | 👍 |

### DIFF
--- a/packages/core/src/internal/io.js
+++ b/packages/core/src/internal/io.js
@@ -140,10 +140,11 @@ const isNotDelayEffect = fn => fn !== delay
 
 export function call(fnDescriptor, ...args) {
   if (process.env.NODE_ENV !== 'production') {
+    const arg0 = typeof args[0] === 'number' ? args[0] : 'ms'
     check(
       fnDescriptor,
       isNotDelayEffect,
-      'instead of writing `yield call(delay, ms)` where delay is an effect from `redux-saga/effects` you should write `yield delay(...)`',
+      `instead of writing \`yield call(delay, ${arg0})\` where delay is an effect from \`redux-saga/effects\` you should write \`yield delay(${arg0})\``,
     )
     validateFnDescriptor('call', fnDescriptor)
   }
--- a/packages/core/src/internal/io.js
+++ b/packages/core/src/internal/io.js
@@ -143,9 +143,7 @@ export function call(fnDescriptor, ...args) {
     check(
       fnDescriptor,
       isNotDelayEffect,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by shinima** [comment]
Failing checks is caused by babel@6 being updated. We need to update out test fixtures.

**Comment by szb512** [comment]
Babel@6 is a BIG problem during the tests, it should only use babel 7.

**Comment by szb512** [comment]
Well, you could edit the travis file.

**Comment by szb512** [comment]
I have tried that and it works.

File: packages/core/src/internal/io.js [hunk 1]
```diff
@@ -143,9 +143,7 @@ export function call(fnDescriptor, ...args) {
     check(
       fnDescriptor,
       isNotDelayEffect,
-      `instead of writing \`yield call(delay, ${String(
-        args[0],
-      )})\` where delay is an effect from \`redux-saga/effects\` you should write \`yield delay(${String(args[0])})\``,
```
shinima: `String(Object.create(null))` will throw an exception. So we could avoid stringifying args[0] here.

Andarist: its a dev path and it's strictly addressing the delay case, so maybe we could do this `typeof args[0] === 'number' ? args[0] : 'ms'`?

PR #1795, opened 2019-02-26
typings: fix `take` typings for array of heterogeneous action creators

aikoven:
Fixes https://github.com/redux-saga/redux-saga/issues/1740#issuecomment-467167529

https://github.com/redux-saga/redux-saga/pull/1769 only fixed `takeEvery` et al. But TS still got confused on similar usage of `take`.


Andarist:
CI failure seems to be unrelated, thanks!

PR #1799, opened 2019-02-28
Fixed configureStore import statement

mikepc:
npm run bootstrap
npm start
yarn start 
These commands would not start the server. 
With this PR the server works as intended.

No.
No
I don't see any tests in the real-world project

Removed a reference that was breaking server startup

I removed a reference that was preventing the server from starting and included what I believe is the ended file. I'm new to saga, so it might be wrong, but everything looks like it's working....

### DIFF
--- a/examples/real-world/store/configureStore.dev.js
+++ b/examples/real-world/store/configureStore.dev.js
@@ -1,12 +1,13 @@
 import { createStore, applyMiddleware, compose } from 'redux'
 import createLogger from 'redux-logger'
 import createSagaMiddleware, { END } from 'redux-saga'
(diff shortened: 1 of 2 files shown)


Andarist:
Thanks for pointing out there is a problem with this example - your fix was not a proper one though. 

You can check out how I've fixed this reference problem [here](https://github.com/redux-saga/redux-saga/commit/a79e2cda8530930c172d1ffb53027adb7443e879) and also had some build/setup issues which I've fixed [here](https://github.com/redux-saga/redux-saga/commit/89a11479a6ed953c522b57ef23b6e14758d64278).

PR #1801, opened 2019-02-28
Use yarn and its workspaces

Andarist:


shinima:
👍 Installing dependencies are much faster than before.

Andarist:
Glad to hear that :)

PR #1823, opened 2019-03-28
Remove unnecessary exports in Beginner Tutorial's code snippet

antododo:
| Fixed Issues?            | n/a |
| Patch: Bug Fix?          |  n/a
| Minor: New Feature?      |  no  |
| Tests Added + Pass?      | na |
| Any Dependency Changes?  |  no  |

After the refactoring, we only need to export the rootSaga and not anymore incrementAsync & watchIncrementAsync as said in the comment:
```
// notice how we now only export the rootSaga
// single entry point to start all Sagas at once
```

### DIFF
--- a/docs/introduction/BeginnerTutorial.md
+++ b/docs/introduction/BeginnerTutorial.md
@@ -176,12 +176,12 @@ function* helloSaga() {
   console.log('Hello Sagas!')
 }
 
-export function* incrementAsync() {
+function* incrementAsync() {
   yield delay(1000)
   yield put({ type: 'INCREMENT' })
 }
 
-export function* watchIncrementAsync() {
+function* watchIncrementAsync() {
   yield takeEvery('INCREMENT_ASYNC', incrementAsync)
 }
 


antododo:
The next part of the tuto is confusing, because now we want to test `incrementAsync` so we need to export it. (But we still don't need to export `watchIncrementAsync`)

So either we add it again when we want to test even if we removed it before when we refactored.

**Option 1:**
We want to test our `incrementAsync` Saga to make sure it performs the desired task.
To be able to test it, we need to export it, change back the file `sagas.js` to:
```
export function* incrementAsync() {
```

Or in the previous snippet we refactor `helloSaga` & `watchIncrementAsync` with `rootSaga`, leave  `incrementAsync` as is and change the comments: 

**Option 2:**
```
// ...
// We keep the export for incrementAsync 
// to test it in the next step
export function* incrementAsync() {
  yield delay(1000)
}

function* watchIncrementAsync() {
  yield takeEvery('INCREMENT_ASYNC', incrementAsync)
}

// notice how we now only export the rootSaga
// single entry point to start all Sagas at once
// except for the export to test incrementAsync
  yield all([
    helloSaga(),
    watchIncrementAsync()
  ])
}
```

Andarist:
Thanks for the PR, seems legit.

PR #1855, opened 2019-05-23
Update HelperWorkerParameters to match saga that don't use action

guillaumewuip:
| Fixed Issues?            | ∅ |
| Tests Added + Pass?      | No |
| Any Dependency Changes?  | No  |

This is a fix on `HelperWorkerParameters`  type helper that is used with `takeEvery`, `takeLatest`, ... effects.

Before this fix : 

```ts
function* sagaUsingAction(
  arg1: string,
  action: ReduxAction,
) {
  console.log(arg1, action);
}

// no problem here because sagaUsingAction use the action as parameter 2
yield takeEvery( 
   'TEST_ACTION_TYPE_1',
   sagaUsingAction,
   'my arg 1',
);

function* sagaNotUsingAction(
 arg1: string,
) {
  console.log(arg1);
}

// problem here
// Argument of type '"TEST_ACTION_TYPE_2"' is not  assignable to parameter of type 'TakeableChannel<{}>'
yield takeEvery(                
   'TEST_ACTION_TYPE_2', 
   sagaNotUsingAction,
   'my arg 1',
);
```

Because the `HelperWorkerParameters` assume we always use `action` in our saga, TS does find a good definition to use for `takeEvery`.

### DIFF
--- a/packages/core/__tests__/typescript/effects.ts
+++ b/packages/core/__tests__/typescript/effects.ts
@@ -754,6 +754,14 @@ function* testTakeEvery(): SagaIterator {
   yield takeEvery('my-action', helperWorker7, 'a', 'b', 'c', 'd', 'e', 'f')
   yield takeEvery('my-action', helperWorker7, 'a', 'b', 'c', 'd', 'e', 'f', 'g')
 
+  const helperWorker8 = (a: 'a', b: 'b', c: 'c', d: 'd', e: 'e', f: 'f', g: 'g') => {}
+
+  // typings:expect-error
+  yield takeEvery('my-action', helperWorker8, 1, 'b', 'c', 'd', 'e', 'f', 'g')
+  // typings:expect-error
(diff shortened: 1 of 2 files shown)


packages/core/__tests__/typescript/effects.ts
```diff
@@ -754,6 +754,14 @@ function* testTakeEvery(): SagaIterator {
   yield takeEvery('my-action', helperWorker7, 'a', 'b', 'c', 'd', 'e', 'f')
   yield takeEvery('my-action', helperWorker7, 'a', 'b', 'c', 'd', 'e', 'f', 'g')
 
+  const helperWorker8 = (a: 'a', b: 'b', c: 'c', d: 'd', e: 'e', f: 'f', g: 'g') => {}
```
guillaumewuip: I've copied `helperWorker7`, but without `action` param to test the new type behavior. Do you want more tests ?

Andarist:
Thanks!

PR #1890, opened 2019-07-12
Export CombinatorEffect, CombinatorEffectDescriptor

adrw:
| Patch: Bug Fix?          |  Downstream libraries using `redux-saga` and exporting sagas require these two types to define the type of the export. Without this type, Typescript complains that the type is not portable because it relies on the reference to this library.   |
| Tests Added + Pass?      |  |

### DIFF
--- a/packages/core/effects.d.ts
+++ b/packages/core/effects.d.ts
@@ -18,7 +18,7 @@ import {
 
 import { FlushableChannel, PuttableChannel, TakeableChannel } from './index'
 
-export { ActionPattern, Effect, Pattern, SimpleEffect, StrictEffect }
+export { ActionPattern, CombinatorEffect, CombinatorEffectDescriptor, Effect, Pattern, SimpleEffect, StrictEffect }
 
 export const effectTypes: {
   TAKE: 'TAKE'


Andarist:
Could you explain in more detail when this is needed? How exporting library tries to use it or what's the exact thing it exports?

adrw:
The Misk-Web SimpleRedux library `@misk/simpleredux` uses it like this to export helpful Redux boilerplate for forms, network, and router use cases: https://github.com/cashapp/misk-web/blob/ff4ac2d29afd5bf947c0c19963ed77c411dadb25/packages/%40misk/simpleredux/src/simpleForm.ts#L181

The above is the existing type definition, which isn't compatible with `1.0.5` and requires a type definition as follows: `IterableIterator<CombinatorEffect<"ALL", SimpleEffect<"FORK", ForkEffectDescriptor>>>`

Andarist:
Hm, seems like `IterableIterator<AllEffect<ForkEffect>>` should work.

adrw:
Thanks! That seems to work for all of our use cases except one, linked below with error following.

https://github.com/cashapp/misk-web/blob/ff4ac2d29afd5bf947c0c19963ed77c411dadb25/packages/%40misk/core/src/builders/createIndex.tsx#L33

```
The inferred type of 'createIndex' cannot be named without a reference to '../../../../../common/temp/node_modules/.registry.npmjs.org/@redux-saga/core/1.0.3/node_modules/@redux-saga/types'. This is likely not portable. A type annotation is necessary.
```

Andarist:
You have to provide a type parameter to `AllEffect`, hard to tell what's needed for you there without knowing what you are trying to achieve, but maybe the same (`IterableIterator<AllEffect<ForkEffect>>`) could work for you in that place too.

neurosnap:
It's been a few years since this PR was introduced.  I'm going to close now due to inactivity.  If we still feel like this PR is useful we can revisit.

Issue #1286, opened 2017-12-05
Documentation/Recipe Request: Using redux-saga with TypeScript


huntwj:
I see that there are type definitions, and some people have gotten them to work. As someone that's new to TypeScript, however, converting my redux-saga to TypeScript has gone less than smoothly.

It would be nice if we could add a page of documentation (preferably written by someone that has had success in doing this...) outlining the process of converting JS sagas to TS and/or writing TS sagas from scratch.

Just in case my issue turns out to be easily solvable by the experts, I'll include it here:

Specifically, I am attempting to use `takeLatest(...)` and call my API request sagas, but I am getting type errors on those that take parameters.

My TypeKeys:
```
export enum TypeKeys {
  FETCH_REQUEST = 'Fetch Request',
  .
  .
  .
}
```

In my watchers generator:
```
    yield takeLatest(TypeKeys.FETCH_REQUEST, fetchRequestSaga);
```

and in my sagas:

```
export function *fetchRequestSaga(getState?: () => any): SagaIterator {
  .
  .
  .
}
```

Then the error is:
```
TS2345: Argument of type 'TypeKeys.FETCH_REQUEST' is not assignable to parameter of type 'Channel<(() => any) | undefined>'.
```

I used to be able to call takeLatest passing in a string as the action type, but things seem to change now that I'm dealing with a TS (string-based) enum.

Andarist:
@aikoven do you know any good resources about the topic?

aikoven:
@huntwj The error you see is a result of TypeScript's way of resolving overloaded signatures. `takeLatest` has two kinds of signatures: one for taking actions and another for taking from channels. For some reason TS failed to match your arguments with the first kind. Then it tried to match with the second kind, and of course it failed too, and returned an error for the last signature it tried.

Unfortunately, I don't know of a way to see the correct error here, except for commenting out all the signatures of `takeLatest` with channels in `node_modules/redux-saga/effects.d.ts`.

You may also try the new typings for the 1.0 (https://github.com/redux-saga/redux-saga/pull/1255), where these two groups of signatures are swapped, so that users are more likely to get the correct error, since `takeLatest` is more often used with actions than channels.

erik-sn:
@huntwj did you ever find a resolution for this?

huntwj:
[...]

Issue #884, opened 2017-03-27
[Typescript] Yield return value type safety


ssynix:
I'm ramping up on redux-saga and Typescript by working on a small project.

Something I've noticed is that the return value of `yield effect(...)` is always `any`. I'm not familiar enough with the platform to understand if this is a limitation of Typescript, or some missing type annotation in redux-saga/my code.

E.g.:
```typescript
function selector(state: State) {
    return state.selected;
}

function* yieldNumber() {
    yield 1;
}

function* saga() {
    let num = yield call(yieldNumber); // num has inferred type of any
    let selected = yield select(selector); // selected: any

    // My workaround
    let num: number = yield call(yieldNumber); // works as long as I don't misrepresent the type
}
```

ssynix:
I believe this is a Typescript limitation, so I filed an issue on their end: https://github.com/Microsoft/TypeScript/issues/14883.

raduflp:
facing the same issue now that I'm trying to adopt Typescript. 
@ssynix  did you find a workaround/pattern to have type safety for the saga's return values?

ssynix:
@raduflp There's no workaround since it's lacking support at the language level. There are several threads mentioned in the issue above that discuss this feature, saying that it was blocked on boolean literal types and default template parameters (both now implemented), but generator yield type propagation still hasn't seen any traction in the past year.

The closest thing to a workaround is still what I wrote in the code sample above: you can manually assert the type yourself to make sure the rest of the code is typesafe, but there's nothing you can do to automatically detect if that original type-assertion is wrong.

nicgirault:
Is there anything new about this issue?

Andarist:
@nicgirault TS@3.6 will supposedly come with better support for generators, so this is probably something which we'll tackle then.

PejmanNik:
Hi @Andarist Is there any in-progress task due to this issue? or help how I can make PR for that?
The [microsoft/TypeScript#31639](https://github.com/microsoft/TypeScript/issues/31639) is closed and it is available with next tag.
Thank you

Andarist:
@PejmanNik this is happening in https://github.com/redux-saga/redux-saga/pull/1892

PejmanNik:
Thank you @Andarist  👍

gilbsgilbs:
[...]

PR #1892, opened 2019-07-14
Leverage TS 3.6 generator return types for Saga Iterators.

gilbsgilbs:
- Add an optional generic return type to `SagaIterator`
- Update effect types to set the proper return type
- Stronger type inference for `CombinatorEffect` (`all()` and `race()`)

See https://github.com/redux-saga/redux-saga/issues/1286#issuecomment-509032314

| Q                        | A |
| Major: Breaking Change?  | None.  |
| Minor: New Feature?      | See above  |
| Tests Added + Pass?      | Should pass, still needs some refactoring |

```typescript
// Current behavior
function* mySaga(): SagaIterator {
  yield call(someEffect);
  return "val";
}

// Possible behavior with this PR
function* mySaga(): SagaIterator<number> {
  yield call(someEffect);
  return "not a number";  // This will fail
}
```

```typescript
// Current behavior
function myEffect(): CallEffect {
  return call(function*(): SagaIterator {
    yield delay(100);
    return "ok";
  });
}

// Possible behavior with this PR
function myEffect(): CallEffect<string> {
  return call(function*(): SagaIterator<string> {
    yield delay(100);
    return "ok";
  });
}
```

```typescript
// Current behavior
function* mySaga(): SagaIterator {
  const rc = race({
    one: call(function*(): SagaIterator {
        yield delay(200);
        return 'foo';
      }),
    two: race({one: delay(1000), two: delay(500)}),
   });
  rc.payload.one; // payload may be an array => type error
  const bar = !Array.isArray(rc.payload) && rc.payload.thee; // bar incorrectly inferred to "CallEffect" or "RaceEffect"  union type
  const one = !Array.isArray(rc.payload) && rc.payload.one; // Imprecise union type. Either "CallEffect" or "RaceEffect", we don't know.
  yield rc;
}

// New behavior with this PR
function* mySaga(): SagaIterator<void> {
  const rc = race({
    one: call(function*(): SagaIterator<'foo'> {
        yield delay(200);
        return 'foo';
      }),
    two: race({one: delay(1000), two: delay(500)}),
   });

  const bar = rc.payload.thee; // type error: no effect named "three"
  const one = rc.payload.two; // CallEffect
  const two = rc.payload.two; // RaceEffect
  const rcOne = two.payload.one; // CallEffect
  yield rc;
}
```

This led me to change the generic type for race. So BC is:

```typescript
  const rc = race<CallEffect>({
    one: delay(100),
    two: delay(200),
   });

  // should be rewritten to

  const rc = race<{one: CallEffect, two: CallEffect}>({
    one: delay(100),
    two: delay(200),
   });

  // or just with type inference

  const rc = race({
    one: delay(100),
    two: delay(200),
   });
```

cc @Andarist


packages/core/effects.d.ts
```diff
@@ -1026,7 +1035,7 @@ export type GetContextEffectDescriptor = string
 /**
  * Returns an effect descriptor to block execution for `ms` milliseconds and return `val` value.
  */
-export function delay<T = true>(ms: number, val?: T): CallEffect
+export function delay<T = true>(ms: number, val?: T): CallEffect<void>
```
Andarist: correct me if I'm wrong but this should return `CallEffect<T>`

Andarist:
Looks really nice! We'll obviously have to sit on it until 3.6 gets released and figure out how to release it. We need to decide if we are fine with dropping support for pre-3.6 versions.

We'd also need to new add type-tests to avoid regressions in the future.

cc @aikoven , might be interesting to you

gilbsgilbs:
> We'll obviously have to sit on it until 3.6 gets released and figure out how to release it. We need to decide if we are fine with dropping support for pre-3.6 versions.

Actually, I don't think it is required to drop support for pre-3.6. We could used [typesVersions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#version-sélection-with-typesversions) introduced in TS 3.1 to provide the previous version for pre-3.6. I didn't succeed to set it up for some reason, but it should definitely be possible.

packages/core/effects.d.ts
```diff
@@ -477,29 +480,35 @@ export function call<Ctx, Fn extends (this: Ctx, ...args: any[]) => any>(
...
-export interface CallEffectDescriptor {
+export interface CallEffectDescriptor<RT> {
   context: any
-  fn: Function
+  fn: ((...args: any[]) => SagaIterator<RT> | Promise<RT> | RT)
   args: any[]
 }
 
+export type SagaReturnType<S extends Function> =
+  S extends (...args: any[]) => SagaIterator<infer RT> ? RT :
```
danielnixon: I think this needs to be something like 
```
ReturnType<S> extends Iterator<any, infer RT, any> ? RT :
```
because ` export type SagaIterator = IterableIterator<StrictEffect>` isn't generic.

packages/types/index.d.ts
```diff
@@ -1,12 +1,12 @@
...
 
-export type Saga<Args extends any[] = any[]> = (...args: Args) => IterableIterator<any>
+export type Saga<Args extends any[] = any[]> = (...args: Args) => Iterator<any>
 
 /**
  * Annotate return type of generators with `SagaIterator` to get strict
  * type-checking of yielded effects.
  */
-export type SagaIterator = IterableIterator<StrictEffect>
+export type SagaIterator<RT = any> = Iterator<StrictEffect, RT>
```
danielnixon: I found I needed this to be `Iterator<StrictEffect, RT, any>`.

danielnixon: Try the following with and without that change:
```
type RT<A extends Effect<any, any>> = 
  A extends Effect<any, CallEffectDescriptor<infer RT>> ? RT :
  A extends Effect<any, CombinatorEffectDescriptor<() => SagaIterator<infer RT>>> ? RT :
  unknown;

function* aSaga(): SagaIterator<number> {
  const effect = call(() => Promise.resolve(1));
  const effectResult: RT<typeof effect> = yield effect;
  
  return effectResult;
}
```

gilbsgilbs: Wow, what neat idea to strongly type effects return type you gave here 🧐 . Thanks.

Yeah, TNext is of type unknown by default. Any is probably our best option.

danielnixon: Thanks. I also found I had to go with `Iterator<Effect, RT, any>` (not `Strict`) to get this to work:

```
export function* raceArrayExample(): SagaIterator<number> {
  const effect = race([shouldInferNumber, shouldInferNumber]);
  const effectResult: RT<typeof effect> = yield effect;

  return Math.max(44, effectResult);
}

export function* raceObjectExample(): SagaIterator<number> {
  const effect = race({one: shouldInferNumber, two: shouldInferNumber});
  const effectResult: RT<typeof effect> = yield effect;

  return Math.max(45, effectResult);
}
```

packages/types/types/ts3.6/index.d.ts
```diff
@@ -0,0 +1,173 @@
+import { Action } from 'redux'
+
+export type Saga<Args extends any[] = any[]> = (...args: Args) => IterableIterator<any>
+
+/**
+ * Annotate return type of generators with `SagaIterator` to get strict
+ * type-checking of yielded effects.
+ */
+export type SagaIterator = IterableIterator<StrictEffect>
```
danielnixon: Did we lose `export type SagaIterator<RT = any> = Iterator<StrictEffect, RT, any>` here?

gilbsgilbs: Yeah, sorry, I messed up my git checkouts apparently. I work on this at night, after work, which is very prone to mistakes 😅. I wanted to keep backwards compatibility with TS<3.6. I'll restore TS3.6 typings later, probably tommorow. The commit should still exist in the tree.

I'll probably also try to drop typings-tester in favor of dtslint to have proper tests for each types versions sometime.

gilbsgilbs:
Hi,

I migrated tests to `dtslint` instead of `typings-tester` in order to easily run the tests against different typescript versions (see [dtslint](https://github.com/gilbsgilbs/redux-saga/tree/dtslint) branch in my fork). I might be completely dumb, but I can't get [`typesVersions`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#version-selection-with-typesversions) to work. If I run `yarn run dtslint types` in `packages/core` on my `dtslint` branch, it fails for TS next because it insists on using `@redux-saga/types` for TS<3.6 :thinking: . No idea why.

Notice that I had to temporarily tweak dtslint into thinking that `3.5` version is actually `next` version, otherwise, dtslint would simply refuse to run the tests. (`mv ~/.dts/typescript-installs/3.5 ~/.dts/typescript-installs/3.5.bak && cp -R ~/.dts/typescript-installs/next ~/.dts/typescript-installs/3.5`)

I'm kind of stuck and it's starting to drive me crazy. I would highly appreciate if somebody could bring me some help on this! (ping @danielnixon?)

Thanks.

anilanar:
@gilbsgilbs `typesVersions` is only checked during module resolution. For that to work

1) `moduleResolution` must be set to `node`
2) Imported module must be a directory with a `package.json` with `types` field.

When you run `dtslint types`, either:
1) `dtslint` tries to import `./types`, `./types` doesn't have a `package.json` so it tries to import `./types/index.d.ts` instead. The solution to this is to put a `package.json` with `typesVersions` inside `./types`.
2) `dtslint` doesn't try to import `./types` and instead it has its own module resolution for the `path` argument. In that case you'd have to mimic github.com/DefinitelyTyped structure. e.g. `./types/redux-saga/{put all files here incl. a package.json}`

henrikra:
Is something still blocking this PR to land? 🤔

Andarist:
That's a question for @gilbsgilbs - if there is anything left to do as part of this PR. Once I get a signal that in his eyes it's ready to merge then I'll ping our TS guy™ to look at this.

gilbsgilbs:
Hey! I'm just back from vacation actually, sorry for the delay :sweat_smile: :palm_tree: :sunglasses: . Appart from what I wrote in my previous comment, I reckon everything is fine. I'll try to see if @anilanar comment can help me out. Thank you very much by the way :heart: .

gilbsgilbs:
I was able to solve the `typesVersions` issue thanks to the `--traceResolution` flag in `tsc`. Good stuff. I then fixed some failing tests and it worked :tada: . This means that:
- we should be able to release this without breaking compatibility for TS<3.6
- there shouldn't be any breaking change for 3.6<=TS
- typings tests were ported to dtslint (instead of [typings-tester](https://github.com/aikoven/typings-tester/)) which is probably the recommended way to test typings now

Notices:
- Sorry for the copy-pasta (especially the tests), I don't think dtslint gives a better option yet :confused: . It's probably the price to pay to avoid breaking changes.
- It may need some advanced tests for TS3.6 (especially some `$ExpectType`), let me know.

Andarist:
@aikoven could u take a look at this?

aikoven:
LGTM 👍

Andarist:
@gilbsgilbs thanks for working on this! I'll try to remember to release this 2morrow

henrikra:
So now when I use for example `const someValue = yield call(someFunction)` then the `someValue` will have right type? :D

gilbsgilbs:
@Andarist @aikoven thanks for the review and merge. May I suggest that we release it as pre-release first? Or you don't think it's necessary?

@henrikra unfortunately not. TS 3.6 doesn't allow this. However, it will open the door to very nice workarounds such as [this](https://github.com/redux-saga/redux-saga/issues/884#issuecomment-513200739) or linting rules. Stay tuned.

henrikra:
Great! Can you guys add some documentation about this to docs? Is there section for redux saga with typescript?

gilbsgilbs:
See #1286. It would also be nice to have an official wrapper library for `yield*` effects.

Andarist:
>  May I suggest that we release it as pre-release first? Or you don't think it's necessary?

I can do that, but not sure how many people would actually try this out before this would hit stable release - so not sure if it's worth it.

Andarist:
Ok, I've managed to publish v1.1.0. It seems that `dtslint` installs TS in `require('os').homedir()` so probably I had some broken version there as next and doing fresh installs of redux-saga repo didn't help (as used TS instance was outside of that repo).

danielnixon:
After upgrading to v.1.1.0 this no longer compiles:
```TypeScript
import { call } from "redux-saga/effects";
```

> Module '"....../node_modules/redux-saga/effects"' has no exported member 'call'.

PR #1895, opened 2019-07-19
Fix take(predicate) TypeScript error

### LINKED ISSUE #1871: TypeScript error when using predicate as the first argument of takeEvery effect
I have such code in my app (Create React App with TypeScript):
```typescript
  yield all([
    handleAPIRequestSaga, // common saga for the whole app.
  ]);
}
```
```typescript
interface HandleAPIRequest {
  prefix: string;
  status: string;
  request: () => Promise<any>;
}

const handleAPIRequestSaga = takeEvery(
  ({ status, request }: HandleAPIRequest) => (
    // requestStatuses.start is 'START'
    isEqual(status, requestStatuses.start) && request
  ),
  handleAPIRequest,
);
```
And TypeScript says the following:
image

Is it correct behavior or not? If it is correct, what approaches to solve the error there are?

`redux-saga`: 1.0.3
`typescript`: 3.5.2

Thanks in advance.

marcandrews:
| Fixed Issues?            | Fixes #1871  |
| Patch: Bug Fix?          | 👍 |
| Tests Added + Pass?      | 👍 |
| Any Dependency Changes?  |    

### Discussion

- [ ] The `take(isMyAction)` test was updated to use an actions union in hopes to match a use-case more similar to that in the real-world.
- [x] The `ActionSubPattern` type definition was updated/corrected to allow the test to pass.

I must admit, I am pretty new to TypeScript, so any comments, suggestions and insight into the greater impact of these changes will be greatly appreciated.

cc: @Andarist, @vasilii-kovalev, @osoftware, @aikoven

### DIFF
--- a/packages/core/__tests__/typescript/effects.ts
+++ b/packages/core/__tests__/typescript/effects.ts
@@ -42,8 +42,8 @@ Object.assign(stringableActionCreator, {
   },
 })
 
-const isMyAction = (action: Action): action is MyAction => {
-  return action.type === 'my-action'
+const isMyAction = (action: MyAction) => {
+  return action.type === 'my-action' && action.customField === 'custom-field'
 }
 
 type ChannelItem = { someField: string }
--- a/packages/types/index.d.ts
+++ b/packages/types/index.d.ts
@@ -26,7 +26,7 @@ export type Pattern<T> = SubPattern<T> | SubPattern<T>[]
 export type ActionSubPattern<Guard extends Action = Action> =
   | GuardPredicate<Guard, Action>
   | StringableActionCreator<Guard>
-  | Predicate<Action>
+  | Predicate<Guard>
   | ActionType
 
 export type ActionPattern<Guard extends Action = Action> = ActionSubPattern<Guard> | ActionSubPattern<Guard>[]


marcandrews:
Given the [recent discussions](https://github.com/redux-saga/redux-saga/issues/1871#issuecomment-520147724) in #1871, please let me know if you would like me to update the tests to utilize a union of action types.

Andarist:
@gilbsgilbs could you maybe take a look at this?

marcandrews:
Actually, now that TypeScript generator support is improved, and #1892 is merged, I am wondering if we should allow the action itself to be returned from the predicate so that the action returned from `take()` and passed to the saga in `take*(pattern, saga)` can be determined without the use type guards.

What do you think @Andarist @gilbsgilbs?

Andarist:
Could you elaborate? Preferably with a simple code that would use the proposed pattern.

marcandrews:
I took a quick look at the TypeScript tests in v1.1.1 and I could not reproduce #1871. With respect to my https://github.com/redux-saga/redux-saga/pull/1895#issuecomment-535592284, it looks like [take* is already working as suggested](https://github.com/redux-saga/redux-saga/blob/master/packages/core/types/ts3.6/effects.test.ts#L926).

🥇@gilbsgilbs

gilbsgilbs:
@marcandrews doesn't the line your pointed out rely on an implicit `any` type? This test did already exist before I moved the files around.

PR #1909, opened 2019-08-14
do not attach cancel functions on noop

gabormagyari:

Hi! I'm testing the library for memory retention. I've found that if the continuation function is `noop`, it attaches the task's `cancel` function to it. Because `noop` is kind of a library global method, it will never be GC'd, leaving the cancel in memory with the iterator and the action in its context. This could lead to having unnecessary sensitive data in memory.

### DIFF
--- a/packages/core/src/internal/proc.js
+++ b/packages/core/src/internal/proc.js
@@ -50,7 +50,9 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
     attaches cancellation logic to this task's continuation
     this will permit cancellation to propagate down the call chain
   **/
-  cont.cancel = task.cancel
+  if (cont) {
+    cont.cancel = task.cancel
+  }
 
   // kicks up the generator
   next()
--- a/packages/core/src/internal/effectRunnerMap.js
+++ b/packages/core/src/internal/effectRunnerMap.js
@@ -155,7 +155,7 @@ function runForkEffect(env, { context, fn, args, detached }, cb, { task: parent 
   const meta = getIteratorMetaInfo(taskIterator, fn)
 
   immediately(() => {
-    const child = proc(env, taskIterator, parent.context, currentEffectId, meta, detached, noop)
+    const child = proc(env, taskIterator, parent.context, currentEffectId, meta, detached, undefined)
 
     if (detached) {
       cb(child)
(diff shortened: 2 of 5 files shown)


Andarist:
Hi! this is a really interesting topic - could you elaborate more on when this becomes a memory leak? I would love to get a better understanding of the situation here.

gabormagyari:
Sure! It is not exactly a memory leak.

Let's assume I'm working on a login form and my login action's payload is my username and password. After running the login saga, at one point a cancel method is attached to `noop` and remains there. 

Taking a heap snapshot after login (see below), you can see there is indeed a `cancel` method on `noop` which is keeping the password in memory. (noop -> cancel -> mainTask -> cancel -> iterator$1 -> parameters[0] -> payload -> password).

Although the saga is finished, this reference keeping my action data in memory until another saga attaches another cancel method. Also, an unwanted side effect is that I can call `noop.cancel()` from anywhere which will call the task's cancel.

Screenshot 2019-08-14 at 14 09 00

Andarist:
Ok, this got me really anxious that there might be some subtle bug in the code - but after investigating it for some time it seems that this `cont` (which is `noop`) never escapes anywhere and thus this particular `cont.cancel` never gets called.

I would like to introduce an extra guard that this won't regress in the future - could you add smth like this to utils?
```js
let noop = () => { }

if (process.env.NODE_ENV !== 'production' && Proxy) {
  noop = new Proxy(noop, {
    set: () => {
      throw internalErr(
        'There was an attempt to assign a property to internal `noop` function.',
      )
    },
  })
}

export { noop }
```

packages/core/src/internal/proc.js
```diff
@@ -50,7 +50,9 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
     attaches cancellation logic to this task's continuation
     this will permit cancellation to propagate down the call chain
   **/
-  cont.cancel = task.cancel
+  if (cont !== noop) {
```
Andarist: I would personally prefer:
- removing passing `noop` as an argument to `proc` altogether (making it undefined in cases where `noop` is being passed)
- changing this to just `if (!cont)`
- adding `noop` as default parameter of `cont` in `newTask` [here](https://github.com/redux-saga/redux-saga/blob/374fde6e70945e0d20b2e4c221b74f643e26bd2b/packages/core/src/internal/newTask.js#L9)

gabormagyari:
Thanks for the suggestions. I added the desired changes. I think `noop` is good for now.

Although, there is a very similar issue with takers and `takeCb`. I think this one is not so easy to fix. Do you want me to open an issue?

Andarist:
> Although, there is a very similar issue with takers and takeCb. I think this one is not so easy to fix. Do you want me to open an issue?
Yes, please do so.

PR #1912, opened 2019-08-20
Fixed object type checker function definition

mjangir:
There is an older bug in ECMA script that it treats null as an object which we should generally exclude if we really want a particular data to be real javascript object. In this PR, I've just put one more check of not null in object type checker function.

### DIFF
--- a/packages/is/src/index.js
+++ b/packages/is/src/index.js
@@ -6,7 +6,7 @@ export const func = f => typeof f === 'function'
 export const number = n => typeof n === 'number'
 export const string = s => typeof s === 'string'
 export const array = Array.isArray
-export const object = obj => obj && !array(obj) && typeof obj === 'object'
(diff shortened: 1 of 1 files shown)


packages/is/src/index.js
```diff
@@ -6,7 +6,7 @@ export const func = f => typeof f === 'function'
 export const number = n => typeof n === 'number'
 export const string = s => typeof s === 'string'
 export const array = Array.isArray
-export const object = obj => obj && !array(obj) && typeof obj === 'object'
+export const object = obj => obj && !array(obj) && typeof obj === 'object' && obj !== null
```
Andarist: that's already handled by `obj &&` part, it wont proceed further when `obj === null`

mjangir: Yeah I got it. Thanks. I didn't look at the first part. My bad.

mjangir:
Its already there. So no need to merge this PR. Closing it.

PR #1914, opened 2019-08-21
actionChannel uses a fixed-size buffer (not unlimited)

tlaitinen:
The documentation says that `actionChannel` buffers all incoming messages without limit. However, it seems that the default buffer has capacity for 10 elements. See https://gist.github.com/tlaitinen/5b46cb63b89883ccce805c5e2dc6cb59 for steps to reproduce a buffer overflow.


Andarist:
Are you using v0.x? This was changed in v1 and currently most channels are using expanding buffer by default - https://github.com/redux-saga/redux-saga/blob/be0d8780b7658f939a0a4a8ad612edbcbdf728fa/packages/core/src/internal/channel.js#L18

tlaitinen:
Oops sorry I thought I had the latest version, but it's 0.16.0 so the documentation is up-to-date.

PR #1920, opened 2019-08-28
doc: takeLatest: "spawns" => "forks"

slorber:
cc @Andarist . Spawn is written on other places as well, not very confortable to modify all of them without getting your approval first


Andarist:
Feel free to edit other places too :)

PR #1922, opened 2019-08-30
Reword api docs for PUT

### LINKED ISSUE #1921: Update API documentation for put
Not sure if anyone else had the same experience, but I was thrown off a little bit by 
https://redux-saga.js.org/docs/api/#putaction

> Creates an Effect description that instructs the middleware to dispatch an action to the Store. This effect is non-blocking and any errors that are thrown downstream (e.g. in a reducer) will not bubble back into the saga. 

The term `non-blocking` somehow led me to expect `put` to behave like some kind of "asynchronous `dispatch`".  This intuition is of course, incorrect, as we can expect the store to immediately receive the action from `put()` unless we have some kind of asynchronous middleware that delays/prevents the action from reaching the store.

I do wonder if this is a valid concern? If it seems like a reasonable mistake to make, I'd be glad to submit a PR to suggest tweaks to the wording.

superburrito:
| Fixed Issues?            | Fixes #1921 |
| Patch: Bug Fix?          |   No |
| Major: Breaking Change?  | No   |

This PR edits the description for `put` effects.

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -349,7 +349,11 @@ You can also pass in a channel as argument and the behavior is the same as [take
 
 ### `put(action)`
 
-Creates an Effect description that instructs the middleware to dispatch an action to the Store. This is similar to calling `dispatch` directly on the action; your Store will be synchronously updated unless you have other middlewares with asynchronous flows that delay the propagation of the action. Downstream errors will be bubbled up.
+Creates an Effect description that instructs the middleware to schedule the dispatching of an action to the store. This dispatch may not be immediate since other tasks might lie ahead in the saga task queue or still be in progress.
+
+You can, however, expect the store to be updated in the current stack frame (i.e. by the next line of code after `yield put(action)`) unless you have other Redux middlewares with asynchronous flows that delay the propagation of the action.
+
+Downstream errors (e.g. from the reducer) will be bubbled up.
 
 - `action: Object` - [see Redux `dispatch` documentation for complete info](http://redux.js.org/docs/api/Store.html#dispatch)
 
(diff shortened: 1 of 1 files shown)


docs/api/README.md
```diff
@@ -349,9 +349,7 @@ You can also pass in a channel as argument and the behavior is the same as [take
 
 ### `put(action)`
 
-Creates an Effect description that instructs the middleware to dispatch an action to the Store.
-This effect is non-blocking and any errors that are thrown downstream (e.g. in a reducer) will
-not bubble back into the saga.
+Creates an Effect description that instructs the middleware to dispatch an action to the Store. This is similar to calling `dispatch` directly on the action; your Store will be synchronously updated unless you have other middlewares with asynchronous flows that delay the propagation of the action. Downstream errors will be bubbled up.
```
Andarist: after consideration - `put` is not entirely synchronous, it gets scheduled in internal queue and is guaranteed to be dispatched within current stack frame, but this might not happen immediately. This is an important part of the behavior and has to be mentioned here.

superburrito: To be candid I'm not quite sure what "immediate" means here, would you mind clarifying?

If the dispatch is done in the current stack frame and the store is updated before the next line of code in the saga function is executed, isn't that considered synchronous? 🤔

Andarist: Well, "nested" puts are kinda flattened/queued. Consider this:
```js
yield put(FOO)

// other saga
yield takeEvery(FOO, function*() {
  // this gets called synchronously in response to `put(FOO)`
  // this BAR dispatch won't happen immediately though
  // it will get called after we exit FOO processing
  yield put(BAR)
})
```

superburrito:
Was motivated by your comments above to peek at the source code for the scheduler and effect runners. Not sure I've got the right mental model, but hopefully the updated description for `put` is now more complete / accurate

Andarist:
Thanks!

PR #1936, opened 2019-09-19
Update README.md

sayjeyhi:
Add some missing descriptions in `API` documentaion

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -599,7 +599,6 @@ Creates an Effect description that instructs the middleware to cancel a task in 
 It allows to reuse destructor-like logic inside a `finally` blocks for both outer (`cancel(task)`) and self (`cancel()`) cancellations.
 
 #### Example
-
 ```javascript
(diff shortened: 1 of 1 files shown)


docs/api/README.md
```diff
@@ -1080,7 +1081,7 @@ A channel is an object used to send and receive messages between tasks. Messages
 
 Every channel has an underlying buffer which defines the buffering strategy (fixed size, dropping, sliding)
 
-The Channel interface defines 3 methods: `take`, `put` and `close`
+The Channel interface defines 4 methods: `take`, `put`, `flush` and `close`
```
Andarist: Actually flush is not available on multicastChannel and put is not available on eventChannel.

jafar-rezaei: @Andarist well, I think this description could be added below this section, on those methods rules

neurosnap:
It's been a few years since this PR was introduced.  I'm going to close now due to inactivity.  If we still feel like this PR is useful we can revisit.

Issue #1941, opened 2019-09-23
Errors on invalid effects in fork


vansosnin:
### Description

Hello! First of all, thank you for this awesome library.

In your [v1 release](https://github.com/redux-saga/redux-saga/releases/tag/v1.0.0) you have stopper interpreting effects returned from `fork`.

![image](https://user-images.githubusercontent.com/8207551/65412258-504f7500-de08-11e9-8918-01b8b37fc706.png)

Though there is written that "_this shouldn't affect any real-life code_", it affected my project :) But it doesn't matter much.

**The question is**: is it possible to throw errors in such situations? This code had just stopped working silently and we couldn't see the problem in errors monitor.

vansosnin:
Why only dev-mode error? If you have error monitoring only in production, and forgot to test this in dev mode, it still can blow up silently.

Andarist:
> Why only dev-mode error? If you have error monitoring only in production, and forgot to test this in dev mode, it still can blow up silently.

Well - if you forget to test in dev mode you are probably in bigger problems than this one 😉I believe that adding a production error for this could be a breaking change and we can't do that without bumping a major version.

vansosnin:
Ofcourse there are some problems) And still it doesn't mean that throwing error is incorrect. It is one of ways to receive feedback and detect problem.
Well, breaking change is an argument. Maybe, you can add this feature to some kind of backlog for version 2?

Andarist:
> Well, breaking change is an argument. Maybe, you can add this feature to some kind of backlog for version 2?

Could be done, but there are no plans right now to work on v2 - the API is stable and I have no ideas what I would refactor that would need breaking changes (only an idea or two for new, but compatible, APIs which can be released as part of the v1 line).

PR #1947, opened 2019-10-01
Add dev mode error for fork

vansosnin:
| Fixed Issues?            | #1941  |
| Patch: Bug Fix?          |  Probably :)  |

Added error logging in dev mode if argument passed to `fork` is an effect.

### DIFF
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -1,6 +1,6 @@
 // TypeScript Version: 3.2
 import { Action, Middleware } from 'redux'
-import { Saga, Buffer, Channel, END, Predicate, SagaIterator, Task } from '@redux-saga/types'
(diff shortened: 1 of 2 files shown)


vansosnin:
Hmm, status lagged, but build completed

image

packages/core/types/index.d.ts
```diff
@@ -241,8 +241,8 @@ export interface RunSagaOptions<A, S> {
 }
 
 export const CANCEL: string
-export const END: END
-export type END = END
+export const END: EndType
```
Andarist: was this change needed? 🤔

vansosnin: I'm not sure. But this is the only way I could fix [strange build error](https://travis-ci.org/redux-saga/redux-saga/jobs/592087832#L720).

Andarist:
Thanks!

PR #1954, opened 2019-10-14
Use changesets

Andarist:
@Noviny would you be so kind to look at this PR? I'll leave a few questions in the comments


.changeset/config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@0.2.1/schema.json",
+  "changelog": "@changesets/cli/changelog",
+  "commit": true,
+  "linked": [["@redux-saga/core", "redux-saga"]],
+  "access": "public"
```
Andarist: can `access` be configured differently for particular packages? I have other repo which mixes private & public packages and I'm wondering if such use case is already covered or if the support for it could be added (I could possibly work on adding it). 

Also - could the default just query `package.json` for a particular package? there is `publishConfig` field in `package.json` that could do a trick. The bot could also validate if that field is added correctly.

.changeset/config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@0.2.1/schema.json",
+  "changelog": "@changesets/cli/changelog",
+  "commit": true,
```
Andarist: Looking through docs I still am not sure what exact result does it have - for both cases. Could you describe in a little bit more detail how this works?

.changeset/README.md
```diff
@@ -0,0 +1,8 @@
+# Changesets
```
Andarist: does changesets read package.json workspaces automatically to determine what a package is and what is not? I'm pretty sure it does, would only like to reassure myself

package.json
```diff
@@ -23,7 +23,11 @@
     "docs:build": "npm run docs:prepare && gitbook build -g redux-saga/redux-saga",
     "docs:watch": "npm run docs:prepare && gitbook serve",
     "docs:publish": "npm run docs:clean && npm run docs:build && cp CNAME _book && touch _book/.nojekyll && gh-pages -d _book -r git@github.com:redux-saga/redux-saga.git -m 'update book' -t",
-    "bundlesize": "lerna run --parallel bundlesize"
+    "bundlesize": "lerna run --parallel bundlesize",
+    "changeset": "changeset",
```
Andarist: is there a way to add changesets (and automate release process) for beta releases etc?

.changeset/config.json
Noviny: Quick q: Do you mean packages that are `private: true` (not published at all) or ones that are `access: restricted` (privately published package). The first we support (we respect private: true), there's a bit more on the second below.

---

Technically, you can use publishConfig to do this, but there's behaviour with scoped packages where it assumes `--access=private`, which if your repo-level config is `public` you generally don't want.

I made an issue around ensuring you can have booth public and restricted packages in a repository:

https://github.com/atlassian/changesets/issues/193

.changeset/README.md
Noviny: Correct, It uses your `workspaces` field to determine what is and is not a package.

If you run `yarn changeset` it will also show you a list of all the things it has detected as packages if you want to verify what it's affecting. I'll make a note to clarify this in the docs.

package.json
Noviny: We're working on prereleases at the moment:

https://github.com/atlassian/changesets/pull/183

.changeset/config.json
Noviny: Making a note to clarify in the docs, but short summary here:

If `commit` is true:

- when you run the `changeset` command, it will, after adding the changeset file, stage and commit for you.
- when you run the version command, after updating the changelogs and versions, it will stage and commit for you.

If `commit` is false, changesets will never stage or commit files.

Noviny:
👋 Thanks for trying it out! Hope my answers were helpful. Will keep an eye on this PR in case you have any other questions or concerns, and will update the docs for places where things weren't clear.

.changeset/config.json
Andarist: > Quick q: Do you mean packages that are private: true (not published at all) or ones that are access: restricted (privately published package). The first we support (we respect private: true), there's a bit more on the second below.

I meant here the latter (privately published) - but packages not published at all are also of my interest, so this is good to know.

> Technically, you can use publishConfig to do this, but there's behaviour with scoped packages where it assumes --access=private, which if your repo-level config is public you generally don't want.

Per package `publishConfig` could alleviate the issue altogether, right? As the decision would be local to a package and explicit.

.changeset/config.json
Andarist: It would be good to have the ability to configure this separately - it seems like I wouldn't like creating changesets to create commits, but I would see myself wanting auto-commits after versioning. 

Also - just a sanity-check - do you verify if there are any uncommitted changes before running version & publish? Seems like it would be good to fail in such situation.

package.json
Andarist: ❤️ great!

package.json
```diff
@@ -23,7 +23,11 @@
...
+    "version-packages": "changeset version",
```
Andarist: with a changeset bot is there any need for this script? versioning is done automatically by a bot, right?

Noviny: With the bot, there's no need for it.

.changeset/config.json
Noviny: > It would be good to have the ability to configure this separately

We had that in v1, and almost removed the `commit` option entirely, but ended up deciding to move to one option.

If you are opting into auto-versioning through the GH action, you don't run the versioning command yourself. Will add this to our documentation around recommendations.

> do you verify if there are any uncommitted changes before running version & publish?

We verify there are no uncommitted changes before making any commit, but not explicitly before version and publish. Changesets theoretically works outside of git, and we've been trying to reduce the print of where we touch git.

I'd probably recommend making that part of any publish script you run (you generally want something like `yarn publish` instead of running `yarn changeset publish` just so you can make a script of `yarn build && changeset publish`)

.changeset/config.json
Noviny: > Per package publishConfig could alleviate the issue altogether

I think we should respect per-package publish, but having a global config for when it's unspecified, mostly because of scoped packages. Most people using scoped packages publicly just do an initial publish with `--access="public"` and don't specify it in the config. When a new package is added later on, this often gets missed, causing initial release pain (Michael Jackson was complaining about this last week).

Letting the repository owner get it 'right' once is pretty powerful because it lowers the needed knowledge to someone contributing a new package.

We should definitely respect per-package config though 😅

.changeset/config.json
Noviny: Err, only, not `yarn publish` but `yarn release` because naming a script publish is a massive footgun I have done before.

Andarist: > We verify there are no uncommitted changes before making any commit, but not explicitly before version and publish. Changesets theoretically works outside of git, and we've been trying to reduce the print of where we touch git.

Well - it seems to me changesets are pretty much coupled to git querying anyway, so don't get that argument here, but I also don't care that much as I intend to automate things with changesets and in such scenario I can't land with uncommitted stuff before those actions (unless my scripts are broken, or I don't .gitignore stuff properly).

As to `changeset publish` - does it run `changeset version` implicitly as well? From what I see the bot doesn't have to be configured to run version command separately.

Andarist: > If you are opting into auto-versioning through the GH action, you don't run the versioning command yourself. 

That's right - but even auto-versioning needs to decide if it should do extra commits or not, right? I would expect this to affect the automated process as well.

.changeset/config.json
Andarist: Gotcha 👌 would it be possible to add validation to changeset bot so it verifies if there is either per package or global `publishConfig` missing?

Andarist:
@Noviny sorry for all the questions - but this thing, in particular, is hard to test in isolation, so I would like to have the process verified with you before actually trying it on production 😉

.changeset/config.json
Noviny: Definitely. To the bot, or to the action? Or to the cli?

.changeset/config.json
Noviny: If you're doing auto-versioning we have to commit the changes no matter what your config, otherwise it couldn't raise the PR.

.changeset/config.json
Noviny: (I would probably vote for the GH action that manages publishing being the place to add it, but it's possible it should just be part of the core library)

Noviny:
@Andarist No worries. Getting versioning right is important and it definitely causes headaches when you don't.

I'd definitely recommend if you haven't, trying adding some changesets and then running the `version` command so you can observe the output. Lets you see all of the bits working save the actual publishing.

.changeset/config.json
Andarist: > Definitely. To the bot, or to the action? Or to the cli?

I don't quite understand the difference between the bot and the action here - like bot "just" wraps the action, right? 

Without thinking about it too much I would vote for the CLI because it's always valuable - no matter if using the bot or not. Bot would only have to handle this CLI error to report it back nicely on GitHub.

.changeset/config.json
Andarist: To raise an issue - yeah, but I'm still a little bit unsure what happens in regards to git when the actual release is being done. We merge a PR so obviously the change lands in git, this triggers the action - it runs versioning command (which probably creates git tags etc) and also runs publish command, but I'm not sure if it this creates a separate commit from the one that has landed into master? And not sure how commit setting applies to that process when taking auto-versioning into account.

mitchellhamilton: An important clarification on what the action does is that it will only run _one of_ version or publish: If there are changesets on master, it will run version, commit to a branch and open a PR. If there aren't changesets on master, it will run publish which will publish to npm and create the git tags but it doesn't modify anything on the file system/there isn't anything that needs to be committed.

Noviny:
@Andarist just thought I'd check in and see if you still had any unanswered questions / concerns 😃

.changeset/config.json
Andarist: Ok, this explains it a bit more for me 👍 I was missing the part that versioning actually already happens on the created branch.

One additional thing I got from this and reading the action's source code - the `commit` option is relevant only when using `changeset` tool by hand. It doesn't affect bot and the action anyhow. Might be good to clarify this in the docs @Noviny

Andarist:
I think I understand the whole flow now, gonna merge this in soon. Thanks for your help @Noviny and @mitchellhamilton ! Can't also wait for prereleases support 😉

PR #1957, opened 2019-10-18
Add Square Logo

### LINKED ISSUE #1705: Icons!
Hello!

This is a general question regarding the logo icons. The vs-code material icon library discussed (and added!) redux icons into their set, but were resistant to the redux-saga icon. I think it's due to the rectangular shape, which makes it difficult.

https://github.com/PKief/vscode-material-icon-theme/issues/255#issuecomment-391993647 

Are there any plans for a square icon option? I'm no graphic artist, but it would be really cool to get the icon set to also advertise redux-saga!

Acesmndr:
| Fixed Issues?            | Fixes #1705  |
| Minor: New Feature?      | Yes |

`Redux-saga` doesn't have a `square logo` so it's quite hard to use the `rectangular` aspect ratio logo in some places where `square` aspect ratio logos are required (as it would be quite small if we used it as a square).
Wanted to redesign the logo but it would need the redesign in other logos as well so this is the best way I could come up with without redesigning the logo.

## Newly Added Square Logo:
Redux-Saga-Logo-Square

## Another alternative:
Redux-Saga-Logo-Square1

### Alternative 2
test-saga


Andarist:
Thanks for doing this! I think I like the alternative version better - it doesn't rotate the main "atom" and thus looks more natural to me. Could you update the PR to use the other one?

Acesmndr:
@Andarist 🙇Sure thing! Done! :)
Just a note on the alternative design is that there is a bit more empty spacing on the right and left edges compared to the other design.

Acesmndr:
@Andarist Are we waiting on something here? I see that the checks aren't running since only the images are changed. Do I need to make any more changes?

Andarist:
I just lost track of this PR - thanks for reminding me about it and for preparing it. Great work! ❤️

Issue #1932, opened 2019-09-17
Examples/descriptions for typescript 3.6.3


simplone:
### Description
I have used only few weeks redux saga and tried to update packages for latest versions. But stuck with upgrade of typescript version. In latest added update for typechecks in generators. 
https://devblogs.microsoft.com/typescript/announcing-typescript-3-6/#stricter-generators
Please provide, if it possible, more info how to use current d.ts files and match types. I hope anybody from community had the same troubles. 

Thanks

Andarist:
Some examples can be found here - https://github.com/redux-saga/redux-saga/pull/1892#issue-297396449

kristian-puccio:
Is there an example of how to get the select effect to propagate state?
Without any extra type annotations I'm getting an any type.

To work around this I'm currently using ReturnType<typeof mySelectorFunc> to annotate the resulting value.

Thanks!

Andarist:
@kristian-puccio this is not yet possible with TS - it can't express a relation between yielded value and the injected one.

kristian-puccio:
Thanks @Andarist for the clarification. 
I'll go look up the TS issues to see if it's on the roadmap

henrikra:
@Andarist But looks like there is this minimal wrapper which makes it possible? https://github.com/agiledigital/typed-redux-saga

Andarist:
Yes, this is worth exploring. I would like to get input on this idea from more people first though. cc @aikoven @gilbsgilbs

kristian-puccio:
@henrikra Awesome! I'm currently deleting all my type definitions from inside my sagas. Soo good!

I have no idea what the `yield*` is about though and I don't like how if you forget the * it defaults to any, I would prefer unknown.

gilbsgilbs:
@kristian-puccio the [`yield*` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield*) is similar to `yield`, except it yields all the values from an iterable instead of one value.

e.g.

```js
function* gen0() {
  yield 1;
  yield 2;
}

// is equivalent to

function* gen1() {
  yield* [1, 2];
}
```

[...]

PR #1962, opened 2019-10-27
Make redux-saga fully TypeScript compatible and type safe

henrikra:
Redux-saga typed generator return values

### DIFF
--- a/packages/core/types/effects.d.ts
+++ b/packages/core/types/effects.d.ts
@@ -906,9 +906,11 @@ export function select(): SelectEffect
 export function select<Fn extends (state: any, ...args: any[]) => any>(
   selector: Fn,
   ...args: Tail<Parameters<Fn>>
-): SelectEffect
+): SagaGenerator<ReturnType<Fn>>
 
 export type SelectEffect = SimpleEffect<'SELECT', SelectEffectDescriptor>
+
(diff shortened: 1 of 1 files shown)


Andarist:
This isn't true though - similarly to why `call` is not type-safe currently (it just returns CallEffect). There is an ongoing discussion on how to make all effects type-safe [here](https://github.com/redux-saga/redux-saga/issues/1932). I'd like to proceed with what we've figured out there soon-ish, but I'm occupied with some other stuff right now.

If you want to help with that work I would gladly receive any help I can get. Weirdly what has to be done now is to migrate docs away from Gitbook to, preferably, gatsby. Unfortunately, Gitbook breaks current CI setup as it has a dependency on `npm` 🤦‍♂

henrikra:
Hmm add detailed list of stuff what should be done then? :D I just want to get redux-saga to be somewhat type safe because now it is far from it :/

Andarist:
Right now the docs migration is important - I suspect there should be some docs-like Gatsby theme that could be used to make this easy, we still want to author docs in markdown.

neurosnap:
Greetings!  We have introduced better typings with ts3.6.  We are watching this PR closely to get even better typing support for redux-saga: https://github.com/microsoft/TypeScript/pull/47548

PR #1975, opened 2019-11-09
Add localhost URL to docs for previewing the output

jtulk:
| Fixed Issues?            | None |
| Patch: Bug Fix?          |  No   |
| Minor: New Feature?      | No  |

**Documentation Update**
Running `npm start` creates the following terminal output:
```
[0000] info  Server running at http://10.60.10.6:9966/ (connect)
[0001] info  LiveReload running on 35729
```
Those unfamiliar with Budo are not prompted as to where the running code can be accessed in a browser. This change lists the output URL explicitly to make it easier for beginners.

### DIFF
--- a/docs/introduction/BeginnerTutorial.md
+++ b/docs/introduction/BeginnerTutorial.md
@@ -25,6 +25,8 @@ To start the application, run:
 $ npm start
 ```
 
+The output will now be accessible at `http://localhost:9966`. 
+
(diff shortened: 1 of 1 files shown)


Andarist:
It's very unlikely - but we might want to change the setup of that example in the future. Hardcoding this port in the documentation might bite us in the future. Maybe it would be better to change the `npm start` script to run with `--open` flag? What do you think?

PR #2010, opened 2020-02-03
update(packages/core/package.json): funding.url uses redux-saga

jakerobers:
xstate was listed as the opencollective url. I believe this should be
https://opencollective.com/redux-saga instead.

| Fixed Issues?            | N/A |
| Patch: Bug Fix?          | No  |
| Any Dependency Changes?  |  No   |


Andarist:
Thanks!

PR #2038, opened 2020-03-05
Docs fix

achillesoo7:
Minor doc fix which can bypass a lot of confusions and it's matching the standards of the documentation.


Andarist:
@achillesoo7 could you point me to places in the documentation where the convention used by you is more prominent?

achillesoo7:
@Andarist it's everywhere, the action name should be in the string, right? but at these places, it's not.

PR #2044, opened 2020-03-26
docs(rootSaga): add the reference link to issue#570 in rootSaga section

Mistyyyy:
Just add the reference link to issue#570 in rootSaga section. which is conveniently direct to the relative issues when reading the docs.


Andarist:
thanks!

PR #2048, opened 2020-04-01
Types: change the type for call() to support generics

jeremy-deutsch:
| Fixed Issues?            | #1286 mentions this |
| Patch: Bug Fix?          |  Should probably be a minor  |
| Any Dependency Changes?  |  `dtslint` 0.9.0 -> 3.4.1  |

Generic functions have never worked super-well with the types for `call`, and as of TS 3.8, creating a `call` effect with a generic function will always return a `CallEffect<unknown>`. Additionally, trying to call `call` with a generic higher-order function gets rejected by the TS compiler in all scenarios I've seen.

In this PR, I've changed the type for `call` to be generic over the parameters and return type of its first argument, rather than being generic over the type of the function itself. This seems to fix the problems that `call` had, and now it really supports regular and higher-order generics.

I also upgraded `dtslint`, since the older version only checked using TS 3.6 (where the higher-order generics wouldn't explicitly error), but the newer version should test using every valid version of TS for the file. I tried using `$ExpectType` in the places where there are now assignment checks, but it didn't work with the `CallEffect` type alias, so I figured assignment checks were fine.

Also, on consistency: this PR only changes the type for exactly one overload of `call`, and for no other effects in the project. If this approach seems like a good idea, I can add it to other functions like `fork` and `spawn`, and see if the other overloads of `call` could benefit from this approach (though it's probably less likely that people are using those with generics).

cc @gilbsgilbs

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -62,7 +62,7 @@
     "@babel/preset-env": "^7.6.3",
     "babel-7-jest": "^21.3.3",
     "babel-plugin-annotate-pure-calls": "^0.4.0",
-    "dtslint": "^0.9.0",
+    "dtslint": "^3.4.1",
     "jest": "^23.5.0",
     "lerna-alias": "^3.0.2",
     "mitt": "^1.1.2",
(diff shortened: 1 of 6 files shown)


neurosnap:
Greetings!  Thanks for your contribution and sorry for the delay.  Could you resolve the conflicts and we can discuss these changes again?

jsakas:
> Greetings! Thanks for your contribution and sorry for the delay. Could you resolve the conflicts and we can discuss these changes again?

@neurosnap I have done that here https://github.com/redux-saga/redux-saga/pull/2273

neurosnap:
Closing since #2273 supersedes.

PR #2065, opened 2020-05-15
Add pointer to the jasmine test framework.

RubenGarcia:
The example uses it() from jasmine, but there is no mention of that in the documentation, so new users unfamiliar with jasmine will be confused.

I found 
https://stackoverflow.com/questions/28353232/what-is-the-it-function-here-doing
which pointed me in the right direction.

| Fixed Issues?            | No issue mentions this, but i can open one if so desired |
| Patch: Bug Fix?          |  Documentation improvement   |
| Major: Breaking Change?  |   No |
| Any Dependency Changes?  |   No |

I added a pointer to the jasmine test framework in https://redux-saga.js.org/docs/advanced/TaskCancellation.html

### DIFF
--- a/docs/advanced/TaskCancellation.md
+++ b/docs/advanced/TaskCancellation.md
@@ -77,7 +77,7 @@ There is another direction where the cancellation propagates to as well: the joi
 
 ## Testing generators with fork effect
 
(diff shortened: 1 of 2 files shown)


RubenGarcia:
If the demo was supposed to be using https://github.com/antoinejaussoin/redux-saga-testing/
which also uses it(), I can change the patch accordingly.

RubenGarcia:
Fix typo in https://redux-saga.js.org/docs/advanced/Testing.html
middlware -> middleware

Andarist:
`describe`/`it` globals are not exclusive to Jasmine - so I think it's better to leave the name of a particular test runner out of this. Especially given the fact that the used here `expect` doesn't quite use Jasmine's matchers.

PR #2070, opened 2020-05-20
Demo for Real-world application

MishaJexia:
| Change | Description |
| Minor: New Feature?      |  Added demo button for Real-world example.  |

This PR adds Runme button-label to run the project from the latest commit with one click.
image

**How it works:**
1. Runme clones the repository and builds a docker image for latest commit;
2. deploys docker image to k8s cluster;
3. creates domain with SSL certificate;
4. shows web application;
5. destroys app instance after 10 minutes.

You can find detailed information on how this works [here](https://runme.io/how-it-works). The small demo you can find [here](https://www.youtube.com/channel/UCEysV237wo321JatUTC6Rlg).

**Why do you need this?**
- Runme is perfect solution to demonstrate the current example of code/repository, anybody can just click the button and see the state;
- Runme totally free, we love open source projects and want to support them;
- Runme has only one limit: you can run one commit no more than once every 5 minutes;

### DIFF
--- a/.runme/Dockerfile
+++ b/.runme/Dockerfile
@@ -2,4 +2,4 @@ FROM node:14
 COPY ./examples/real-world /app
 WORKDIR /app
 RUN npm install -s --no-progress
-ENTRYPOINT npm run start
+ENTRYPOINT cross-env BABEL_ENV=cjs node --require @babel/register server.js
--- a/.runme/Dockerfile
+++ b/.runme/Dockerfile
@@ -1,0 +1,5 @@
+FROM node:14
+COPY ./examples/real-world /app
+WORKDIR /app
+RUN npm install -s --no-progress
+ENTRYPOINT npm run start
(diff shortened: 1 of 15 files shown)


Andarist:
The tool itself looks really cool, but I'm rather wary about introducing any new tools into the codebase. We are already relying on CodeSandbox which has similar capabilities and more people in the JS/React community know it.

MishaJexia:
Sure, no push. Take your time to check the tool. I can give extra info about it if needed. With Runme you will be running the latest code that you have in GitHub. And it will happen without any extra step. Not sure if it is the same o CodeSandbox. And with this RP there is no code change, it is only extra folder .runme. Which can be easily removed in case of any troubles. 
For readme.md I've added link after 1 example - a real-world application that can be easily removed as well.  I personally do not see a big change in the community's way of working. But I might do not see some details what you have. :) I will let you decide "to be or not to be" © talented guy but feel free to ask me if extra info is needed. As well I can continue to support in case of any issue appears with runme runer. cheers

PR #2159, opened 2021-02-08
Move documentation to Docusaurus

### LINKED ISSUE #1469: Update GitBooks
I've noticed that redux' docs got a nice lifting, thanks for the new version of GitBook. Would be nice to upgrade too.

Redux commit introducing the change - https://github.com/reduxjs/redux/commit/b307091af4d7e846d9e5f080fb81caf4a8b4aab1

We might want to postpone merging this though, because we don't want to update docs before releasing v1 and from what I understand new Gitbook comes with auto-releases.

acdvs:
| Fixed Issues?            | Closes #1469 |
| Tests Added + Pass?      | No new tests |

This begins the documentation move from GitBook to [Docusaurus](https://docusaurus.io) since GitBook is no longer supported as OSS.

The new documentation will have a full home page (to be filled out more over time).


Andarist:
Please ping me whenever you feel this can be reviewed/looked at - I'm not sure what's the current stage of this PR (It's a draft on GitHub so not sure if I should dive into it just now). I really appreciate you sharing the work publicly right from the start though!

acdvs:
@Andarist It's pretty much ready to go now. The content could always be improved upon later, but for the most part everything is there. Let me know your initial thoughts.

Andarist:
Awesome! I will try to take a look at this by the end of the week.

Andarist:
@acdvs I haven't forgotten about this - gonna take a look in the following days, was busy with some other work and didn't manage to get to this one yet.

website/package.json
```diff
@@ -0,0 +1,36 @@
+{
```
Andarist, line 1: I wasn't sure how I'm supposed to install this - given the monorepo structure and all. So I've added locally `"website"` to workspaces here:
https://github.com/redux-saga/redux-saga/blob/b3664a10d3421e2445b596159195e947fc365cc4/package.json#L3

Seems like the best approach for it, WDYT?

website/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "private": true,
+  "scripts": {
+    "docusaurus": "docusaurus",
+    "start": "docusaurus start",
+    "build": "docusaurus build",
+    "swizzle": "docusaurus swizzle",
+    "deploy": "docusaurus deploy",
+    "serve": "docusaurus serve",
+    "clear": "docusaurus clear"
+  },
```
Andarist, line 13: could u explain what those scripts are for? I'm mainly interested in:
- docusaurus
- swizzle
- deploy (if there any config for deployment?)

Andarist (commented):
1. I've noticed that "Edit on GitHub" button (the one below articles) is not working - it's pointing to an incorrect location

2. Since URLs have changed - I'm wondering: would it be possible to add a redirect (even client-side) for `.html` pages (current) to the ones without that? It seems that it might allow us to just migrate to new URLs without breaking the old ones, WDYT?

website/package.json
acdvs: I don't think adding anything to the workspaces is necessary here because the website is completely independent from the rest of the repo in terms of files and modules. As far as I know (correct me if I'm wrong), including it as a workspace would make it available to the rest of the repo, which has no use.

website/package.json
acdvs: Those scripts are all created by default with a new Docusaurus installation. Here's [the documentation](https://v2.docusaurus.io/docs/cli) on what each does, though it doesn't seem to mention the first `docusaurus` script.

- `swizzle` lets you customize specific theme components
- `deploy` deploys to GH Pages and is documented futher [here](https://v2.docusaurus.io/docs/deployment#deploying-to-github-pages)
  - As the owner, you'll need to set up specific settings so the `deploy` script can run through a GH action or another CI

acdvs:
There's a [Docusaurus plugin](https://v2.docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects) which can do extension redirects (e.g., `/page.html` to/from `/page`) and path redirects (e.g., `/old/path` to/from `/new/path`). I'll set it up, but note that it only takes effect on production and not local dev.

website/docusaurus.config.js
```diff
@@ -145,5 +145,8 @@ module.exports = {
       },
     ],
   ],
-  plugins: ['docusaurus-plugin-sass'],
+  plugins: [
+    'docusaurus-plugin-sass',
+    ['@docusaurus/plugin-client-redirects', { fromExtensions: ['html'] }],
```
Andarist, line 150: easier than I've imagined 🤣

website/package.json
Andarist: Yes, you are right - I was mainly thinking about the installation part of this. I guess we can just have an additional `yarn.lock` in the `website` directory that would only be installed when working on the docs.

website/package.json
Andarist: I guess the `docusaurus` just acts as a proxy to its commands and options, so u can do smth like `yarn docusaurus start --flag` or whatever

Thansk for the links!

Andarist:
Thanks for all the hard work! It looks awesome - I will give it another look before merging but I don't expect this to be delayed any further. I should release this in the upcoming week - I hope that I won't screw up the deployment 😅

website/docusaurus.config.js
acdvs: Partly why this framework is so great! 😎

acdvs:
Great! Thanks for all the feedback. Let me know if you have any other questions until then.

Andarist:
It's alive! Thank you for all your hard work and for being patient with me.

I've noticed now that on the deployed site there are 2 problems:
- images are not served correctly from the static directory, which is somewhat odd as they are correctly placed there in the repository
- OpenCollective Backers/Sponsors are empty - the HTML there is pretty much empty 🤔 

Would you be able to take a look at those?

Issue #1800, opened 2019-02-28
unable to make API call using map of items with yield all [ ]


NikhilDevop:
I have an array of items where I make API call for each item in array. I am using react native debugger for debugging ([https://github.com/jhen0409/react-native-debugger][1]), the problem is when I place debugger statement in the function where I make API call (i.e., inside callAPI function) the debugger doesn't stop there and is not even entering into callAPI function,

But when I remove the '[ ]' symbol in the yield call([ ]) the debugger stops in the callAPI function and an error " Objects are not valid as a React child ( found Error: Request failed with status code 400). If you meant to render a collection of children, use an array instead) "  is thrown. The following is the syntax with which I am facing problem.

    const callAPI = async ( client: ServiceRootClient, move: moveReq): Promise<any> => {

            debugger;

            const results = await client

                .forOne(singletons => singletons.find(singleton => 

                 singleton.$Name === "Item"))

                .forOperation(ops => ops.find(op => op.$Name === "Move"))

                .withParameters((params, parameterSet) => {

                    params.map(param => {

                        if (param.$Name === "Data") {

                            parameterSet.add(param, movement);

                        }

                    });

                })

                .execute();
        
        return results;
    };
    
  

     export function* applyItemMovements(action: itemManagement.Action) {

      try {
    
    	const result = yield all([

                    ItemMoveRequests.map(ItemMoveRequest=>

                        call(callAPI, action.metadata.client, moveReq);

                    )

                ]);

          } catch(error) {

                //print error

          }

    }

I am using '[ ]' inside yield all to make API calls for all array items so that if one API call fails that doesn't stop the other items in the array calling the API.

Also, I am not sure whether this is the way to make API call so that remaining items call API even if one item in array fails or should I do any modifications in the yield. How to make a call to  callAPI function? and can I also pass a label to each call of item ?

  [1]: https://github.com/jhen0409/react-native-debugger

shinima:
What you need may be...

```javascript

export function* applyItemMovements(action) {
  try {
    // create an array of promises
    const promises = ItemMoveRequests.map(ItemMoveRequest => callAPI(action.metadata.client, moveReq))
    // pass `promises` to all
    const result = yield all(promises)
  } catch (error) {
    //print error
  }
}
```

---

> ...remaining items call API even if one item in array fails...

You can use `Promise#catch` to avoid errors in one API-call propogating to `all` effect.

```javascript
[...]

PR #2160, opened 2021-02-15
added "allSettled" functionality

goodwin64:
| Fixed Issues?            | https://github.com/redux-saga/redux-saga/issues/1800 |
| Patch: Bug Fix?          |  No  |
| Tests Added + Pass?      | Yes, but prettier warns about `console.warn` |
| Any Dependency Changes?  | No  |

Added the `allSettled` functionality with the same API as in Promise.

### DIFF
--- a/packages/core/src/internal/io.js
+++ b/packages/core/src/internal/io.js
@@ -279,3 +279,16 @@ export function setContext(props) {
 }
 
 export const delay = call.bind(null, delayP)
+
+export const allSettled = effects =>
+  all(
+    effects.map(effect =>
+      call(function* settle() {
+        try {
+          return { status: 'fulfilled', value: yield effect }
+        } catch (err) {
+          return { status: 'rejected', reason: err }
+        }
+      }),
+    ),
+  )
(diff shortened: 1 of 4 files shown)


packages/core/src/internal/io.js
```diff
@@ -279,3 +279,16 @@ export function setContext(props) {
 }
 
 export const delay = call.bind(null, delayP)
+
+export const allSettled = effects =>
+  all(
+    effects.map(effect =>
```
Andarist, line 285: Currently, the redux-saga's code assumes ES5 environment and thus we cannot use generator syntax in our source code. All other helpers are implemented using the custom `fsmiterator` approach which looks like this: https://github.com/redux-saga/redux-saga/blob/8f8d9e1d8bb957e9d3253ea250b317fde4fafb09/packages/core/src/internal/sagaHelpers/throttle.js#L5 

I understand that this is somewhat quirky so if you don't want to look into rewriting this to accommodate our needs I will try to find the time to do it later myself - just let me know.

goodwin64: Thanks for the note. I need some time to read about the `fsm-iterator` approach to understand its' internals. I'll write here soon whether I can adjust this PR to this approach.

goodwin64: No ideas for now. I'd be grateful if you will take it once to work 🙏😇

Andarist: Sure - I will handle this sometime later. If you still are interested in understanding `fsmIterator` a little bit better I've dug up an old comment of mine which from what I remember shed some light on this one: https://github.com/redux-saga/redux-saga/issues/570#issuecomment-256717377

PR #2170, opened 2021-03-18
Fix website links and images

acdvs:
| Fixed Issues?            | n/a |
| Tests Added + Pass?      | n/a |

- Several links and images were broken due to absolute paths. Changed to relative paths.
- Code blocks (especially on the home page) were being rendered without the correct style unless the page was refreshed.

### DIFF
--- a/website/src/components/CodeSnippet.jsx
+++ b/website/src/components/CodeSnippet.jsx
@@ -1,6 +1,6 @@
 /* eslint-disable react/jsx-key */
 
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import Highlight, { defaultProps } from 'prism-react-renderer';
 
@@ -8,6 +8,12 @@ import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import useThemeContext from '@theme/hooks/useThemeContext';
 
 function CodeSnippet({ code }) {
+  const [mounted, setMounted] = useState(false);
+
+  useEffect(() => {
+    setMounted(true);
+  }, []);
+
   const { siteConfig: { themeConfig: { prism = {} }}} = useDocusaurusContext();
   const { isDarkTheme } = useThemeContext();
   const prismTheme = isDarkTheme ? prism.darkTheme : prism.theme;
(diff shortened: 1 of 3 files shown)


Andarist:
Sponsors are still not rendered correctly - they arrive in the SSRed HTML but are removed during React hydration (at least it looks like it). 🤔

website/src/components/CodeSnippet.jsx
```diff
@@ -1,20 +1,27 @@
...
 
 import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import useThemeContext from '@theme/hooks/useThemeContext';
 
 function CodeSnippet({ code }) {
+  const [mounted, setMounted] = useState(false);
+
+  useEffect(() => {
+    setMounted(true);
+  }, []);
```
Andarist, line 15: could you explain the rationale behind this? I understand what it does (remounts the `<Highlight/>`) but I'm wondering why this is needed here.

acdvs:
Still looking into this. Everything shows up fine in development (via both `start` and `build` methods) but it's not translating to the production build for some reason.

website/src/components/CodeSnippet.jsx
acdvs: It just ensures that the component and its contents are mounted _after_ the code style is loaded.

Andarist: what the code style (css for that?) is and how it relates to mounting?

Andarist:
> Still looking into this. Everything shows up fine in development (via both start and build methods) but it's not translating to the production build for some reason. I would've marked this as a draft if I knew this beforehand.

Not sure what is the exact Babel-related setup used here regarding dev & prod environments (didn't really looked into it) but I've debugged this and:
```js
[...new Array(COUNT)]
```
was being transpiled to:
```js
// most likely caused by Babel's "loose" mode
[].concat(new Array(COUNT))
```
which broke the iteration as this has been producing an array with empty slots instead of an array with undefineds. I've fixed this by using `Array.from`. This probably breaks compatibility with older browsers but the site has been built using CSS vars etc so that ship has already sailed 😉

acdvs:
Ah that makes sense. Didn't even consider how it was being transpiled. Thanks for the help!

PR #2234, opened 2021-12-02
clarify race() behavior when one effect throws

sjbarag:
| Fixed Issues?            | No issues filed |
| Patch: Bug Fix?          | Maybe – just a docs update, no behavior changes. |

The explanation surrounding the example `race()` use didn't previously mention that if `call(fetchUsers)` rejects, that rejection reason is thrown by `race`.  Clarify what happens when a racing effect throws.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -877,8 +877,10 @@ function* fetchUsersSaga() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


neurosnap (approved):
Thank you for your contribution!

PR #2241, opened 2022-01-13
Makes the Task interface generic

lourd:
Adds the generic type argument to the whole interface instead of individual methods. Upstreaming a [fix from typed-redux-saga](https://github.com/agiledigital/typed-redux-saga/blob/master/types/index.d.ts#L277-L282).

| Fixed Issues?            | No open issue for this |
| Patch: Bug Fix?          | Yes, in a way |
| Minor: New Feature?      | Not really  |


Andarist:
I agree with the change but it's on a verge of being a breaking change. Could you elaborate on practical problems with the current types? What kind of code will be affected (improved?) by this?

lourd:
The key use case is being able to infer the return type of `yield join(task)`. See https://github.com/agiledigital/typed-redux-saga/pull/648.

To avoid it being a breaking change, should I change the signature from `T extends any` to `T = any`?

packages/types/types/index.d.ts
```diff
@@ -143,7 +143,7 @@ export type END = { type: '@@redux-saga/CHANNEL_END' }
  * The Task interface specifies the result of running a Saga using `fork`,
  * `middleware.run` or `runSaga`.
  */
-export interface Task {
+export interface Task<T extends any> {
```
Andarist, line 146:
```suggestion
export interface Task<T = unknown> {
```

packages/types/types/index.d.ts
```diff
@@ -155,7 +155,7 @@ export interface Task {
   /**
    * Returns task return value. `undefined` if task is still running
    */
-  result<T = any>(): T | undefined
+  result(): T | undefined
```
Andarist, line 158:
```suggestion
  result<R = T>(): R | undefined
```

packages/types/types/index.d.ts
```diff
@@ -165,7 +165,7 @@ export interface Task {
    * - resolved with task's return value
    * - rejected with task's thrown error
    */
-  toPromise<T = any>(): Promise<T>
+  toPromise(): Promise<T>
```
Andarist, line 168:
```suggestion
  toPromise<R = T>(): Promise<R>
```

Andarist:
What do you think about the suggested changes? I'm also still thinking if maybe we should just go with your initial suggestion. I've recently written a SemVer Policy for XState and I could just follow my own line of thinking here as well: https://github.com/statelyai/xstate#typescript-changes . It's just a little bit scarier to do such breaking changes here, considering that Redux Saga is not in active development right now.

lourd:
I think making `T` default to `unknown` would be more of a breaking change than defaulting it to `any`, since that would cause type errors anywhere someone is using the value from `T.result()` or `T.toPromise()`.

Also, thinking on this more, doing `Task<T extends any>` is no different than `Task<T>`. 

So I would recommend going with `Task<T = any>` if you want to avoid build breakages out there, or `Task<T>` if it's an acceptable level of change.

lourd:
Hey @Andarist, can we move forward on this please?

lourd:
Hey @Andarist & @neurosnap, can we move forward on this or #2004, what's left to resolve?

Andarist:
As this is basically a duplicate of https://github.com/redux-saga/redux-saga/pull/2004, I'm going to close this PR here and move forward with the other one

PR #2250, opened 2022-02-09
Add multicastChannel documentation

edolix:
| Q                        | A |
| Documentation  |  Yes  |

## What this PR does

It adds documentation for the `multicastChannel`. We're using them in the current project and i thought it would be super helpful to have some (basic) docs for others too!


neurosnap (approved):
Thanks so much for this contribution, I think the description and examples are great and we should merge.

docs/advanced/Channels.md
```diff
@@ -293,3 +293,90 @@ In the above example, we create a channel using the `channel` factory. We get ba
...
+  while (true) {
+    const payload = yield take(channel)
+    // Process the request
+    console.log('mainWorker', payload)
+  }
+}
+```
+
+To solve the problem we need to use the `multicastChannel` that will *broadcast* the action to all the workers simultaneously.
+> Note that using `take` with the `multicastChannel` requires us to pass the additional argument *pattern* - that you can use to filter the actions to `take`.
```
Andarist, line 343: It is indeed true:
https://github.com/redux-saga/redux-saga/blob/24e9a68d621fd2a57a29b3b80e4b54ecb22fa593/packages/core/src/internal/io.js#L38-L40

Although, in retrospect, it seems like a weird limitation. This should only be a requirement for `stdChannel`.

PR #2261, opened 2022-02-26
Update CI to use latest node versions

### LINKED ISSUE #2260: Update CI to modern nodejs versions
Right now our CI runs for [3 node versions](https://github.com/redux-saga/redux-saga/blob/master/.github/workflows/nodejs.yml#L11): 8, 10, and 12.

These are pretty old nodejs versions.  [v12 is EoL this April.](https://nodejs.org/en/about/releases/)  Furthermore we aren't testing the latest LTS versions (14, 16, 17).  My hunch is that support for 8, 10, and 12 aren't particularly useful.

I think we should just support 14, 16, 17 and reflect that in our CI.

neurosnap:
| Fixed Issues?            | Fixes #2260  |

The goal of this PR is to update CI to work on the latest versions of nodejs.

https://github.com/neurosnap/redux-saga/actions/runs/1917541095

### DIFF
--- a/.changeset/itchy-worms-give.md
+++ b/.changeset/itchy-worms-give.md
@@ -3,4 +3,4 @@
 'redux-saga': patch
 ---
 
-require cps callback for all overloads to support typescript 4.x
+Require `CpsCallback` in all functions passed to the `cps` effect creator. This fixes a regression caused by TS 4.0 changing the behavior around spreading `never` into tuple types
--- a/.changeset/itchy-worms-give.md
+++ b/.changeset/itchy-worms-give.md
@@ -1,0 +1,6 @@
+---
+'@redux-saga/core': patch
+'redux-saga': patch
+---
+
+require cps callback for all overloads to support typescript 4.x
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   ],
   "scripts": {
     "lint": "eslint packages/*/{src,test}",
-    "test": "lerna run --concurrency 1 --parallel test",
+    "test": "lerna run test",
     "clean": "lerna run --parallel clean && lerna clean --yes && rimraf node_modules",
     "build": "lerna run --parallel build",
     "counter": "yarn start --prefix examples/counter",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   ],
   "scripts": {
     "lint": "eslint packages/*/{src,test}",
-    "test": "lerna run --parallel test",
+    "test": "lerna run --concurrency 1 --parallel test",
     "clean": "lerna run --parallel clean && lerna clean --yes && rimraf node_modules",
     "build": "lerna run --parallel build",
     "counter": "yarn start --prefix examples/counter",
--- a/packages/core/types/ts3.6/effects.d.ts
+++ b/packages/core/types/ts3.6/effects.d.ts
@@ -529,7 +529,11 @@ export function apply<Ctx, Fn extends (this: Ctx, ...args: any[]) => any>(
   args: Parameters<Fn>,
 ): CallEffect<SagaReturnType<Fn>>
 
+type Cast<A, B> = A extends B ? A : B
+type AnyFunction = (...args: any[]) => any
+
 type RequireCpsCallback<Fn extends (...args: any[]) => any> = Last<Parameters<Fn>> extends CpsCallback<any> ? Fn : never
+type RequireCpsNamedCallback<Ctx, Name extends keyof Ctx> = Last<Parameters<Cast<Ctx[Name], AnyFunction>>> extends CpsCallback<any> ? Name : never
 
 /**
  * Creates an Effect description that instructs the middleware to invoke `fn` as
@@ -553,8 +557,8 @@ export function cps<Fn extends (...args: any[]) => any>(
  * `yield cps([localStorage, 'getItem'], 'redux-saga')`
  */
 export function cps<Ctx extends { [P in Name]: (this: Ctx, ...args: any[]) => void }, Name extends string>(
-  ctxAndFnName: [Ctx, Name],
-  ...args: CpsFunctionParameters<Ctx[Name]>
+  ctxAndFnName: [Ctx, RequireCpsNamedCallback<Ctx, Name>],
+  ...args: AllButLast<Parameters<Ctx[Name]>>
 ): CpsEffect<ReturnType<Ctx[Name]>>
 /**
  * Same as `cps([context, fn], ...args)` but supports passing `context` and
@@ -563,16 +567,16 @@ export function cps<Ctx extends { [P in Name]: (this: Ctx, ...args: any[]) => vo
  * `fn` can be a string or a function.
  */
 export function cps<Ctx extends { [P in Name]: (this: Ctx, ...args: any[]) => void }, Name extends string>(
-  ctxAndFnName: { context: Ctx; fn: Name },
-  ...args: CpsFunctionParameters<Ctx[Name]>
+  ctxAndFnName: { context: Ctx; fn: RequireCpsNamedCallback<Ctx, Name> },
+  ...args: AllButLast<Parameters<Ctx[Name]>>
 ): CpsEffect<ReturnType<Ctx[Name]>>
 /**
  * Same as `cps(fn, ...args)` but supports passing a `this` context to `fn`.
  * This is useful to invoke object methods.
  */
 export function cps<Ctx, Fn extends (this: Ctx, ...args: any[]) => void>(
-  ctxAndFn: [Ctx, Fn],
-  ...args: CpsFunctionParameters<Fn>
+  ctxAndFn: [Ctx, RequireCpsCallback<Fn>],
+  ...args: AllButLast<Parameters<Fn>>
 ): CpsEffect<ReturnType<Fn>>
 /**
  * Same as `cps([context, fn], ...args)` but supports passing `context` and
@@ -581,8 +585,8 @@ export function cps<Ctx, Fn extends (this: Ctx, ...args: any[]) => void>(
  * `fn` can be a string or a function.
  */
 export function cps<Ctx, Fn extends (this: Ctx, ...args: any[]) => void>(
-  ctxAndFn: { context: Ctx; fn: Fn },
-  ...args: CpsFunctionParameters<Fn>
+  ctxAndFn: { context: Ctx; fn: RequireCpsCallback<Fn> },
+  ...args: AllButLast<Parameters<Fn>>
 ): CpsEffect<ReturnType<Fn>>
 
 export type CpsFunctionParameters<Fn extends (...args: any[]) => any> = Last<Parameters<Fn>> extends CpsCallback<any>
--- a/packages/core/types/ts3.6/effects.d.ts
+++ b/packages/core/types/ts3.6/effects.d.ts
@@ -585,7 +585,7 @@ export function cps<Ctx, Fn extends (this: Ctx, ...args: any[]) => void>(
 
 export type CpsFunctionParameters<Fn extends (...args: any[]) => any> = Last<Parameters<Fn>> extends CpsCallback<any>
   ? AllButLast<Parameters<Fn>>
-  : never
+  : ['must provide a function with cps callback']
 
 export interface CpsCallback<R> {
   (error: any, result: R): void
--- a/packages/core/types/ts3.6/effects.d.ts
+++ b/packages/core/types/ts3.6/effects.d.ts
@@ -529,6 +529,8 @@ export function apply<Ctx, Fn extends (this: Ctx, ...args: any[]) => any>(
   args: Parameters<Fn>,
 ): CallEffect<SagaReturnType<Fn>>
 
+type RequireCpsCallback<Fn extends (...args: any[]) => any> = Last<Parameters<Fn>> extends CpsCallback<any> ? Fn : never
+
 /**
  * Creates an Effect description that instructs the middleware to invoke `fn` as
  * a Node style function.
@@ -542,8 +544,8 @@ export function apply<Ctx, Fn extends (this: Ctx, ...args: any[]) => any>(
  */
 export function cps<Fn extends (cb: CpsCallback<any>) => any>(fn: Fn): CpsEffect<ReturnType<Fn>>
 export function cps<Fn extends (...args: any[]) => any>(
-  fn: Fn,
-  ...args: CpsFunctionParameters<Fn>
+  fn: RequireCpsCallback<Fn>,
+  ...args: AllButLast<Parameters<Fn>>
 ): CpsEffect<ReturnType<Fn>>
 /**
  * Same as `cps([context, fn], ...args)` but supports passing a `fn` as string.
@@ -583,7 +585,9 @@ export function cps<Ctx, Fn extends (this: Ctx, ...args: any[]) => void>(
   ...args: CpsFunctionParameters<Fn>
 ): CpsEffect<ReturnType<Fn>>
 
-export type CpsFunctionParameters<Fn extends (...args: any[]) => any> = AllButLast<Parameters<Fn>>
+export type CpsFunctionParameters<Fn extends (...args: any[]) => any> = Last<Parameters<Fn>> extends CpsCallback<any>
+  ? AllButLast<Parameters<Fn>>
+  : never
 
 export interface CpsCallback<R> {
   (error: any, result: R): void
--- a/packages/core/types/ts3.6/effects.d.ts
+++ b/packages/core/types/ts3.6/effects.d.ts
@@ -583,9 +583,7 @@ export function cps<Ctx, Fn extends (this: Ctx, ...args: any[]) => void>(
   ...args: CpsFunctionParameters<Fn>
 ): CpsEffect<ReturnType<Fn>>
 
-export type CpsFunctionParameters<Fn extends (...args: any[]) => any> = Last<Parameters<Fn>> extends CpsCallback<any>
-  ? AllButLast<Parameters<Fn>>
-  : ['must provide a function with cps callback']
+export type CpsFunctionParameters<Fn extends (...args: any[]) => any> = AllButLast<Parameters<Fn>>
 
 export interface CpsCallback<R> {
   (error: any, result: R): void
--- a/packages/core/types/ts3.6/effects.test.ts
+++ b/packages/core/types/ts3.6/effects.test.ts
@@ -285,9 +285,6 @@ function* testApply(): SagaIterator {
 function* testCps(): SagaIterator {
   type Cb<R> = (error: any, result: R) => void
 
-  // $ExpectError
-  yield cps((a: number) => {})
-
   yield cps(cb => {
     cb(null, 1)
   })
@@ -296,7 +293,7 @@ function* testCps(): SagaIterator {
   })
 
   yield cps<(cb: Cb<string>) => void>(cb => {
-    cb(null, 1)  // $ExpectError
+    cb(null, 1) // $ExpectError
   })
   yield cps<(cb: Cb<number>) => void>(cb => {
     cb(null, 1)
--- a/packages/core/types/ts3.6/effects.test.ts
+++ b/packages/core/types/ts3.6/effects.test.ts
@@ -285,6 +285,9 @@ function* testApply(): SagaIterator {
 function* testCps(): SagaIterator {
   type Cb<R> = (error: any, result: R) => void
 
+  // $ExpectError
+  yield cps((a: number) => {})
+
   yield cps(cb => {
     cb(null, 1)
   })
@@ -293,7 +296,7 @@ function* testCps(): SagaIterator {
   })
 
   yield cps<(cb: Cb<string>) => void>(cb => {
-    cb(null, 1) // $ExpectError
+    cb(null, 1)  // $ExpectError
   })
   yield cps<(cb: Cb<number>) => void>(cb => {
     cb(null, 1)
--- a/packages/core/types/ts3.6/effects.test.ts
+++ b/packages/core/types/ts3.6/effects.test.ts
@@ -288,6 +288,9 @@ function* testCps(): SagaIterator {
   // $ExpectError
   yield cps((a: number) => {})
 
+  // $ExpectError
+  yield cps((a: number, b: string) => {}, 42)
+
   yield cps(cb => {
     cb(null, 1)
   })
--- a/packages/core/types/ts3.6/effects.test.ts
+++ b/packages/core/types/ts3.6/effects.test.ts
@@ -377,6 +377,10 @@ function* testCps(): SagaIterator {
       cb(null, this.foo)
     },
   }
+  const objWithoutCb = {
+    foo: 'bar',
+    getFoo(arg: string) {},
+  }
 
   // $ExpectError
   yield cps([obj, obj.foo])
@@ -386,6 +390,8 @@ function* testCps(): SagaIterator {
   yield cps([obj, obj.getFoo], 1)
   yield cps([obj, obj.getFoo], 'bar')
   yield cps<typeof obj, (arg: string, cb: Cb<string>) => void>([obj, obj.getFoo], 'bar')
+    // $ExpectError
+  yield cps([objWithoutCb, objWithoutCb.getFoo])
 
   // $ExpectError
   yield cps([obj, 'foo'])
@@ -395,6 +401,8 @@ function* testCps(): SagaIterator {
   yield cps([obj, 'getFoo'], 1)
   yield cps([obj, 'getFoo'], 'bar')
   yield cps<typeof obj, 'getFoo'>([obj, 'getFoo'], 'bar')
+  // $ExpectError
+  yield cps([objWithoutCb, 'getFoo'])
 
   // $ExpectError
   yield cps({ context: obj, fn: obj.foo })
@@ -403,6 +411,8 @@ function* testCps(): SagaIterator {
   // $ExpectError
   yield cps({ context: obj, fn: obj.getFoo }, 1)
   yield cps<typeof obj, (arg: string, cb: Cb<string>) => void>({ context: obj, fn: obj.getFoo }, 'bar')
+  // $ExpectError
+  yield cps({ context: objWithoutCb, fn: objWithoutCb.getFoo })
 
   // $ExpectError
   yield cps({ context: obj, fn: 'foo' })
@@ -412,6 +422,8 @@ function* testCps(): SagaIterator {
   yield cps({ context: obj, fn: 'getFoo' }, 1)
   yield cps({ context: obj, fn: 'getFoo' }, 'bar')
   yield cps<typeof obj, 'getFoo'>({ context: obj, fn: 'getFoo' }, 'bar')
+  // $ExpectError
+  yield cps({ context: objWithoutCb, fn: 'getFoo' })
 }
 
 function* testFork(): SagaIterator {
--- a/packages/redux-saga/package.json
+++ b/packages/redux-saga/package.json
@@ -29,7 +29,7 @@
   "bundlesize": [
     {
       "path": "./dist/redux-saga.umd.min.js",
-      "maxSize": "6.6 Kb"
+      "maxSize": "6.62 Kb"
     }
   ],
   "author": "Yassine ELOUAFI <yelouafi@gmail.com>",
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        node: [8, 10, 12]
+        node: [14, 16]
 
     steps:
       - uses: actions/checkout@v1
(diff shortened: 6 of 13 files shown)


packages/core/types/ts3.6/effects.test.ts
```diff
@@ -285,9 +285,6 @@ function* testApply(): SagaIterator {
 function* testCps(): SagaIterator {
   type Cb<R> = (error: any, result: R) => void
 
-  // $ExpectError
```
neurosnap, line 288: I'm not sure why this is failing in `dtslint`.  I can see the error inside my text editor but even after upgrading dtslint I'm stilling seeing the error.

I felt like the error check itself was relatively harmless so I removed it for now.  I'd be happy to keep digging into this to figure out why it's failing for node v16.

package.json
```diff
@@ -6,7 +6,7 @@
   ],
   "scripts": {
     "lint": "eslint packages/*/{src,test}",
-    "test": "lerna run --parallel test",
+    "test": "lerna run --concurrency 1 --parallel test",
```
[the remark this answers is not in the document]
Andarist: What would dropping `--parallel` from here do? I mean - semantically `--parallel --concurrency 1` is super weird so I wonder if it just wouldn't be more clean to drop `--parallel` from here?

packages/redux-saga/package.json
```diff
@@ -29,7 +29,7 @@
   "bundlesize": [
     {
       "path": "./dist/redux-saga.umd.min.js",
-      "maxSize": "6.6 Kb"
+      "maxSize": "6.62 Kb"
```
Andarist, line 32: q: any reason behind the change here? as far as i can see there are no changes in this PR that would affect the bundlesize

package.json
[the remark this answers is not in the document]
neurosnap: You are probably right, lemme try that.

packages/redux-saga/package.json
neurosnap: I'm assuming upgrading node created different outputs for some reason?  I can do a diff on the two bundles and see what has changed.

neurosnap: Running a production build on v12.22.10 I'm getting the same exact bundle size as I am on v16.  My guess is if we fixed the dslint issue on master we would still have a failing test for bundlesize.

My vote is to let this change through so we can focus on other cleanup issues.

Andarist: For reference - this is the last commit that has executed this check: https://github.com/redux-saga/redux-saga/commit/a8c968b77243f59fbe672ca56691761664e060bf . It's somewhat weird as it shouldn't even rely on other CI checks... 🤔 as in, it should execute regardless of the status of those other CI checks

q: how did you check that this value has to be adjusted in the first place? no CI checks were executed in this PR

neurosnap: It seems like once our CI has a single failure all the jobs bail.

> q: how did you check that this value has to be adjusted in the first place? no CI checks were executed in this PR

https://github.com/neurosnap/redux-saga/runs/5341759335?check_suite_focus=true

packages/core/types/ts3.6/effects.test.ts
neurosnap: Okay so I spent a few hours last night debugging this issue.  `dtslint` downloads `typescript@next` which is targeting [4.7.0-dev.20220301](https://www.npmjs.com/package/typescript/v/4.7.0-dev.20220301).

I'm not exactly sure _why_ but our type tests were relying on an error being expected that is no longer present in 4.x.

This line was not throwing an error in 4.x: https://github.com/redux-saga/redux-saga/blob/master/packages/core/types/effects.test.ts#L289

In 4.x that line was resolving to this overload: https://github.com/redux-saga/redux-saga/blob/master/packages/core/types/effects.d.ts#L516

Prior to 4.x that line would resolve to https://github.com/redux-saga/redux-saga/blob/master/packages/core/types/effects.d.ts#L550 which would throw the error we are expecting.

For whatever reason, in 4.x, they changed the behavior of rest parameters + the usage of `never`.  Instead of `never` I think what we actually wanted was to throw an error.

I saw that you posted in typescript on a proposal to add `throw` in a conditional type and that there are workarounds.  I don't think the code I supplied technically works the way we want it to, but by replace `never` with something that doesn't match the code provided for our test, it fails properly and everything works.

https://github.com/redux-saga/redux-saga/blob/master/packages/core/types/effects.d.ts#L555

Andarist: Thank you for the detailed investigation! I know how time-consuming this stuff can be so I appreciate that you have followed this rabbit hole.

It's somewhat weird that [TS playground on 3.x](https://www.typescriptlang.org/play?ts=3.6.3#code/JYWwDg9gTgLgBAbzgYzAZzgXzgMyhEOAcigFMATAVwA8BaNAQwHMGB6UnHU5GNIgKFCRYiOAGVmDAJIxSUBjGhZc+QiQo16kgf1LVh8HJQB2PYBGMAqOLLQwAwugAUASgBc4yTLkKlCfnBwAJ7ApAA25CjOTgwexpQgAEZyLnAAvAB8iJgu-JhAA) doesn't error for this case. Maybe that's some weirdness of the playground.

To the best of my understanding, [this TS playground](https://www.typescriptlang.org/play?ts=3.6.3#code/JYWwDg9gTgLgBAbzgGQIYGcYBo4CUCmAbvlOvnAL5wBmUEIcA5DAJ5j7oDGUwYMAtDACuYADb5GcAPRS4rdugBccAEwA6dQEYAUNuAA7GCWqpO5AMJh051KNEAjUwGsAPLgB8ibXDgAKEnRQyqj6LDhQHEKiMMq4AJTKhBDAACbecJwhZqIA-L4JcEmp2hS68uQAKqjAoi7IcPgAHkb6KehwISwA2gC6ngC8fr5qI6LKyHFw-Z6dk00tbUMAFsGhOCNqMXAG1CRwFZPTHaGTOftwyvpEJGVs5ACCdgBCQjBomHUNzfit7Z29Azw11I+BcVRqbmBZDq7lht3YcEs6AAYkJ9JwYMAIPoAAqoKCoED4IykFzI-RfBbtYYjfEAcyUx26PUOM1CgPeMBceIJRJJ6DJ+lhlJ+iyRNjsjk4rk67nSZ0eohebwwXJ5hOJJAF5LhPkuwN0KXwnFE+PI1DRGKxFM4VkFIt+fg29MZ-xZUzZLHcvmo+mU5PWtKgDOUSNR6Mx2PVfK1gvcBSKaW0tvQvl8qEuQhA9hIrMQFDiQA) should error (this would replicate the situation when in <4.0 this overload **didn't** match) but for some reason, it doesn't. Perhaps this is dependent on some compilation flags? I've tried turning off some strict settings but it didn't change the outcome.

It's an interesting case that might help me to understand some TS details - so I hope that I will find time this week to test this out locally. 

PS. you can hit `y` on GitHub to create a permalink with the hash in it, so the copied links in comments stay always pointing to the very same code lines (I've already updated links in your comment).

neurosnap: I played around with it a little more, I think < 4.0 the rest parameter (the params following (a: number) => {} inside `cps`) was getting evaluated to `[]` which is not assignable to never.  However, in > 4.0 `[]` can be assignable to `never` so it is fitting into that slot.

Regardless, in our types we assumed that for `cps` the `CpsFunctionParameters` will never hit the negative on the conditional type, which is incorrect in > 4.0 which is why it’s being permitted (`[]` can be assigned to `never`).

Andarist: For my own reference, so I don't lose this, here we have a TS playground reproducing the error in 3.6:
[TS playground](https://www.typescriptlang.org/play?ts=3.6.3#code/C4TwDgpgBACgThSA7AJgHgCoFcwBtoQAewEqAzlAIZIgDaAugDRQCCKKpKAfFALxQAKAQH0AXK3admAOlnCAjOOx4IASj49qIdURLlBA2dLFQAlkgBmEOFABKEMllzB1vTTXUB+Ow6fAo4kgQAG7WANwAUBGgkFAAkmQAYuamJJg4+FC6nBRaDMzJSKkQzHGWKSQ8-ADeEVD1WQC2YKDihcWRDVBIAPZIAKLNrQZGJsr4ru7aWcQ5BibmVjbJcGTAMnKKZpbWPmuTVB5Q3gntafZrBRUl8eVFlQHdIeF1DYvX4mUW15EAvrTjAizfQMY5QADkECGIHBj0BMz0KAoAkWu36+EapBcoO86KhWNB2X08O84Pe9wgsPE4N6A2hVKeoTg9CiHAAxrhKAgoDFoMIADKUNbpFQIuZ5JhQAAiEAslD8fEZ1iqUFqXShLRA4hlcr8nQaZHMAHN8EoMkDERRaKibABlHr4PGYpDAehg+2OjFYx5BJn6+qNPymcRCUZmlQHLQ6YFI+biLQbYxbG1QAByukjR28guF6eIPG1svlzn922+FPDmSJsZROxsTqxqhxUAbLp9zzgfwB5rFILdpI1oAZ8OrVq0-YhhqQJspcJ7o8EKdb2Iny8JMYoJIh5OKDPBgecpgZvusLOi4GgOeAIqrG8OdHoKoFQuvgK4kXPsWE9iZZAgN4tcUaHyWAEG+Qhe1jCVFQYFU1QaQctVA2VTEIUtaUGTUQ0MORKzUDR72jS04zLXYVkuKBRmTOs0wzAiozBb8Oz-NA83WZDkHQeAUMIK5VmALgCyVTtXnqHcSHEeCunqfpbFsAB5WxqQAYWoXp-AQX9oGoMtrh5c1wVLLplPkqV+lTFgAFl+mpL5rkBQzRKgX4uxHO9aATSjZAlCdTmuADmBpPpMKHQLxMpITIXpM9eR8LSAMg3JgMfRUmPit8P1ijBKFMXA0H5RL71gxVQ1kXBxH5TNpgXIQAAt4xoRNgHEFMMCqrwoAwdsmSiWKWFwXAACEsGAK98sKiUVR-awWOy3K0Gm1Z-35QSuCicwSDgOU2WgZSwDIVSBoAI0oNkAGsFp4eCBGsOAejgBqQGYBBHGccRbFUcRgh6UwUFeNlqB23BPAET6oG+36Il+XqLygPakiwJA2WAUw+hgLlKExTayDQRIkEKnDpC5I0yEehgqpVMb0bgTGIGx3GkEEwr4cO3ATvOtAtDW+pvH6oaRqpjGsZmhnBNeQIO1ZCAOS5aALER5HUfxtl9oZgmjGJ0mivoCmBFeCwkDaJBGFeDW4BJ8R4cSBWUbRoW6ZFvG1rBiG-oiFWyCEShAiwRojusA5ql+VQgA)

packages/core/types/ts3.6/effects.d.ts
```diff
@@ -583,9 +583,7 @@ export function cps<Ctx, Fn extends (this: Ctx, ...args: any[]) => void>(
   ...args: CpsFunctionParameters<Fn>
 ): CpsEffect<ReturnType<Fn>>
 
-export type CpsFunctionParameters<Fn extends (...args: any[]) => any> = Last<Parameters<Fn>> extends CpsCallback<any>
-  ? AllButLast<Parameters<Fn>>
-  : ['must provide a function with cps callback']
+export type CpsFunctionParameters<Fn extends (...args: any[]) => any> = AllButLast<Parameters<Fn>>
```
Andarist, line 586: Hmm, is this change correct? Now both 3.6 and 4.x **don't error** on the case that we are discussing and also we'd fail to error on this:
```ts
yield cps((a: number, b: string) => {}, 42)
```

neurosnap: CI is passing for it: https://github.com/neurosnap/redux-saga/actions/runs/1917541095

packages/core/types/ts3.6/effects.test.ts
Andarist: I think your analysis is mostly correct but in reality, we hit the negative branch of this conditional in all of those TS versions (take look at the bottom):
[TS playground 3.6](https://www.typescriptlang.org/play?ts=3.6.3#code/C4TwDgpgBACgThSA7AJgHgCoFcwBtoQAewEqAzlAIZIgDaAugDRQCCKKpKAfFALxQAKAQH0AXK3admAOlnCAjOOx4IASj49qIdURLlBA2dLFQAlkgBmEOFABKEMllzB1vTTXUB+Ow6fAo4kgQAG7WANwAUBGgkFAAkmQAYuamJJg4+FC6nBRaDMzJSKkQzHGWKSQ8-ADeEVD1WQC2YKDihcWRDVBIAPZIAKLNrQZGJsr4ru7aWcQ5BibmVjbJcGTAMnKKZpbWPmuTVB5Q3gntafZrBRUl8eVFlQHdIeF1DYvX4mUW15EAvrTjAizfQMY5QADkECGIHBj0BMz0KAoAkWu36+EapBcoO86KhWNB2X08O84Pe9wgsPE4N6A2hVKeoTg9CiHAAxrhKAgoDFoMIADKUNbpFQIuZ5JhQAAiEAslD8fEZ1iqUFqXShLRA4hlcr8nQaZHMAHN8EoMkDERRaKibABlHr4PGYpDAehg+2OjFYx5BJn6+qNPymcRCUZmlQHLQ6YFI+biLQbYxbG1QAByukjR28guF6eIPG1svlzn922+FPDmSJsZROxsTqxqhxUAbLp9zzgfwB5rFILdpI1oAZ8OrVq0-YhhqQJspcJ7o8EKdb2Iny8JMYoJIh5OKDPBgecpgZvusLOi4GgOeAIqrG8OdHoKoFQuvgK4kXPsWE9iZZAgN4tcUaHyWAEG+Qhe1jCVFQYFU1QaQctVA2VTEIUtaUGTUQ0MORKzUDR72jS04zLXYVkuKBRmTOs0wzAiozBb8Oz-NA83WZDkHQeAUMIK5VmALgCyVTtXnqHcSHEeCunqfpbFsAB5WxqQAYWoXp-AQX9oGoMtrh5c1wVLLplPkqV+lTFgAFl+mpL5rkBQzRKgX4uxHO9aATSjZAlCdTmuADmBpPpMKHQLxMpITIXpM9eR8LSAMg3JgMfRUmPit8P1ijBKFMXA0H5RL71gxVQ1kXBxH5TNpgXIQAAt4xoRNgHEFMMCqrwoAwdsmSiWKWFwXAACEsGAK98sKiUVR-awWOy3K0Gm1Z-35QSuCicwSDgOU2WgZSwDIVSBoAI0oNkAGsFp4eCBGsOAejgBqQGYBBHGccRbFUcRgh6UwUFeNlqB23BPAET6oG+36Il+XqLygPakiwJA2WAUw+hgLlKExTayDQRIkEKnDpC5I0yEehgqpVMb0bgTGIGx3GkEEwr4cO3ATvOtAtDW+pvH6oaRqpjGsZmhnBNeQIO1ZCAOS5aALER5HUfxtl9oZgmjGJ0mivoCmBFeCwkDaJBGFeDW4BJ8R4cSBWUbRoW6ZFvG1rBiG-oiNk+jWKADZKyhAiwRojusA5qmh939oEA3VBhr9ny9-hBZp4XVjQXkegsb3GZ4AB6bPugDoO4E-PlhH6O8E5fNBqdp+m04zg2mYXFnKGO06Lq5sF5EeAAGKBc6gbugA)
[TS playground 4.6](https://www.typescriptlang.org/play?ts=4.6.2#code/C4TwDgpgBACgThSA7AJgHgCoFcwBtoQAewEqAzlAIZIgDaAugDRQCCKKpKAfFALxQAKAQH0AXK3admAOlnCAjOOx4IASj49qIdURLlBA2dLFQAlkgBmEOFABKEMllzB1vTTXUB+Ow6fAo4kgQAG7WANwAUBGgkFAAkmQAYuamJJg4+FC6nBRaDMzJSKkQzHGWKSQ8-ADeEVD1WQC2YKDihcWRDVBIAPZIAKLNrQZGJsr4ru7aWcQ5BibmVjbJcGTAMnKKZpbWPmuTVB5Q3gntafZrBRUl8eVFlQHdIeF1DYvX4mUW15EAvrTjAizfQMY5QADkECGIHBj0BMz0KAoAkWu36+EapBcoO86KhWNB2X08O84Pe9wgsPE4N6A2hVKeoTg9CiHAAxrhKAgoDFoMIADKUNbpFQIuZ5JhQAAiEAslD8fEZ1iqUFqXShLRA4hlcr8nQaZHMAHN8EoMkDERRaKibABlHr4PGYpDAehg+2OjFYx5BJn6+qNPymcRCUZmlQHLQ6YFI+biLQbYxbG1QAByukjR28guF6eIPG1svlzn922+FPDmSJsZROxsTqxqhxUAbLp9zzgfwB5rFILdpI1oAZ8OrVq0-YhhqQJspcJ7o8EKdb2Iny8JMYoJIh5OKDPBgecpgZvusLOi4GgOeAIqrG8OdHoKoFQuvgK4kXPsWE9iZZAgN4tcUaHyWAEG+Qhe1jCVFQYFU1QaQctVA2VTEIUtaUGTUQ0MORKzUDR72jS04zLXYVkuKBRmTOs0wzAiozBb8Oz-NA83WZDkHQeAUMIK5VmALgCyVTtXnqHcSHEeCunqfpbFsAB5WxqQAYWoXp-AQX9oGoMtrh5c1wVLLplPkqV+lTFgAFl+mpL5rkBQzRKgX4uxHO9aATSjZAlCdTmuADmBpPpMKHQLxMpITIXpM9eR8LSAMg3JgMfRUmPit8P1ijBKFMXA0H5RL71gxVQ1kXBxH5TNpgXIQAAt4xoRNgHEFMMCqrwoAwdsmSiWKWFwXAACEsGAK98sKiUVR-awWOy3K0Gm1Z-35QSuCicwSDgOU2WgZSwDIVSBoAI0oNkAGsFp4eCBGsOAejgBqQGYBBHGccRbFUcRgh6UwUFeNlqB23BPAET6oG+36Il+XqLygPakiwJA2WAUw+hgLlKExTayDQRIkEKnDpC5I0yEehgqpVMb0bgTGIGx3GkEEwr4cO3ATvOtAtDW+pvH6oaRqpjGsZmhnBNeQIO1ZCAOS5aALER5HUfxtl9oZgmjGJ0mivoCmBFeCwkDaJBGFeDW4BJ8R4cSBWUbRoW6ZFvG1rBiG-oiNk+jWKADZKyhAiwRojusA5qmh939oEA3VBhr9ny9-hBZp4XVjQXkegsb3GZ4AB6bPugDoO4E-PlhH6O8E5fNBqdp+m04zg2mYXFnKGO06Lq5sF5EeAAGKBc6gbugA)

It's also quite understanding that if we adjust the value returned from that "negative" branch then... the error goes away. Even though originally it states this:
```
function cps<(a: number) => void>(fn: (a: number) => void, ...args: never): void
Argument of type '[]' is not assignable to parameter of type 'never'.
```

So this really has to be a change in TS that just starts to treating `never` differently in such situations with spreads.

Technically we could replace that value with `Parameters<Fn>`, as technically the passed in callback doesn't have to actually use the `cpsCallback` that gets injected into it by the saga's runtime. On the other hand... this most likely indicates an error on the user's side, as such effect just wouldn't ever complete, so it's probably better to be pragmatic here than "technically correct". In this very situation this would produce the error on the line that we are discussing but it would allow such a call: 
```
yield cps((a: number) => {}, 42)
```

When I think about it - I don't really see a better way to handle this here than what you've done before: using something that can't be matched. We can't use a good generic constraint as `(...args: any[], last: Foo) => any` is not valid. So we have to "defer" the error to conditional types etc. There is another way than to expect some artificial second argument here though and this could look something like this:
[TS playground](https://www.typescriptlang.org/play?ts=3.6.3#code/C4TwDgpgBACgThSA7AJgHgCoFcwBtoQAewEqAzlAIZIgDaAugDRQCCKKpKAfFALxQAKAQH0AXK3admAOlnCAjOOx4IASj49qIdURLlBA2dLFQAlkgBmEOFABKEMllzB1vTTXUB+Ow6fAo4kgQAG7WANwAUBGgkFAAkmQAYuamJJg4+FC6nBRaDMzJSKkQzHGWKSQ8-ADeEVD1WQC2YKDihcWRDVBIAPZIAKLNrQZGJsr4ru7aWcQ5BibmVjbJcGTAMnKKZpbWPmuTVB5Q3gntafZrBRUl8eVFlQHdIeF1DYvX4mUW15EAvrTjAizfQMY5QADkECGIHBj0BMz0KAoAkWu36+EapBcoO86KhWNB2X08O84Pe9wgsPE4N6A2hVKeoTg9CiHAAxrhKAgoDFoMIADKUNbpFQIuZ5JhQAAiEAslD8fEZ1iqUFqXShLRA4hlcr8nQaZHMAHN8EoMkDERRaKibABlHr4PGYpDAehg+2OjFYx5BJn6+qNPymcRCUZmlQHLQ6YFI+biLQbYxbG1QAByukjR28guF6eIPG1svlzn922+FPDmSJsZROxsTqxqhxUAbLp9zzgfwB5rFILdpI1oAZ8OrVq0-YhhqQJspcJ7o8EKdb2Iny8JMYoJIh5OKDPBgecpgZvusLOi4GgOeAIqrG8OdHoKoFQuvgK4kXPsWE9iZZAgN4tcUaHyWAEG+Qhe1jCVFQYFU1QaQctVA2VTEIUtaUGTUQ0MORKzUDR72jS04zLXYVkuKBRmTOs0wzAiozBb8Oz-NA83WZDkHQeAUMIK5VmALgCyVTtXnqHcSHEeCunqfpbFsAB5WxqQAYWoXp-AQX9oGoMtrh5c1wVLLplPkqV+lTFgAFl+mpL5rkBQzRKgX4uxHO9aATSjZAlCdTmuADmBpPpMKHQLxMpITIXpM9eR8LSAMg3JgMfRUmPit8P1ijBKFMXA0H5RL71gxVQ1kXBxH5TNpgXIQAAt4xoRNgHEFMMCqrwoAwdsmSiWKWFwXAACEsGAK98sKiUVR-awWOy3K0Gm1Z-35QSuCicwSDgOU2WgZSwDIVSBoAI0oNkAGsFp4eCBGsOAejgBqQGYBBHGccRbFUcRgh6UwUFeNlqB23BPAET6oG+36Il+XqLx8ABHLBTAQPaDsoY7TouxIkEKnDpC5I0yEehgqpVMaYC5ShMU2sg0CxwTCpRw7cBO860C0HhvCx7rrFZCAOS5aALCwJA2WAUw+igNl9tp7GaqMfHCaK+gSYEV4LCQd6IARpGIEZtHmYxmWuEYV55bgAnxH6oaRrJimqZmo21rBiG-oiKWyCEShAiwRojusA5ql+VQ3f2z3vd9-2CMD5gABYACZg-dsPuh9v24GYd2QwDoOo6DkOPYEL2U4j9PJf2rPc+z2OE4iIA)

Both solutions are IMHO fine though.

neurosnap: Thanks for digging deeper!  How would you like to proceed?  I'm okay with whatever path you think we should take.

packages/core/types/ts3.6/effects.d.ts
Andarist: This is surprising - could you try to check why that is? Last time I've checked this didn't pass on the master branch but it did after your change, so perhaps this is a missing test case that we should add to guard us from regressions

packages/core/types/ts3.6/effects.test.ts
Andarist: Let's use my proposed solution - it feels more in the spirit of traditional types~.

packages/core/types/ts3.6/effects.d.ts
Andarist: I can confirm that this example typechecks although it shouldn't:
[TS playground](https://www.typescriptlang.org/play?ts=3.6.3#code/C4TwDgpgBACgThSA7AJgHgCoFcwBtoQAewEqAzlAIZIgDaAugDRQCCKKpKAfFALxQAKAQH0AXK3admAOlnCAjOOx4IASj49qIdURLlBA2dLFQAlkgBmEOFABKEMllzB1vTTXUB+Ow6fAo4kgQAG7WANwAUBGgkFAAkmQAYuamJJg4+FC6nBRaDMzJSKkQzHGWKSQ8-ADeEVD1WQC2YKDihcWRDVBIAPZIAKLNrQZGJsr4ru7aWcQ5BibmVjbJcGTAMnKKZpbWPmuTVB5Q3gntafZrBRUl8eVFlQHdIeF1DYvX4mUW15EAvrTjAizfQMY5QADkECGIHBj0BMz0KAoAkWu36+EapBcoO86KhWNB2X08O84Pe9wgsPE4N6A2hVKeoTg9CiHAAxrhKAgoDFoMIADKUNbpFQIuZ5JhQAAiEAslD8fEZ1iqUFqXShLRA4hlcr8nQaZHMAHN8EoMkDERRaKibABlHr4PGYpDAehg+2OjFYx5BJn6+qNPymcRCUZmlQHLQ6YFI+biLQbYxbG1QAByukjR28guF6eIPG1svlzn922+FPDmSJsZROxsTqxqhxUAbLp9zzgfwB5rFILdpI1oAZ8OrVq0-YhhqQJspcJ7o8EKdb2Iny8JMYoJIh5OKDPBgecpgZvusLOi4GgOeAIqrG8OdHoKoFQuvgK4kXPsWE9iZZAgN4tcUaHyWAEG+Qhe1jCVFQYFU1QaQctVA2VTEIUtaUGTUQ0MORKzUDR72jS04zLXYVkuKBRmTOs0wzAiozBb8Oz-NA83WZDkHQeAUMIK5VmALgCyVTtXnqHcSHEeCunqfpbFsAB5WxqQAYWoXp-AQX9oGoMtrh5c1wVLLplPkqV+lTFgAFl+mpL5rkBQzRKgX4uxHO9aATSjZAlCdTmuADmBpPpMKHQLxMpITIXpM9eR8LSAMg3JgMfRUmPit8P1ijBKFMXA0H5RL71gxVQ1kXBxH5TNpgXIQAAt4xoRNgHEFMMCqrwoAwdsmSiWKWFwXAACEsGAK98sKiUVR-awWOy3K0Gm1Z-35QSuCicwSDgOU2WgZSwDIVSBoAI0oNkAGsFp4eCBGsOAejgBqQGYBBHGccRbFUcRgh6UwUFeNlqB23BPAET6oG+36Il+XqLygPakiwJA2WAUw+hgLlKExTayDQRIkEKnDpC5I0yEehgqpVfqhpGsb0bgTGIGx3GkEE1kIA5LloAsRHkdR-G2X25mCaMYnSaK+gKYEV4LCQNokEYV4RbgEnxHhxIeZRtGMaxmbmbWsGIb+iIBbIIRKECLBGiO6xmCO8Q1jgY0DmqX5mAAFgAJlUIA)

packages/core/types/ts3.6/effects.test.ts
Andarist: I've pushed out this change and reverted stuff from [here](https://github.com/redux-saga/redux-saga/pull/2261/files#r825283246) + added a test that I've mentioned there.

We could potentially recheck if other overloads of `cps` don't suffer from some similar issues. Perhaps the `RequireCpsCallback` strategy should be used there too, that doesn't have to be done in this PR but it's also not the worst moment to investigate this.

Andarist: I've also pushed out now a new commit that introduces "cps callback requirement" to all other overloads. I've hacked this around somewhat quickly and I didn't have time to fully validate my changes (although they seem to work as far as I can tell, at least for TS >=3.6). Let's see what CI has to say about this.

neurosnap: Awesome, looks great!  How would you like to proceed with this PR?

Andarist: Hm, could u recheck if all CI checks pass now? Im on mobile and its hard to navigate this for me right now. I dont see any checks at all here - perhaps because this is a fork. To fix that we should add `pull_request` as a trigger here:

https://github.com/redux-saga/redux-saga/blob/b854e928514d7d5bdafb9e68fbeb097163ae8ee3/.github/workflows/nodejs.yml#L3

neurosnap: Okay I made your suggestion to our github action and pushed it up.  CI is now running for this PR and everything passed.

Andarist (approved):

neurosnap:
@Andarist I just reviewed your changes, they look good to me!  I say we merge.

Andarist:
The type changes are fixing an issue - we probably should add a changeset here that would describe this fix. Could you add one? If not - I will get to it later.

neurosnap:
Added a changeset.  Let me know if I need to make any changes!

.changeset/itchy-worms-give.md
```diff
@@ -0,0 +1,6 @@
+---
+'@redux-saga/core': patch
+'redux-saga': patch
+---
+
+require cps callback for all overloads to support typescript 4.x
```
Andarist, line 6:
```suggestion
Require `CpsCallback` in all functions passed to the `cps` effect creator. This fixes a regression caused by TS 4.0 changing the behavior around spreading `never` into tuple types
```

PR #2273, opened 2022-03-24
change the type for call() to support generics

jsakas:
I have only rebased the original pull request to resolve conflicts in the yarn.lock. This change would be useful for me as well.

Original pull request: https://github.com/redux-saga/redux-saga/pull/2048

> Q	A
> Fixed Issues?	#1286 mentions this
> Patch: Bug Fix?	Should probably be a minor
> Major: Breaking Change?	No
> Minor: New Feature?	Yes
> Tests Added + Pass?	Yes
> Any Dependency Changes?	`dtslint` 0.9.0 -> 3.4.1
> Generic functions have never worked super-well with the types for `call`, and as of TS 3.8, creating a `call` effect with a generic function will always return a `CallEffect<unknown>`. Additionally, trying to call `call` with a generic higher-order function gets rejected by the TS compiler in all scenarios I've seen.
> 
> In this PR, I've changed the type for `call` to be generic over the parameters and return type of its first argument, rather than being generic over the type of the function itself. This seems to fix the problems that `call` had, and now it really supports regular and higher-order generics.
> 
> I also upgraded `dtslint`, since the older version only checked using TS 3.6 (where the higher-order generics wouldn't explicitly error), but the newer version should test using every valid version of TS for the file. I tried using `$ExpectType` in the places where there are now assignment checks, but it didn't work with the `CallEffect` type alias, so I figured assignment checks were fine.
> 
> Also, on consistency: this PR only changes the type for exactly one overload of `call`, and for no other effects in the project. If this approach seems like a good idea, I can add it to other functions like `fork` and `spawn`, and see if the other overloads of `call` could benefit from this approach (though it's probably less likely that people are using those with generics).
> 
> cc @gilbsgilbs

### DIFF
--- a/packages/core/types/ts3.6/effects.d.ts
+++ b/packages/core/types/ts3.6/effects.d.ts
@@ -459,10 +459,10 @@ export interface ChannelPutEffectDescriptor<T> {
  *   Promise as result, or any other value.
  * @param args An array of values to be passed as arguments to `fn`
  */
-export function call<Fn extends (...args: any[]) => any>(
-  fn: Fn,
-  ...args: Parameters<Fn>,
-): CallEffect<SagaReturnType<Fn>>
+export function call<Args extends any[], Return>(
+  fn: (...args: Args) => Return | Promise<Return> | SagaIterator<Return>,
+  ...args: Args,
+): CallEffect<Return>
 /**
  * Same as `call([context, fn], ...args)` but supports passing a `fn` as string.
  * Useful for invoking object's methods, i.e.
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -62,7 +62,7 @@
     "@babel/preset-env": "^7.6.3",
     "babel-7-jest": "^21.3.3",
     "babel-plugin-annotate-pure-calls": "^0.4.0",
-    "dtslint": "^0.9.0",
+    "dtslint": "^3.4.1",
     "jest": "^23.5.0",
     "lerna-alias": "^3.0.2",
     "mitt": "^1.1.2",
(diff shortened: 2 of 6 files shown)


neurosnap:
Greetings!  Thanks for cleaning up this PR.  I just wanted to let you know that we are currently trying to update our CI in #2260 which also includes bumping dtslint to fix a bug in our type tests.  Once that PR is merged, I'll make reviewing this a priority.  Thanks for your patience!

packages/core/types/ts3.6/effects.d.ts
```diff
@@ -459,10 +459,10 @@ export interface ChannelPutEffectDescriptor<T> {
  *   Promise as result, or any other value.
  * @param args An array of values to be passed as arguments to `fn`
  */
-export function call<Fn extends (...args: any[]) => any>(
-  fn: Fn,
-  ...args: Parameters<Fn>,
-): CallEffect<SagaReturnType<Fn>>
+export function call<Args extends any[], Return>(
```
Andarist, line 462: Twitter thread that explains this change: https://twitter.com/TitianCernicova/status/1501237698991898624

neurosnap:
@jsakas Hello!  Could you resolve the conflicts against `master`?  I'll spend some time this week reviewing your PR.

neurosnap:
@jsakas Greetings!  Would you be willing to resolve the conflicts in this PR so I can review it?

neurosnap:
Greetings!  I'm going to close this PR due to inactivity.  If you want to continue work on it then please say so and I'll open it back up!

Issue #2276, opened 2022-04-03
I suppose it's better use .toEqual instead of assert.deepEqual


Andrii256:
This is not in code, this is documentation. So I removed template comments.
Here is link https://redux-saga.js.org/docs/introduction/BeginnerTutorial

There are code example where you use 
```js
test('incrementAsync Saga test', (assert) => {
// ...
assert.deepEqual(/* ... */)
// ...
```

I spend a while to google what `assert` is (I'm browser js engineer, not node.js engineer, so no: waht is `assert` was not obvious for me)

I think it's better replace by `.toEqual()` jest matcher.
May be it will not so beautifull, but it will be more clear for people who read.

Thanks

neurosnap:
Thanks for submitting this GH issue.  I understand the potential confusion since most people use `jest`, but rewriting the docs is time consuming and if we do rewrite them, it will be a holistic rewrite.

PR #2277, opened 2022-04-03
Replaced non-clear assert by clear .toEqual

Andrii256:
https://stackoverflow.com/questions/15313418/what-is-assert-in-javascript
For the real. I dont know what is assert and what does it means in current context. But I (and everybody who use jest) know what is .toEqual. so...

| ------------------------ | ----------------------------------------------------------------------  |
| Fixed Issues?               | [#2276](https://github.com/redux-saga/redux-saga/issues/2276) |
| Docs updated?           | Yes |

### DIFF
--- a/docs/introduction/BeginnerTutorial.md
+++ b/docs/introduction/BeginnerTutorial.md
@@ -323,28 +323,17 @@ import test from 'tape'
 import { put, call } from 'redux-saga/effects'
 import { incrementAsync, delay } from './sagas'
 
-test('incrementAsync Saga test', (assert) => {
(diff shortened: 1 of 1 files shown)


Andarist:
Those examples are using [tape](https://www.npmjs.com/package/tape) for testing. It's not that popular nowadays so I'm OK with replacing it - but then we should go over all of the examples and replace it everywhere

Andrii256:
@Andarist Mateusz, sorry for so long delay, but anyway. Now you can see fixed result in this PR https://github.com/redux-saga/redux-saga/pull/2286

PR #2295, opened 2022-05-09
Adds return type parameter to Saga

lourd:
| Fixed Issues?            | No existing issue |
| Minor: New Feature?      | 👍 |

In my application I have a function `runSaga` that takes a saga as a parameter and returns the `Task` that the saga returns when it's begun. I want to get the proper type of the Task's result, but am not able to do so with the current `Saga` type definition.

```ts
function* sag() {
  return 42
}

type RunSaga = <T extends any[]>(
  saga: Saga<T>,
  ...args: T
) => Task<R>

const result = runSaga(saga).result() // type of any
```

By adding `TReturn` to `Saga`, `result` will be inferred correctly

```ts
type RunSaga = <T extends any[], R>(
  saga: Saga<T, R>
  ...args: T
) => Task<R>

const result = runSaga(saga).result() // type of `number | undefined`
```

Here's a [TypeScript playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcDKBDA5hgPAQSiwGc5QZgA7AE1I0sQG0BdOAXjnqeYBo4AVAErAYAVyiV2nBgD4pACgB0yjEWIAuOIRIBKdnICSFKBhjRcXPkJHjKMgFD2kKOINGVMOKbn5kQFGjoGFj5BGXl7ODhibAxNTzx+UJkeSLhlRVUSTX57PTY5fgxiAGtcMMcAGxE4KHdNNw9Y719yKlppbmT5GJx42J9kvgysjQF8uQBvNOr4AEtjKV6MJRU1HRmaqGAxwU34SnIpBeAoRUP-eQ2ouYAzOHkLmEVqCEO9aaio7dIOJ8UAG4YSqiYBpAC+aW2YgkcE+Xx+okqME0V30tR2qS+cDmxAAwvQAMbASrVaioiZwW7A4jALFfXGNShzShYCno6mVWn0qJmAAKUAgAFtccB2QU4ALhaLFD8IJUAcB5AAWABMOh5cEJRJJ4qm4M1p0FUD1cEoSMqmtpMDxbwo-lNk0hUXBnFIRVK9kh9lu7kJMDmbwAVFSIBBHqIhZpzUKAEanD5QmywmNeqo1GDFEpSOqUeS3MN8NUbWYY34ILOynZImBXRz2FnGanEgRZnxSLhTNIAeiDQbScBDwhhlFIMDqqDuCAAFqhM6U4NPipQAOTwRBbZOHahwaAzwUAd0k9DIUGNA6D3bSjPczNZV00sbD1XoPb7F9cW7HE5x9xgs4rBcl1IeMqC1HVSWAagLyvG58QgskHzgJ95WAV8ol7fsviHL9AOzaFbDgIEQWARQ4AAA3cahgFuFkoPI388JxUhiADUlalvFksBgpNiBrJDfAAHzgKiaLo6CMPfbDPxHMcs33CAj1PY0yMomgxO3Bip3nbNcWiNjKg4yg7247DYOU6AkK4OBhNE2jtzfLCohw2TOElQURVpOAD2nOZCWnZiyAWWcTQ-ABaMt5UVHcD2CvCV1IAjYWI0FwoxAArYB-Sg7y4p0hKFKUo1oB43kIClTylR0TQKtFHwHEkpzBzgAlKGJLkZzneT5AMe4FkC1i5nY3MTOuZzzO1NqSSQgEIDmCTohEW1KHtGBcDxPwAg6CBY0y-1wjAQUwDGPlVADYF1pkaqiLm6DwSAA) I was using to play with the types

This is a non-breaking change since it's optional with a default of `any`.

This is very related, but slightly separate from #2241.


lourd:
Sorry for the double-ping @Andarist and @neurosnap, wasn't sure how to add reviewers. Thanks in advance for reviewing this 🙏

packages/types/types/ts3.6/index.d.ts
```diff
@@ -1,6 +1,6 @@
 import { Action } from 'redux'
 
-export type Saga<Args extends any[] = any[]> = (...args: Args) => Iterator<any>
+export type Saga<Args extends any[] = any[], TReturn = any> = (...args: Args) => Iterator<any, TReturn>
```
Andarist, line 3: Note for me (and other reviewers) - this change is unique to 3.6+ types because previous versions of the `Iterator` didn't accept more than one generic.

Andarist:
Could you add type-level tests for this change?

lourd:
👍  Done.

As part of doing that I also updated the types for `SagaMiddleware#run` and `runSaga`. It made sense to propagate the return type out to those functions and use them for tests, now that the generic Task changes have been merged.

lourd:
What do you think @Andarist?

packages/core/types/ts3.6/index.d.ts
```diff
@@ -166,7 +166,7 @@ export interface SagaMiddleware<C extends object = {}> extends Middleware {
    * @param saga a Generator function
    * @param args arguments to be provided to `saga`
    */
-  run<S extends Saga>(saga: S, ...args: Parameters<S>): Task
+  run<Args extends any[], TReturn>(saga: Saga<Args, TReturn>, ...args: Args): Task<TReturn>
```
Andarist, line 169: This is a change in the declared signature. What does it give us? Can you come up with any real-life scenario? 

I know *some* stuff about the difference between the two but I'd like to hear out your thoughts and motivation.

lourd: It is a change, but given it's just changing inferred generic type parameters, I don't think it's a breaking change. I don't think it's common for people to specify the saga's type, they just pass the saga.

I'm not sure what kind of description or specific use cases you're looking for beyond the example I gave in the PR description, and of course the tests showing that the task's type was not being inferred before.

> In my application I have a function runSaga that takes a saga as a parameter and returns the Task that the saga returns when it's begun. I want to get the proper type of the Task's result, but am not able to do so with the current Saga type definition.

lourd: Please let me know if I can be clearer or if you think the tests don't demonstrate the improvement — getting an specific, inferred type for the task's return instead of `any`

Andarist: But does it provide a meaningful improvement anywhere? Can we add a test that would "prove" that?

lourd: I'm confused, are you not seeing the two tests that are a part of this PR?

Andarist: I see them but I don't see from them how exactly changing the generics here has made them better. 

What I mean is that to make them work it would probably be enough to just extract the return type like this:
```ts
type SagaReturnType<S extends Saga<any, any>> = S extends Saga<any, infer R> ? R : never

// ...
run<S extends Saga>(saga: S, ...args: Parameters<S>): Task<SagaReturnType<S>>
```

lourd: Ah I see — you understand the point and value of adding the return type to `Saga`, but you have a way you would prefer to implement this piece of it.

I don't have a preference between the two forms, the way I wrote it was just how I'd figured out how to do it.

Went ahead and changed it to your suggested form 👍

packages/types/types/ts3.6/index.d.ts
```diff
@@ -2,7 +2,9 @@ export interface Action<T extends string = string> {
   type: T
 }
 
-export type Saga<Args extends any[] = any[]> = (...args: Args) => Iterator<any>
+export type Saga<Args extends any[] = any[], TReturn = any> = (...args: Args) => Iterator<any, TReturn>
+
+export type SagaReturnType<S extends Saga<any, any>> = S extends Saga<any, infer R> ? R : never
```
Andarist, line 7: There is a case in which we might get `never` back from this. I don't see any particular problems with this... but it's fine somewhat coincidentally. It's fine because the computed return type is only used within `toPromise` and `result`.

It's fine for `toPromise` as a never-returning iterator should, well, never complete - so the returned promise won't ever resolve and thus it doesn't quite matter.

It's fine for `result` because `T` is there included within a union with `undefined`, `never` is always eliminated from unions, so the only leftover will always be just `undefined` in such a case and that's correct.

This [TS playground](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mAVAAhmYBTATljATAYTnCmwAoBKZAbwChkHkAbDKVZAXmQAZ7GB3ABYwWyMlCwgMVOoznIAnjAxM8qANTrkfBgF9a+2hAQBnNjDZc0mHPiIlyFWrVCRYCFAEMwYOFE+khMRgpFiUAFzIAOIYYNgBcFgAPGAgALYARtgANMhxAG7YAHw0Osys7Fy8ckIiGGISUjJlckoqajCa2nL6hsZgZqhQAEycyN6+-oH2IY7OUAoADvUAyp4A5p5JAIJY6ybIGAAepGB4B94KANoAumOXt7kAKgBKrCBYYPdgCiVcZAA6IGePYmSK7fZUDglACSoQSyUuzzeUA+YCK8yWqw2nhRaKeWKSK0OJ1i52Qa02SSR4x+RT+FJJp3JlK2NOs2GQLxKAH4uchIgVsJjlsgnhgzABGMasvGfAnLJILZZwYCodDYXBBBxYEoAej1yH4iQA1iZaAbGAA9Hki+risyjLiy97ywnKjCq2mTAJ2YKhfWGvBwCVDZAQXC+2kKZCLLBwDIsNIHYCJZAgEy2y0MG3OWh4DAQJgg+quaDwL6eCAQDCLKArCRodZkMxYSKtpsUSL5OAwPDOKs1usNrBNsjUXTjA5CrBOWiD2v1xtgZsTqd5DCFLDIAA+6bOGGAaHwFCAA) might show what kind of edge cases I was thinking about while reviewing this change.

I just wonder if we should have any tests to check against potential regressions here. 🤔 Maybe I'm overthinking this though.

It's also worth noting that `T` in `Task` is covariant and thus `Saga<any, never>` satisfies `Saga<any, any>`. It's always worth re-checking this because the same is not true for contravariant parameters: [TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgMIHsBucrDuAHnQFcxkAVAPgG8AoZB5MdABSnQFtgBnCALmQAKAJTIAvJWRtOPCASq0AvrVpgAngAcUAQXFosOPIRARM0SRAAekEABNu+7LnxgC+NZID8yAIzIBAAy0APTBjAB6niqgkLCIKADKYMAANikYTkauJGRUdIxMrOxcvAIi4pLSJXIKyqqaKABCekmp6QbOxqbmyFY29sitaRmGLm4gHsjefoEhYQyR0eDQ8Ej64FBwmWOgFDT0jAhwaWVwAuSiEsiY6MC2SiqhTAAWPMhHIMgARiga7F8pCAcODJBD1LRoPQYDZbUZdMxQCzWCB2BzQsCbbaEdxeXz+ZBBJ4LTxAA)

packages/core/types/ts3.6/runSaga.test.ts
```diff
@@ -7,7 +7,8 @@ declare const effect: StrictEffect
 declare const iterator: Iterator<any>
 
 function testRunSaga() {
-  const task0: Task = runSaga<{ foo: string }, { baz: boolean }, () => SagaIterator>(
+  // $ExpectType Task<void>
+  const task0 = runSaga<{ foo: string }, { baz: boolean }, Saga<undefined[], void>>(
```
Andarist, line 11: It feels like you might have wanted to just provide no arguments:
```suggestion
  const task0 = runSaga<{ foo: string }, { baz: boolean }, Saga<[], void>>(
```

Andarist (approved):
LGTM, let's just wait a day or two to see if @neurosnap has any additional thoughts on this one before merging this

packages/core/types/ts3.6/runSaga.test.ts
lourd: Tried that and got this error:

```
ERROR: 11:60  expect  Compile error in typescript@3.7 but not in typescript@3.8.
Fix with a comment '// Minimum TypeScript Version: 3.8' just under the header.
Type 'Saga<[], void>' does not satisfy the constraint 'Saga<any[], any>'.
  Type 'any[]' is not assignable to type '[]'.
    Types of property 'length' are incompatible.
      Type 'number' is not assignable to type '0'.
```

So left as-is.

neurosnap:
@lourd We are ready to merge, we just need a changeset.  Could you please create one for this PR?  Thanks!

lourd:
@neurosnap can you point me towards directions on how to do that? I tried before but couldn't figure it out

neurosnap:
If you run yarn changeset it should walk you through the process.

Also feel free to read more about it here https://github.com/changesets/changesets

PR #2296, opened 2022-05-13
Updates types to eliminate void-emitting channel pitfall

lourd:
| Minor: New Feature?      | 👍 |

I've run into the pitfall a few times recently of making a void-emitting event channel. I decided that it would be great if the types enforced that incompatibility instead of happily accepting the invalid type, hiding the bug until run time.

### DIFF
--- a/.changeset/fresh-rats-switch.md
+++ b/.changeset/fresh-rats-switch.md
@@ -1,0 +1,7 @@
+---
+'@redux-saga/core': minor
+'@redux-saga/types': minor
+'redux-saga': minor
+---
+
+Updates Channel type to eliminate void-emitting pitfall
--- a/packages/core/types/channels.test.ts
+++ b/packages/core/types/channels.test.ts
@@ -59,6 +59,23 @@ function testChannel() {
   c1.flush((messages: Array<{foo: string}> | END) => {});
 
   c1.close();
+
+  // Testing that we can't define channels that pass void or undefined
+  // $ExpectError
+  const voidChannel: Channel<void> = channel();
+  // $ExpectError
+  const voidChannel2 = channel<void>();
+  // $ExpectError
+  const undefinedChannel = channel<undefined>();
+  // $ExpectError
+  channel().put();
+  // $ExpectError
+  channel().put(undefined);
+
+  // Testing that we can pass primitives into channels
+  channel().put(42);
+  channel().put('test');
+  channel().put(true);
 }
 
 function testEventChannel(secs: number) {
@@ -100,6 +117,21 @@ function testEventChannel(secs: number) {
   c1.flush((messages: number[] | END) => {});
 
   c1.close();
+
+  // $ExpectError
+  const c4: EventChannel<void> = eventChannel(() => () => {})
+
+  // $ExpectError
+  const c5 = eventChannel<void>(emit => {
+    emit()
+    return () => {}
+  })
+
+  const c6 = eventChannel(emit => {
+    // $ExpectError
+    emit()
+    return () => {}
+  })
 }
 
 function testMulticastChannel() {
@@ -121,4 +153,11 @@ function testMulticastChannel() {
   c1.flush((messages: Array<{foo: string}> | END) => {});
 
   c1.close();
+
+  // $ExpectError
+  const c3: MulticastChannel<void> = stdChannel()
+  // $ExpectError
+  const c4 = multicastChannel<void>()
+  // $ExpectError
+  const c5 = stdChannel<void>()
 }
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -1,6 +1,6 @@
 // TypeScript Version: 3.2
 import { Action, Middleware } from 'redux'
-import { Saga, Buffer, Channel, END as EndType, Predicate, SagaIterator, Task } from '@redux-saga/types'
+import { Saga, Buffer, Channel, END as EndType, Predicate, SagaIterator, Task, NotUndefined } from '@redux-saga/types'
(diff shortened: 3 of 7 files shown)


Andarist:
I'm unsure about the change because I'm not sure what the mentioned pitfall is. Could you back it up with some examples?

The current types seem to be correct to me as this is a valid situation: [TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgMIAs4hBANgHgBUA+ZAbwChlkw4BrCACgQCMAuZRgWwgGde4AcwgdCASmQBeUgDcA9sAAmYjvKVVkABwCuYbnwHDRK5GsUaYubb3TN2nYJC69RAbQC6E6aYXLVvjQRcOV4mEzMKAF8KCkUIILgoFAQ5EF4wZAR0DgwsHAIzYhisgDodPW0QOJhQCGVi9BLaBkZGL1IySLEgA)

lourd:
Exactly, the types allow you to do what will end up with a runtime exception. In your example `ch.put(undefined)` throws an exception.

lourd:
Do you want an example of that written in addition to yours and the several tests that are a part of this PR?

Andarist:
Ah, I forgot about this runtime check:
https://github.com/redux-saga/redux-saga/blob/2d2214e9ca8949892c0a7a23ceef39fa32d13939/packages/core/src/internal/channel.js#L38

Thanks for clarifying! I think that the original intention was there to avoid allowing `channel.put()` and not necessarily `channel.put(undefined)` but the runtime code would have to differentiate those cases to perform this runtime validation and that's probably not worth the added complexity. 

I've done some testing (`{}` is always confusing to me in TS) and it looks like everything is alright. Could we add some simple types tests to ensure that both `null` and other primitives (let's just check a channel with numbers or smth) work OK? I've skimmed through the existing tests and I don't see any like that.

lourd:
Ok, added tests for covering primitives.

Doing `channel().put(null)` appears to have an error in TypeScript 3.4 — or at least that popped up when I was running the tests locally. Not sure what to do to resolve that.

Andarist:
This is an extracted repro case for the mentioned problem with null: [TS playground](https://www.typescriptlang.org/play?ts=3.3.3#code/C4TwDgpgBAcg9sAqgOwCYQGYEtkVVAXigG8BfKAHymQFcAbOgKEdEigFEYARQkqViAC4oAcgACYgE54aADwC0AZwCGAc2UB6AMIAJAIIwY7ADIB9TlxFRSzHMAiSMygMbQAQjQwYHAHgAqAHwkjFBQWIrsALZgoAAUAJTCAEZwcHQQysghUGA0wLGREIoqqkJQfolQAG5wWKjZwMoA1hAJwn6UUDRomDh42Rh0NIoAFm3lANoAuow2jHYOTq5QWiOZuHT+UBCy9miKsAgo6Ni4qEHEDc2tzknCBUUlZR1UFvGEQTV1lV-1obn5QrFNTPTpvYS-AZDUaxW73LD2SKKdrTMHcd4ET61VA-bHZZx0OCKVq4uqzZjoAnKaRQDDdZzALBwZBQZxrZAbLY7PaoA7wJA9U54AKxJKebySAD8wg8Xl8gUqq3WEE2gQA3Mw2cq6AkAHQA2K0BjxNVAA)

If we take a closer look at it and hover over `channel()` call then we might see:
```ts
function channel<{}>(buffer?: Buffer<{}> | undefined): Channel<{}>
```
Of course, this just "propagates" further and we might see a similar result when hovering over `.put(null)`:
```ts
(method) Channel<{}>.put(message: {} | END): void
```

It acts as if `null` was never part of the constraint. We can even simplify the minal repro case further and repro the same thing with just:
```ts
declare function test<T extends {} | null>(): T
test()
```

My guess is just that behavior in TS around this somehow changed a long time ago - which is interesting, cause this is the exact behavior that we might experience without `strictNullChecks` (without it we won't have a type error though but `null` gets "ellided" from the constraint). IIRC `strictNullChecks` was already a thing back then.

I think that this particular test can just be omitted for those TS@<3.6 tests.

lourd:
Ok, sounds good.

How do you disable the test? The failure's coming from the non-3.6 directory

```
Error: redux-saga/packages/core/types/channels.test.ts:77:17
ERROR: 77:17  expect  Compile error in typescript@3.4 but not in typescript@3.5.
Fix with a comment '// Minimum TypeScript Version: 3.5' just under the header.
Argument of type 'null' is not assignable to parameter of type '{} | END'.

    at redux-saga/node_modules/dtslint/bin/index.js:198:19
    at Generator.next (<anonymous>)
    at fulfilled (/Users/louis/code/redux-saga/node_modules/dtslint/bin/index.js:6:58)
```

I tried putting that minimum version comment at the top of the file but it doesn't seem to have an effect.

Andarist:
Basically, those files are "sibling" files and for the most part, they are the same (and should be kept the same as much as possible):
```
packages/core/types/channels.test.ts
packages/core/types/ts3.6/channels.test.ts
```

The only difference between them is the **minimum** TS version they are tested against. I didn't fully confirm this but I think the non-ts3.6 files are still tested with TS@>=3.6. 

So to remove a "test" you should just remove the discussed assertion from the file in which it has been added. We should add this section of the changes to the `packages/core/types/ts3.6/channels.test.ts` file though (including the `null`-related test).

lourd:
K, done @Andarist

Andarist (approved):
We just need some changesets here. Remember that you are not constrained to creating a single changeset per PR. I would probably add a single one for `@redux-saga/types` here and the second one for `@redux-saga/core`+`redux-saga` (each describing the change from the perspective of the included packages)

PR #2297, opened 2022-05-13
Adds isAborted method and joiners array property to mock task

### LINKED ISSUE #2205: createMockTask does not properly mock task
### Steps to reproduce

Testing a function that uses the `join` operation, create a test that uses `createMockTask` to mock the return of a call to fork.

`join` on mocked tasks that are not complete will result in an exception.

In order to work around this issue, I added `isAborted` and `joiners` to the mock task.

#### Example

https://github.com/celo-org/wallet/pull/508/files#diff-0d5091378c9a747cc2c61efd98db534427cc6a0a649978ace1ea6c471bfcad10R37-R53

### Actual results

Exception with either `isAborted` not being a function, or `Cannot read property 'push' of undefined` (when `joiners` is not included on the mock)

### The Expected results

`createMockTask` returns an object that can be used in place of a Task in tests.

lourd:
| Fixed Issues?            | Fixes #2205 |
| Tests Added + Pass?      | No new tests |

When doing saga integration testing (using [`expectSaga` from redux-saga-test-plan](https://redux-saga-test-plan.jeremyfairbank.com/integration-testing/)), creating a mock task (passed into the saga using the [`fork.fn`](https://redux-saga-test-plan.jeremyfairbank.com/integration-testing/mocking/static-providers.html#:~:text=by%20fn.-,fork.fn,-Match%20fork%20by) static providers) that's passed to a real `join` effect throws an error. Adding `isAborted` and the `joiners` array are necessary for it to not throw an error.

During code review, it was determined that `createMockTask` should have an API more similar to the real task implementation. `setRunning` was deprecated, and trying to change the state of a task more than once will now throw an error.

Also includes new unit tests for covering basic usage, and a small updates to the docs with slightly improved instructions.

### DIFF
--- a/packages/testing-utils/src/index.js
+++ b/packages/testing-utils/src/index.js
@@ -1,5 +1,17 @@
 import { TASK } from '@redux-saga/symbols'
-import { RUNNING, CANCELLED, ABORTED, DONE } from '@redux-saga/core/src/internal/task-status'
+
+// Keep in sync with @redux-saga/core/src/internal/task-status
+const RUNNING = 0
+const CANCELLED = 1
+const ABORTED = 2
+const DONE = 3
+
+const statusToStringMap = {
+  [RUNNING]: 'Running',
+  [CANCELLED]: 'Cancelled',
+  [ABORTED]: 'Aborted',
+  [DONE]: 'Done',
+}
 
 export const cloneableGenerator = generatorFunc => (...args) => {
   const history = []
@@ -19,6 +31,15 @@ export const cloneableGenerator = generatorFunc => (...args) => {
   }
 }
 
+const assertStatusRunning = status => {
+  if (status !== RUNNING) {
+    const str = statusToStringMap[status]
+    throw new Error(
+      `The task is no longer Running, it is ${str}. You can't change the status of a task once it is no longer running.`,
+    )
+  }
+}
+
 export function createMockTask() {
   let status = RUNNING
   let taskResult
@@ -31,19 +52,20 @@ export function createMockTask() {
     isAborted: () => status === ABORTED,
     result: () => taskResult,
     error: () => taskError,
-    cancel: () => {},
+    cancel: () => {
+      assertStatusRunning(status)
+      status = CANCELLED
+    },
     joiners: [],
 
-    setRunning: () => (status = RUNNING),
     setResult: r => {
+      assertStatusRunning(status)
       taskResult = r
       status = DONE
     },
     setError: e => {
+      assertStatusRunning(status)
       taskError = e
-      status = ABORTED
-    },
-    setIsAborted: () => {
       status = ABORTED
     },
   }
--- a/packages/testing-utils/src/index.js
+++ b/packages/testing-utils/src/index.js
@@ -20,15 +20,18 @@ export const cloneableGenerator = generatorFunc => (...args) => {
 
 export function createMockTask() {
   let _isRunning = true
+  let _isAborted = false
   let _result
   let _error
 
   return {
     [TASK]: true,
     isRunning: () => _isRunning,
+    isAborted: () => _isAborted,
     result: () => _result,
     error: () => _error,
     cancel: () => {},
+    joiners: [],
 
     setRunning: b => (_isRunning = b),
     setResult: r => (_result = r),
--- a/packages/testing-utils/src/index.js
+++ b/packages/testing-utils/src/index.js
@@ -1,4 +1,5 @@
 import { TASK } from '@redux-saga/symbols'
+import { RUNNING, CANCELLED, ABORTED, DONE } from '@redux-saga/core/src/internal/task-status'
 
 export const cloneableGenerator = generatorFunc => (...args) => {
   const history = []
@@ -19,22 +20,31 @@ export const cloneableGenerator = generatorFunc => (...args) => {
 }
 
 export function createMockTask() {
-  let _isRunning = true
-  let _isAborted = false
-  let _result
-  let _error
+  let status = RUNNING
+  let taskResult
+  let taskError
 
   return {
     [TASK]: true,
-    isRunning: () => _isRunning,
-    isAborted: () => _isAborted,
-    result: () => _result,
-    error: () => _error,
+    isRunning: () => status === RUNNING,
+    isCancelled: () => status === CANCELLED,
+    isAborted: () => status === ABORTED,
+    result: () => taskResult,
+    error: () => taskError,
     cancel: () => {},
     joiners: [],
 
-    setRunning: b => (_isRunning = b),
-    setResult: r => (_result = r),
-    setError: e => (_error = e),
+    setRunning: () => (status = RUNNING),
+    setResult: r => {
+      taskResult = r
+      status = DONE
+    },
+    setError: e => {
+      taskError = e
+      status = ABORTED
+    },
+    setIsAborted: () => {
+      status = ABORTED
+    },
   }
 }
--- a/packages/testing-utils/src/index.js
+++ b/packages/testing-utils/src/index.js
@@ -58,6 +58,18 @@ export function createMockTask() {
     },
     joiners: [],
 
+    /**
+     * @deprecated Use `setResult`, `setError`, or `cancel` to change the
+     * running status of the mock task.
+     */
+    setRunning: () => {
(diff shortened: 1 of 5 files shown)


packages/testing-utils/src/index.js
```diff
@@ -20,15 +20,18 @@ export const cloneableGenerator = generatorFunc => (...args) => {
...
 export function createMockTask() {
   let _isRunning = true
+  let _isAborted = false
   let _result
   let _error
 
   return {
     [TASK]: true,
     isRunning: () => _isRunning,
+    isAborted: () => _isAborted,
```
Andarist, line 30: This always returns a constant value - `false`. Shouldn't we expose a similar setter as we do for other "statuses" (`setRunning`, `setError`, etc)? I also wonder - some of those can't be true at the same time, so I think that it would make sense to create an internal `status` variable, initialize it to `running'` and then switch it appropriately based on those setters. That way we won't end up with an impossible task with `t.isRunning() === true && t.isAborted() === true`

Andarist: In general, it's fairly easy to inspect how the real task behaves here:
https://github.com/redux-saga/redux-saga/blob/4da7bd91d1548218e600c43592fc14ec83819d5e/packages/core/src/internal/newTask.js#L118-L141

packages/testing-utils/types/index.d.ts
```diff
@@ -107,7 +107,8 @@ export interface SagaIteratorClone extends SagaIterator {
 export function createMockTask(): MockTask
 
 export interface MockTask extends Task {
-  setRunning(running: boolean): void
+  setRunning(): void
```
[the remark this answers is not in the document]
Andarist: We could add a `@deprecated` signature back and handle both at the implementation level - `task.setRunning()` should just be synonymous with `task.setRunning(true)`. I'm not 100% sure what `task.setRunning(false)` should correspond to though... it seems like canceled would fit better here than aborted

packages/testing-utils/types/index.d.ts
```diff
@@ -107,7 +107,8 @@ export interface SagaIteratorClone extends SagaIterator {
...
   setResult(result: any): void
   setError(error: any): void
+  setAborted(): void
```
Andarist, line 113: On the second thought - `setAborted` is somewhat synonymous to `setError(err?: any)`. Thoughts?

packages/testing-utils/src/index.js
```diff
@@ -19,19 +20,31 @@ export const cloneableGenerator = generatorFunc => (...args) => {
...
+    isAborted: () => status === ABORTED,
+    result: () => taskResult,
+    error: () => taskError,
     cancel: () => {},
+    joiners: [],
 
-    setRunning: b => (_isRunning = b),
-    setResult: r => (_result = r),
-    setError: e => (_error = e),
+    setRunning: () => (status = RUNNING),
```
Andarist, line 37: just a random thought without full analysis - perhaps `setRunning` should just be deprecated? it seems to me that making a mock task forcefully take a transition that isn't possible in the real code is asking for trouble.

Valid transitions are those:
- `running` -> `done` (can be achieved by calling `setResult`)
- `running` -> `aborted` (can be done by calling `setError`)
- `running` -> `canceled` (could be done by `cancel`)

Thoughts?

packages/testing-utils/src/index.js
```diff
@@ -1,4 +1,5 @@
 import { TASK } from '@redux-saga/symbols'
+import { RUNNING, CANCELLED, ABORTED, DONE } from '@redux-saga/core/src/internal/task-status'
```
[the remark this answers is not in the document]
Andarist: I don't think we actually need to import those from anywhere, they can just be inlined in this module

packages/testing-utils/types/index.d.ts
lourd: Agreed, I don't think setAborted is needed. The equivalent to the saga behavior API is generator.throw, so having just setError makes sense to me.

packages/testing-utils/types/index.d.ts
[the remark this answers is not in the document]
lourd: For all of the times I've used this in the past, I've always had to do:

```
mockTask.setResult(X)
mockTask.setRunning(false)
```

This is necessary to get the rest of the saga to correctly proceed when joining on a mock task.

I've never had to do `setRunning(true)` because the tasks are already running when created.

I think taking out setRunning altogether is simpler than trying to guess what people would expect. Maybe that means publishing another major version? I think since it's a testing utility if we did it in a minor version it would be acceptable.

packages/testing-utils/src/index.js
```diff
@@ -19,19 +20,31 @@ export const cloneableGenerator = generatorFunc => (...args) => {
...
     [TASK]: true,
-    isRunning: () => _isRunning,
-    result: () => _result,
-    error: () => _error,
+    isRunning: () => status === RUNNING,
+    isCancelled: () => status === CANCELLED,
+    isAborted: () => status === ABORTED,
+    result: () => taskResult,
+    error: () => taskError,
     cancel: () => {},
```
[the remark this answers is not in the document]
lourd: Done.

packages/testing-utils/src/index.js
lourd: I agree, I think that's a good idea.

Went ahead and added checks for that and tests to cover it :)

lourd:
@Andarist addressed all your comments and updated the PR description, PTAL.

packages/testing-utils/types/index.d.ts
```diff
@@ -107,7 +107,6 @@ export interface SagaIteratorClone extends SagaIterator {
 export function createMockTask(): MockTask
 
 export interface MockTask extends Task {
-  setRunning(running: boolean): void
```
Andarist, line 110: I totally agree with this change - I am concerned about this being a breaking change though. I'm not opposed to releasing a new major version of the Redux Saga or accompanying packages but this seems like a minor cleanup that doesn't necessarily require us to immediately bump a major version.

This PR will probably change the behavior of this task mock slightly, one way or another. But removing a known method is a big breaking change (for this package alone). Could we think about something less disruptive here? My preference would be to add this method back and introduce `@deprecated` + `console.warn` for what we want to phase out.

packages/testing-utils/src/index.js
```diff
@@ -18,20 +31,54 @@ export const cloneableGenerator = generatorFunc => (...args) => {
...
+      `The task is no longer Running, it is ${str}. You can't change the status of a task once it is no longer running.`,
+    )
+  }
+}
+
 export function createMockTask() {
-  let _isRunning = true
-  let _result
-  let _error
+  let status = RUNNING
```
Andarist, line 44: nit: in general, it feels that keeping this as a number doesn't buy us anything here since we have to remap this later to string anyway when throwing an error. We could just ditch that if we'd use string status here.

I don't feel strongly about this though - both approaches are fine.

packages/testing-utils/src/index.js
```diff
@@ -18,20 +31,54 @@ export const cloneableGenerator = generatorFunc => (...args) => {
...
+    isRunning: () => status === RUNNING,
+    isCancelled: () => status === CANCELLED,
+    isAborted: () => status === ABORTED,
+    result: () => taskResult,
+    error: () => taskError,
+    cancel: () => {
+      assertStatusRunning(status)
+      status = CANCELLED
+    },
+    joiners: [],
```
Andarist, line 59: I have actually never used this mock task myself - it was here when I got involved in the project and I never bothered to tweak/remove it. I'm also very rusty when it comes to this codebase as I wasn't actively using it for a couple of last years. I wonder though - how useful this mock is in its current shape. 

It seems to me that this now won't throw when passed to a `join` but at the same time the `join` effect will never get resolved. 

I wonder if `setError` works OK when it comes to aborting the parent task etc. Do we have any tests for that? Does it work?

Andarist (approved):
Just a changeset away from landing this :)

packages/testing-utils/src/index.js
lourd: All of the new tests in this PR are pretty comprehensive IMO of describing what behavior works vs what doesn't. If you set the error before it's passed to a join effect it will correctly throw and abort the parent task.

PR #2308, opened 2022-06-20
add package entry points

### LINKED ISSUE #2306: Is there a proper way to import `redux-saga/effects` in ESM?
In an ESM, the traditional `import { xxx } from 'redux-saga/effects'` will throws `ERR_UNSUPPORTED_DIR_IMPORT`.

It seems node just doesn't support nested package & treat it as a directory.

example:

put below in same dir then `npm i` then `node main.js`:

```json
// package.json
{
  "name":"saga-test",
  "type":"module",
  "dependencies":{
    "redux-saga":"^1.1.3"
  }
}
```

```js
// main.js
import * as effects from "redux-saga/effects"

console.log(effects)
```

neurosnap:
| Fixed Issues?            | Fixes #2306 |

This PR adds package entry points for `redux-saga`.  This will support situations where users would like to use `redux-saga` in `deno`.  Please read #2306 for more context.

### DIFF
--- a/packages/core/import-condition-proxy.mjs
+++ b/packages/core/import-condition-proxy.mjs
@@ -1,0 +1,4 @@
+export * from './dist/redux-saga-core.cjs.js'
+import _default from './dist/redux-saga-core.cjs.js'
+
+export default _default.default
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -9,6 +9,19 @@
     "effects",
     "types"
   ],
+  "exports": {
+    ".": {
+      "types": "./index.d.ts",
+      "browser": "./dist/redux-saga-core.esm.js",
+      "default": "./dist/redux-saga-core.cjs.js"
+    },
+    "./effects": {
+      "types": "./effects/effects.d.ts",
+      "browser": "./dist/redux-saga-effects.esm.js",
+      "default": "./dist/redux-saga-effects.cjs.js"
+    },
+    "./package.json": "./package.json"
+  },
   "scripts": {
     "test": "yarn test:types && yarn test:jest",
     "test:jest": "jest",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -7,18 +7,20 @@
   "files": [
     "dist",
     "effects",
-    "types"
+    "types",
+    "import-condition-proxy.mjs"
   ],
   "exports": {
-    ".": {
-      "types": "./index.d.ts",
-      "browser": "./dist/redux-saga-core.esm.js",
-      "default": "./dist/redux-saga-core.cjs.js"
-    },
     "./effects": {
-      "types": "./effects/effects.d.ts",
-      "browser": "./dist/redux-saga-effects.esm.js",
+      "types": "./types/ts3.6/effects.d.ts",
+      "module": "./dist/redux-saga-effects.esm.js",
       "default": "./dist/redux-saga-effects.cjs.js"
+    },
+    ".": {
+      "types": "./types/ts3.6/index.d.ts",
+      "module": "./dist/redux-saga-core.esm.js",
+      "import": "./import-condition-proxy.mjs",
+      "default": "./dist/redux-saga-core.cjs.js"
     },
     "./package.json": "./package.json"
   },
--- a/packages/deferred/import-condition-proxy.mjs
+++ b/packages/deferred/import-condition-proxy.mjs
@@ -1,0 +1,4 @@
+export * from './dist/redux-saga-deferred.cjs.js'
+import _default from './dist/redux-saga-deferred.cjs.js'
+
+export default _default.default
--- a/packages/deferred/package.json
+++ b/packages/deferred/package.json
@@ -7,8 +7,18 @@
   "typings": "./index.d.ts",
   "files": [
     "dist",
-    "*.d.ts"
+    "*.d.ts",
+    "import-condition-proxy.mjs"
   ],
+  "exports": {
+    ".": {
+      "types": "./index.d.ts",
+      "module": "./dist/redux-saga-deferred.esm.js",
+      "import": "./import-condition-proxy.mjs",
+      "default": "./dist/redux-saga-deferred.cjs.js"
+    },
+    "./package.json": "./package.json"
+  },
   "scripts": {
     "clean": "rimraf dist",
     "prebuild": "yarn clean",
--- a/packages/delay-p/import-condition-proxy.mjs
+++ b/packages/delay-p/import-condition-proxy.mjs
@@ -1,0 +1,4 @@
+export * from './dist/redux-saga-delay-p.cjs.js'
+import _default from './dist/redux-saga-delay-p.cjs.js'
+
+export default _default.default
--- a/packages/delay-p/package.json
+++ b/packages/delay-p/package.json
@@ -7,8 +7,18 @@
   "typings": "./index.d.ts",
   "files": [
     "dist",
-    "*.d.ts"
+    "*.d.ts",
+    "import-condition-proxy.mjs"
   ],
+  "exports": {
+    ".": {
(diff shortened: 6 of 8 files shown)


package.json
```diff
@@ -4,6 +4,20 @@
...
+  "exports": {
+    ".": {
+      "types": "./packages/redux-saga/index.d.ts",
+      "browser": "./packages/redux-saga/dist/redux-saga-core.esmodules-browsers.js",
+      "umd": "./packages/redux-saga/dist/redux-saga.umd.js",
+      "require": "./pacages/redux-saga/dist/redux-saga-core-npm-proxy.cjs.js"
+    },
+    "./effects": {
+      "types": "./packages/redux-saga/effects.d.ts",
+      "browser": "./packages/redux-saga/dist/redux-saga-effects.esmodules-browsers.js",
```
israelidanny, line 16: It might also be useful to add a `development` (.js file) and `production` (min.js file) condition here for browsers?

neurosnap: https://nodejs.org/api/packages.html#community-conditions-definitions

It seems reasonable to provide both and we already have the builds so I updated the PR to include them.

israelidanny (approved):

israelidanny (approved):

package.json
neurosnap: That would be great if you could.  I tested esm and cjs via esbuild and everything appeared to work correctly.

israelidanny: It all looks good to me from over here, too!

package.json
```diff
@@ -4,6 +4,26 @@
     "examples/*",
     "packages/*"
   ],
+  "exports": {
```
Andarist, line 7: 1. this doesn't add any value to the root `package.json` as that is only a private manifest for this monorepo. It doesn't have a direct effect on published packages though. This should be added to the `packages/core/package.json`
2. we should also add `package.json` to the list of `exports`
3. there are some typos in those file paths, `pacages` -> `packages`
4. I'm not aware of any tool that would understand the `umd` condition, I think this can be removed
5. the paths that you have used for the `browser` condition could be seen as a minor breaking change - we should continue using the `package.json#module` path for this condition
6. it's best to put `default` at the end of "group" instead of `require`

neurosnap: Woops, nice catch!  I've addressed your changes.  I'm going to test them locally and then report back.

packages/core/package.json
```diff
@@ -9,6 +9,19 @@
     "effects",
     "types"
   ],
+  "exports": {
+    ".": {
+      "types": "./packages/redux-saga/index.d.ts",
```
Andarist, line 14: all of those paths need to be relative to this `package.json` but you have used some paths that are relative to the root of the monorepo. Remember that once we publish a package the monorepo "shell" won't be available - it's just an isolated package

neurosnap: Right right, still getting used to a lerna / workspaces workflow for redux-saga, my apologies for the back and forth.

So I understand correctly, do I also need to update `packages/redux-saga` with entry points?  Right now its package.json file contains:

```
"main": "./dist/redux-saga-core-npm-proxy.cjs.js",
"module": "./dist/redux-saga-core-npm-proxy.esm.js",
"unpkg": "./dist/redux-saga.umd.min.js",
```

neurosnap:
Alright I tested this with cjs and esm and things appear to be working like they did previously.  I think this is ready for another look, @Andarist

israelidanny:
This looks amazing @neurosnap @Andarist

packages/core/package.json
```diff
@@ -9,6 +9,19 @@
     "effects",
     "types"
   ],
+  "exports": {
+    ".": {
+      "types": "./index.d.ts",
```
Andarist, line 14:
```suggestion
      "types": "./types/ts3.6/index.d.ts",
```

packages/core/package.json
```diff
@@ -9,6 +9,19 @@
...
+      "browser": "./dist/redux-saga-core.esm.js",
+      "default": "./dist/redux-saga-core.cjs.js"
+    },
+    "./effects": {
+      "types": "./effects/effects.d.ts",
```
Andarist, line 19:
```suggestion
      "types": "./types/ts3.6/effects.d.ts",
```

Andarist:
@neurosnap I've pushed out some changes here to make sure that this works for all of the packages and with more tools. s always when it comes to those things I'm a little bit nervous about this since it's so easy to get things wrong. I would appreciate it if you could test this with node and some popular bundlers.

neurosnap (commented):
I tested the following build tools:
- deno
- esbuild (esm and cjs)
- webpack (esm)

The only one that didn't work was deno, but that's probably because I couldn't figure out how to import a deno module locally.

```bash
$ cat index.ts
import { put, takeEvery } from "/home/erock/dev/redux-saga/packages/redux-saga/import-condition-proxy.mjs";

console.log(put);
console.log('hello world');
```

```bash
deno run index.ts
error: Uncaught SyntaxError: The requested module './dist/redux-saga-core-npm-proxy.cjs.js' does not provide an export named 'default'
import _default from './dist/redux-saga-core-npm-proxy.cjs.js'
       ^
    at <anonymous> (file:///home/erock/dev/redux-saga/packages/redux-saga/import-condition-proxy.mjs:2:8)
```

Andarist:
When using deno directly (without a bundling CDN in the middle) then you probably should reach for:
```ts
import { put, takeEvery } from "/home/erock/dev/redux-saga/packages/redux-saga/dist/redux-saga-effects-npm-proxy.esm.js";
```

This isn't quite a part of the public API (we don't want to be bound to the internal structure of the package, to how files are laid out etc) - but it should work, for the most part (once you build packages in the monorepo).

I don't think that most people are consuming packages directly from npm when using deno though, so this probably doesn't quite represent the real-life usage anyway. I guess we don't really need to worry about deno all that much right now - if we get reports that it doesn't work then we will learn more about how people are actually trying to use this in deno and we'll be able to fix the issue.

q: have you tested this in node?

packages/core/import-condition-proxy.mjs
```diff
@@ -0,0 +1,4 @@
+export * from './dist/redux-saga-core.cjs.js'
```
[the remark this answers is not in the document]
Andarist: I want to avoid [the dual package hazard](https://nodejs.org/api/packages.html#dual-package-hazard) and this wrapper acts as an ES module wrapper, like the one described [here](https://nodejs.org/api/packages.html#approach-1-use-an-es-module-wrapper)

You can import CJS packages in node. So in theory we could just not provide the `import` condition at all. However, there is a super annoying issue in node.js when it comes to default imports for CJS modules - they always resolve to the whole "namespace" (`module.exports`) and not to the actual `module.exports.default`. So this file just imports the CJS module (as it is allowed) and "adapts" its shape to the one that we actually document in the docs etc.

Andarist:
I've verified that it works in node.js

israelidanny:
@Andarist with deno, having an `import_map.json` that maps the following works for me:
```
"redux-saga": "https://esm.sh/redux-saga@1.2.0",
 "redux-saga/": "https://esm.sh/redux-saga@1.2.0/",
```

I can then easily do all the imports very much like node. The slash maps all the sub-packages like redux-saga /effects

Andarist:
That's good to know! I've seen that Deno is also going to have native support for resolving `node_modules` soon-ish. When that happens we might need to tweak our `package.json#exports` to include a `deno` condition. But for now, let's wait and see how this plays out.

israelidanny:
One really weird thing that happens though is that redux-saga.d.ts is empty:
***@***.***/es2022/redux-saga.d.ts

Somehow the packaging misses that altogether and the `esm.sh` resolver
can't find the index.d.ts types and convert them right.

Andarist:
This might be an issue to be raised with esm.sh service

PR #2309, opened 2022-06-22
support ability for custom effects

### LINKED ISSUE #1238: Custom effects

neurosnap:
| Fixed Issues?            | Fixes #1238  |

The purpose of this PR is to support custom effects.  It employs a similar approach to this PR https://github.com/redux-saga/redux-saga/pull/1524

We allow the end-user to provide the `customEffects` key to the middleware options which allows users to `yield` to effects that they define.  This opens up possibilities for community effects as well as effects we could package in separate libraries (e.g. if we don't want `fetcher` inside our `core` package for this PR https://github.com/redux-saga/redux-saga/pull/2307).

```js
import createSagaMiddleware, { makeEffect, resolvePromise } from 'redux-saga';

function runFetchEffect(env, { url, request }, cb) {
  const controller = new AbortController()
  const signal = controller.signal
  const result = fetch(url, { signal, ...request })
  // this is critical to provide
  cb.cancel = () => controller.abort()
  resolvePromise(result, cb)
}

const FETCH_FX = 'FETCH';
const fetcher = (url, request) => makeEffect(FETCH_FX, { url, request });

const sagaMdw = createSagaMiddleware({
  customEffects: {
    [FETCH_FX]: runFetchEffect,
  },
})

sagaMdw.run(function* example() {
  const resp = yield fetcher('https://api.com');
  const data = yield call([resp, 'json']);
  console.log(data);
})
```

### DIFF
--- a/packages/core/src/internal/proc.js
+++ b/packages/core/src/internal/proc.js
@@ -1,7 +1,6 @@
 import * as is from '@redux-saga/is'
 import { IO, TASK_CANCEL } from '@redux-saga/symbols'
 import { RUNNING, CANCELLED, ABORTED, DONE } from './task-status'
-import effectRunnerMap from './effectRunnerMap'
 import resolvePromise from './resolvePromise'
 import nextEffectId from './uid'
 import { asyncIteratorSymbol, noop, shouldCancel, shouldTerminate } from './utils'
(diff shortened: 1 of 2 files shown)


Andarist:
IIRC this is already implementable using something like this:
```ts
import { CANCEL } from '@redux-saga/symbols'

const fetch = (url) => call(function () {
    const controller = new AbortController();
    const signal = controller.signal;
    const promise = globalThis.fetch(url, { signal });
    promise[CANCEL] = () => controller.abort();
    return promise
});
```

So I wonder how useful it would be to add support for custom effects if almost anything can already be composed using the provided primitives.

neurosnap:
The idea was to assist in building a new library `saga-toolkit` to preserve `redux-saga`.  If I can build custom effects in that library and it functions like the effects defined inside this library, then that works for me.

Andarist:
> The idea was to assist in building a new library saga-toolkit to preserve redux-saga. If I can build custom effects in that library and it functions like the effects defined inside this library, then that works for me.

I think that it should be doable to build such a library based on the existing pieces - if not then we can revive this PR later, based on the found shortcomings.

PR #2332, opened 2022-10-30
use configureStore instead of createStore nowadays

LordMoMA:
| Fixed Issues?           | `Fixes #1, Fixes #2`  |


Andarist:
I think that it would be better to add a second example than change this one. The new example could be put first though

LordMoMA:
Great, do u need my help to add the new version above the all the old ones?
If so, do u have any specific template to follow if they both appear on the
docs.

Thanks.
David

Issue #1805, opened 2019-03-03
Throttle channel


twiggler:
The [documentation](https://redux-saga.js.org/docs/api/#throttlems-channel-saga-args) states that it is possible to pass a channel to `throttle`. 

However, when I do the following:
```
const progressChan = yield channel(buffers.sliding(.));
...
yield throttle(., progressChan, ...)
````

actionChannel borks about an invalid pattern.

restrry:
Sorry, that's mistake in docs. Right now it works only with pattern. Need to support a channel as well
https://github.com/redux-saga/redux-saga/blob/master/packages/core/src/internal/sagaHelpers/throttle.js

twiggler:
No problem.
Should I create a pull request for fixing the mistake in the docs, or are you going to remove it?

restrry:
would be great to try to implement it. ideally all effects should work in the same way

twiggler:
@restrry I will give it a shot

twiggler:
Some thoughts about implementing this:

Throttling actions is currently implemented using an `actionmap`. There is no guarantee when  throttling a channel that the  channel to be throttled is created with a sliding buffer of size 1. One method  which works with any channel buffer is to `take` one and then `flush` the channel.

Unfortunately in this case the `throttle` operation is backed by two possible implementations, depending if the supplied argument is a pattern or channel. If we accept a superfluous `flush` on the actionchannel in case of throttling a pattern, both implementations are almost identical.

neurosnap:
Closing due to inactivity.  I realize this is a feature request but since there hasn’t been any activity in years, I’m going to assume it’s not a hotly desired feature.  If this is a wrong assumption, I’d be happy to reopen and continue to discuss.

rssfrncs:
just ran into this today any good user land workarounds? 👀

neurosnap:
Hi @rssfrncs

Sorry for the late response.  I haven't had time to investigate adding throttling to our channels.  In theory it should be possible.  I don't know of any workarounds at this point but I'll spend some time this weekend looking into it.

I'm going to reopen this issue.

isc30:
Please we need a fix, this is very confusing when facing the bug

neurosnap:
@isc30  I'll add this to the top of my list

isc30:
Hey thanks, I'm also happy to drop a PR sometime tomorrow if that helps

isc30:
[...]

PR #2340, opened 2022-11-16
bug: throttle now accepts a channel as originally intended 

neurosnap:
Closes: #1805

This allows a pattern **or** channel to be passed to `throttle`.  It's a feature we've never had before but it's also something that is "supported" in our docs.

| Fixed Issues?           | #1805  |
| Minor: New Feature?     | Yes                                                                       |

### DIFF
--- a/.changeset/neat-wolves-wink.md
+++ b/.changeset/neat-wolves-wink.md
@@ -1,0 +1,5 @@
+---
+'@redux-saga/core': patch
+---
+
+throttle now accepts a channel as originally intended
--- a/.changeset/neat-wolves-wink.md
+++ b/.changeset/neat-wolves-wink.md
@@ -1,4 +1,5 @@
 ---
+'redux-saga': patch
 '@redux-saga/core': patch
 ---
 
--- a/packages/core/src/internal/sagaHelpers/throttle.js
+++ b/packages/core/src/internal/sagaHelpers/throttle.js
@@ -1,11 +1,11 @@
+import * as is from '@redux-saga/is'
 import fsmIterator, { safeName } from './fsmIterator'
 import { take, fork, actionChannel, delay } from '../io'
 import * as buffers from '../buffers'
 
-export default function throttle(delayLength, pattern, worker, ...args) {
+export default function throttle(delayLength, patternOrChannel, worker, ...args) {
   let action, channel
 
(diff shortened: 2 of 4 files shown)


packages/core/src/internal/sagaHelpers/throttle.js
```diff
@@ -1,34 +1,55 @@
...
 import * as buffers from '../buffers'
 
-export default function throttle(delayLength, pattern, worker, ...args) {
+export default function throttle(delayLength, patternOrChannel, worker, ...args) {
   let action, channel
 
-  const yActionChannel = { done: false, value: actionChannel(pattern, buffers.sliding(1)) }
   const yTake = () => ({ done: false, value: take(channel) })
-  const yFork = ac => ({ done: false, value: fork(worker, ...args, ac) })
+  const yFork = (ac) => ({ done: false, value: fork(worker, ...args, ac) })
```
Andarist, line 10: hm, locally i dont end up with those parens - are you sure that your IDE is using the local version of Prettier?

Andarist (commented):
TS type definitions should also be adjusted here

packages/core/src/internal/sagaHelpers/throttle.js
neurosnap: You're right, I have my IDE setup to autoformat based on my preferences.

neurosnap:
@Andarist as far as I can tell, we always intended on supporting channels for `throttle` as the docs and the typedefs seem to support it: https://github.com/redux-saga/redux-saga/blob/9828e7c0a87c7d388ab0906b77a9b065ef0fb1e7/packages/core/types/ts3.6/effects.d.ts#L1148-L1162

https://github.com/redux-saga/redux-saga/issues/1805

Andarist:
A changeset file should be added here. Its content will directly be included in the generated changelog entry.

.changeset/neat-wolves-wink.md
```diff
@@ -0,0 +1,5 @@
+---
+'@redux-saga/core': patch
```
Andarist, line 2: Since `redux-saga` is a more user-facing package - let's add this to its CHANGELOG too

```suggestion
'redux-saga': patch
'@redux-saga/core': patch
```

PR #2344, opened 2022-12-04
bug: we want to check the actual results of the saga

neurosnap:
And not the expected values.  I came across this error when reading all the specs.

| Fixed Issues?           |  |
| Patch: Bug Fix?         |                Yes                                                        |

### DIFF
--- a/packages/core/__tests__/interpreter/base.js
+++ b/packages/core/__tests__/interpreter/base.js
@@ -143,7 +143,7 @@ test('saga yielded falsy values', () => {
   const task = middleware.run(genFn)
   const expected = [false, undefined, null, '', 0, NaN]
   return task.toPromise().then(() => {
-    expect(isNaN(last(expected))).toBe(true) // saga must inject back yielded falsy values
+    expect(isNaN(last(actual))).toBe(true) // saga must inject back yielded falsy values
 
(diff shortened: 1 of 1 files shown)


packages/core/__tests__/interpreter/base.js
```diff
@@ -143,7 +143,7 @@ test('saga yielded falsy values', () => {
   const task = middleware.run(genFn)
   const expected = [false, undefined, null, '', 0, NaN]
   return task.toPromise().then(() => {
-    expect(isNaN(last(expected))).toBe(true) // saga must inject back yielded falsy values
+    expect(isNaN(last(actual))).toBe(true) // saga must inject back yielded falsy values
```
Andarist, line 146: would the previous version fail if a more correct `Number.isNaN` would be used here?
```suggestion
    expect(Number.isNaN(last(actual))).toBe(true) // saga must inject back yielded falsy values
```

neurosnap: Tests pass all the same with your suggestion using `Number.isNaN(last(expected))`

Andarist: I'm not sure what I was even asking here 🤣 my question didn't make a lot of sense.

PR #2360, opened 2023-02-03
Update effects.d.ts code docs to reflect puts error bubbling behavior

shaunwhyte:
The documentation here doesn't match what the real behavior is. (Also inconsistent with the readme docs)

Fixed the code comment on put which stated that errors from a reducer would not bubble back into the saga where as they actually do. 

The code doc is inconsistent with https://redux-saga.js.org/docs/api/#putaction

Code pen to prove behavior https://codesandbox.io/s/redux-saga-example-put-bubbles-error-vmyvvm

Also came across this tests which, as far as I can tell, proves that errors are bubbled
https://github.com/redux-saga/redux-saga/blob/38ac71b0a5bcbf086bd55f77edf6be12a4078e89/packages/core/__tests__/interpreter/put.js#L113

### DIFF
--- a/packages/core/types/effects.d.ts
+++ b/packages/core/types/effects.d.ts
@@ -359,8 +359,8 @@ export type HelperWorkerParameters<T, Fn extends (...args: any[]) => any> = Last
 
 /**
  * Creates an Effect description that instructs the middleware to dispatch an
- * action to the Store. This effect is non-blocking and any errors that are
- * thrown downstream (e.g. in a reducer) will not bubble back into the saga.
+ * action to the Store. This effect is non-blocking, any errors that are
+ * thrown downstream (e.g. in a reducer) will bubble back into the saga.
  *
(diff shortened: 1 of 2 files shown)


Andarist (approved):
@neurosnap could you also take a look at this? IIRC this is the correct change but it would be great if you could validate this with the runtime/test suite

neurosnap:
This is correct, our `put` effect catches errors from `store.dispatch`.  

https://github.com/redux-saga/redux-saga/blob/38ac71b0a5bcbf086bd55f77edf6be12a4078e89/packages/core/src/internal/effectRunnerMap.js#L72

The `dispatch` function in `redux` is what calls the root reducer:

https://github.com/reduxjs/redux/blob/ece840f7e5030ed460894f97a63196baadd9194b/src/createStore.ts#L274

So when a user calls `yield put()` we will catch any errors emitted from `store.dispatch` which includes errors inside the reducer.

PR #2382, opened 2023-06-08
Add leading and trailing options for debounce helper #2380

### LINKED ISSUE #2380: could we add leading/trailing edge options for debounce?
This is more of a feature request than a bug. If you think this feature might make sense then maybe I will raise a PR for it.

Could we add options for debounce to configure whether the saga is called on the leading edge and/or the trailing edge? (Compare to the [lodash debounce function](https://lodash.com/docs/4.17.15#debounce)).

This would be a very useful feature because by setting `leading: true` I can ensure I respond immediately to user actions without having to wait for the debounce period to expire.

ensconced:
| Fixed Issues?           | Fixes #2380|
| Patch: Bug Fix?         |                     No                                                   |
| Major: Breaking Change? |        No                                                                 |
| Minor: New Feature?     |    Yes                                                                   |
| Any Dependency Changes? |   No                                                                     |

This adds `leading` and `trailing` options to the `debounce` helper, as suggested in the linked issue. The first parameter to `debounce` can now be an options object of format `{ delayLength: number; leading?: boolean; trailing?: boolean }`, or for backwards compatibility can still just be a `number` specifying `delayLength`.

`leading` defaults to `false` and `trailing` defaults to `true`. The behaviour in this case should be unchanged compared to on `main`.

The behaviour of "leading" and "trailing" should essentially match how it works in [lodash](https://lodash.com/docs/4.17.15#debounce).

### DIFF
--- a/packages/core/src/internal/sagaHelpers/debounce.js
+++ b/packages/core/src/internal/sagaHelpers/debounce.js
@@ -1,8 +1,16 @@
 import fsmIterator, { safeName } from './fsmIterator'
 import { delay, fork, race, take } from '../io'
 
-export default function debounceHelper(delayLength, patternOrChannel, worker, ...args) {
-  let action, raceOutput
+export default function debounceHelper(delayLengthOrOptions, patternOrChannel, worker, ...args) {
+  let action, raceOutput, delayLength, leading, trailing
+
+  if (typeof delayLengthOrOptions === 'number') {
+    delayLength = delayLengthOrOptions
+    leading = false
+    trailing = true
+  } else {
+    ;({ delayLength, leading = false, trailing = true } = delayLengthOrOptions)
+  }
 
   const yTake = { done: false, value: take(patternOrChannel) }
   const yRace = {
@@ -16,6 +24,7 @@ export default function debounceHelper(delayLength, patternOrChannel, worker, ..
   const yNoop = (value) => ({ done: false, value })
 
   const setAction = (ac) => (action = ac)
+  const unsetAction = () => (action = undefined)
   const setRaceOutput = (ro) => (raceOutput = ro)
 
   return fsmIterator(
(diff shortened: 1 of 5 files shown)


packages/core/src/internal/sagaHelpers/debounce.js
```diff
@@ -1,8 +1,16 @@
 import fsmIterator, { safeName } from './fsmIterator'
 import { delay, fork, race, take } from '../io'
 
-export default function debounceHelper(delayLength, patternOrChannel, worker, ...args) {
-  let action, raceOutput
+export default function debounceHelper(delayLengthOrOptions, patternOrChannel, worker, ...args) {
```
neurosnap, line 4: What happens if both `leading` and `trailing` are set to `true`?

neurosnap (commented):
Thanks for submitting this PR!  I had one comment about the API and then we also need to ensure our types are updated to reflect this change.

See here for a starting place: https://github.com/redux-saga/redux-saga/blob/main/packages/core/types/ts3.6/effects.d.ts#L1220

packages/core/src/internal/sagaHelpers/debounce.js
ensconced: The saga will run at the start of the wait period, and if any further actions occur during the wait period, it will run again at the end of the wait period. This matches the behavior of the lodash debounce function with `leading: true, trailing: true`. I imagine this will be a common use-case - it's how I will use it personally.

There is an example of this in the tests.

ensconced: I have deployed a little example here https://heartfelt-semifreddo-a4737f.netlify.app/
(Hit any key on the keyboard to see the effect of a 1000ms debounce with the various options).

Source code is [here](https://github.com/ensconced/redux-saga-trailing-leading-demo/blob/master/index.js)

ensconced:
@neurosnap I just wanted to gently bump this - do you have any concerns about this change? Or are there any further changes required?

neurosnap:
Thanks for the bump.  I haven't forgotten about this PR, just need to make time to fully review and test it.  I'll try to get this reviewed by end of week.

packages/core/src/internal/sagaHelpers/debounce.js
```diff
@@ -1,8 +1,16 @@
...
+  let action, raceOutput, delayLength, leading, trailing
+
+  if (typeof delayLengthOrOptions === 'number') {
+    delayLength = delayLengthOrOptions
+    leading = false
+    trailing = true
+  } else {
+    ;({ delayLength, leading = false, trailing = true } = delayLengthOrOptions)
```
neurosnap, line 12: This syntax reads a little awkward, could we rework this?

neurosnap (changes requested):
We still need types updated for https://github.com/redux-saga/redux-saga/blob/main/packages/core/types/ts3.6/effects.d.ts#L1220

Issue #2396, opened 2023-10-13
How to use package that use redux-saga as dependency when its in webpack externals?


krutoo:
hi, I'm having difficulty building a project with redux-saga for Node.js via webpack

in the project I use my own package that uses redux-saga as dependency

in the webpack I specified that my package should be included in the bundle via webpack-node-externals allowList

after launching the application in Node.js it crashes with an error exactly in the place where my package uses redux-saga

at this point error is: createSagaMiddleware is not a function

I see that CJS and ESM versions of redux-saga is different

krutoo:
I have similar question in redux toolkit: https://github.com/reduxjs/redux-toolkit/discussions/3756

Can we provide named export of createSagaMiddleware in redux-saga package?

neurosnap:
I have no issues exporting `createSagaMiddleware` as a named export and will happily accept a PR for it.

Andarist:
> external code uses require('redux-saga') that use exports object as function instead its "default" property

How is this possible? You can't do this in node today with our package. You have to use `.default` to access this function

krutoo:
> > external code uses require('redux-saga') that use exports object as function instead its "default" property
> 
> How is this possible? You can't do this in node today with our package. You have to use `.default` to access this function

In project you can see that happens

I think webpack generates bundle that uses require() bun based on ESM version of my package that uses redux-saga as dependency

you can see build/main.js

krutoo:
I also ask that question in webpack but at this moment there is no answers

https://github.com/webpack/webpack/discussions/17743

krutoo:
> I have no issues exporting `createSagaMiddleware` as a named export and will happily accept a PR for it.

@neurosnap @Andarist Should i make PR for it?

krutoo:
@neurosnap I make PR can you review?

https://github.com/redux-saga/redux-saga/pull/2399

PR #2399, opened 2023-10-26
Named export createSagaMiddleware added to @redux-saga/core

krutoo:
- named export `createSagaMiddleware` added to `@redux-saga/core`

According to #2396

| Q                       | A                                      |
| Fixed Issues?           | 
| Minor: New Feature?     | Yes                                                                       |

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -1,5 +1,5 @@
 export { CANCEL, SAGA_LOCATION } from '@redux-saga/symbols'
-export { default } from './internal/middleware'
+export { default, createSagaMiddleware } from './internal/middleware'
 
 export { runSaga } from './internal/runSaga'
 export { END, isEnd, eventChannel, channel, multicastChannel, stdChannel } from './internal/channel'
--- a/packages/core/src/internal/middleware.js
+++ b/packages/core/src/internal/middleware.js
@@ -3,7 +3,7 @@ import { check, assignWithSymbols, createSetContextWarning } from './utils'
 import { stdChannel } from './channel'
 import { runSaga } from './runSaga'
 
-export default function sagaMiddlewareFactory({ context = {}, channel = stdChannel(), sagaMonitor, ...options } = {}) {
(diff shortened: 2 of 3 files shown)


krutoo:
@Andarist there is some new details about the problem

https://github.com/webpack/webpack/discussions/17743#discussioncomment-7272647

Andarist:
The resolution I'd like to take here is to migrate our build setup to https://github.com/preconstruct/preconstruct/ . I have part of the work done on some old branch but never got to finishing it. I believe that would resolve the issue since it would output dist files in a way that they would be compatible with a wide range of tools

krutoo:
> The resolution I'd like to take here is to migrate our build setup to https://github.com/preconstruct/preconstruct/ . I have part of the work done on some old branch but never got to finishing it. I believe that would resolve the issue since it would output dist files in a way that they would be compatible with a wide range of tools

@Andarist thanks a lot, but can you also add named export for createSagaMiddleware?

cristian-eriomenco:
Can we take babysteps and merge this PR first? Its causing quite a bit of headache on rollup, side of things:
```
import createSagaMiddleware, { SagaMiddleware } from 'redux-saga';
...
const sagaMiddleware = createSagaMiddleware();
```

for cjs gets translated into:

```
var createSagaMiddleware = require('redux-saga');
...
const sagaMiddleware = createSagaMiddleware();
```

which lacks the `.default()`

A named export would be very handy, expecially with the VSCode's auto-import plugins which don't work with default exports.

krutoo:
@cristian-eriomenco I would also really like to add named exports, it is convenient and will help solve assembly problems in our case

shaunvdberg:
Hey @krutoo @Andarist 

Any update on the above?

krutoo:
@shaunvdberg @cristian-eriomenco @Andarist I still think the fastest and easiest solution would be to add a named export of `createSagaMiddleware` because named exports work fine in all conditions

anandfresh (approved):

krutoo:
@Andarist can we merge and publish it or make package dual esm+cjs? We have still a problem with using redux-saga trough other npm package

krutoo:
bump @Andarist

PR #2405, opened 2023-12-25
refactor scheduler.js

swchen:
| Fixed Issues?           |  |

refactor and make `redux-saga/packages/core/src/internal/scheduler.js` easier to understand.


Andarist:
Honestly, I'm rather hesitant to take this in. The previous implementation is quite battle-tested and I don't have the mental energy to compare it against the new one to assess if their behaviors are identical.

neurosnap:
Thanks for the contribution!  I agree with @Andarist this is not something we want to refactor at this point in time.  Closing.

PR #2707, opened 2025-02-26
Avoid saga getting stuck when the callstack overflows

philippotto:
| Fixed Issues?           | contributes to #1592 |
| Patch: Bug Fix?         |  Yes                                                                      |
| Major: Breaking Change? |  No                                                                      |
| Minor: New Feature?     | No                                                                       |
| Any Dependency Changes? |  No                                                                      |

## Problem

As described in #1592 (which was opened ~6.5 years ago), repeated CALL effects can lead to a silently stuck saga state. Here's a short example for a saga that would get stuck silently:

```js
function* sagaThatWillGetStuck() {
  for (let i = 0; i < 1500; i++) {
    yield io.call(someOtherSaga)
  }
}
```

The problem is that repeated CALL effects are processed in a recursive manner which can lead to an overflow of the call stack. Fixing this properly would likely require switching from recursion to iteration for the effect processing (as proposed [here](https://github.com/redux-saga/redux-saga/issues/1592#issuecomment-419857204)), but attempts to do so haven't been successful yet (and weren't prioritized probably).

## Impact

In my opinion, the bug described in #1592 is quite significant. As a library user, running into this bug can happen easily and finding it is very hard since there is no actual error reported. I use redux-saga quite heavily and this is the second time this bug has bitten me. Working around the issue is often simple (even though, not ideal), but finding it actually can be very hard.

## Fix

My PR intends to fix the "silent" and "stuck" issue when a stack overflow occurs. It does NOT fix that the stack overflow will occur. Still, this is an important step into the right direction in my opinion.

The fix works by throwing an error when an effect that was already settled as successful is tried to be settled again as unsuccessful. I'm not sure whether this impacts other scenarios, too (and whether they might be valid?), but my impression is that this should not happen, anyway. Here's why this scenario is triggered in the stackoverflow case:

- at some point, `runCallEffect` cannot call `fn` because the call stack overflows
- the catch-clause is entered. however, `cb(error, true)` won't succeed either, because the stack will overflow once again
- now, the catch-clauses of the parent invocations will also be triggered. similarily, some of these will also crash due to a stack overflow.
- at some point, the call stack has shrunken a bit so that a catch-clause in `runCallEffect` can successfully call `cb(error, true)` for some effect. however, that effect was already settled before by a previous `cb(result)` call.
- since an effect cannot be settled twice, the saga got stuck. therefore, throwing an error in this case "properly" escalates the error.

I hope this makes sense. It's a bit hard to explain, but feel free to step through with a debugger (however, note that some debuggers may crash when the stack overflow is reached).

## Next steps
The current state of the PR is my suggestion for how to mitigate the actual problem a bit. I'm open to suggestions and happy to discuss this. I'm very eager to improve the situation, since I rely heavily on redux-saga and the current state seems a bit brittle, to be honest.

### DIFF
--- a/packages/core/src/internal/effectRunnerMap.js
+++ b/packages/core/src/internal/effectRunnerMap.js
@@ -122,6 +122,11 @@ function runCallEffect(env, { context, fn, args }, cb, { task }) {
 
     cb(result)
   } catch (error) {
+    // Remove again?
+    if (env._eventStackDepth > 1000) {
+      /* eslint-disable no-console */
+      console.warn('More than 1000 effects are currently processed recursively.')
+    }
     cb(error, true)
   }
 }
--- a/packages/core/__tests__/interpreter/stackoverflow.js
+++ b/packages/core/__tests__/interpreter/stackoverflow.js
@@ -1,0 +1,38 @@
+import { createStore, applyMiddleware } from 'redux'
+import sagaMiddleware from '../../src'
+import * as io from '../../src/effects'
+
+test('saga escalates a stack overflow to the parent saga', () => {
+  const callAttempts = 1500
+  let actualCallCount = 0
+  let didCatchInWrapper = false
+
+  const rootReducer = () => ({})
+
+  const middleware = sagaMiddleware()
+  createStore(rootReducer, {}, applyMiddleware(middleware))
+
+  function genFnChild() {
+    actualCallCount++
+  }
+
+  function* genFnParent() {
+    for (let i = 0; i < callAttempts; i++) {
+      yield io.call(genFnChild)
+    }
+  }
+
+  function* genWrapper() {
+    try {
+      yield io.call(genFnParent)
+    } catch (exc) {
+      didCatchInWrapper = true
+    }
+  }
+
+  const task = middleware.run(genWrapper)
+  return task.toPromise().then(() => {
+    expect(actualCallCount).toBeLessThan(callAttempts)
+    expect(didCatchInWrapper).toBe(true)
+  })
+})
--- a/packages/core/src/internal/proc.js
+++ b/packages/core/src/internal/proc.js
@@ -102,7 +102,9 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
       }
 
       if (!result.done) {
+        env._eventStackDepth++
         digestEffect(result.value, parentEffectId, next)
+        env._eventStackDepth--
       } else {
         /**
           This Generator has ended, terminate the main task and notify the fork queue
@@ -162,10 +164,16 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
       And We can't complete an already cancelled effectId
     **/
     let effectSettled
+    let didEffectError
 
     // Completion callback passed to the appropriate effect runner
     function currCb(res, isErr) {
       if (effectSettled) {
+        if (!didEffectError && isErr) {
+          throw new Error(
+            'An already successfully settled effect was tried to be settled again as crashed. This might be due to a stack overflow.',
+          )
+        }
         return
       }
 
@@ -181,6 +189,9 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
 
       if (isErr) {
         sagaError.setCrashedEffect(effect)
+        didEffectError = true
+      } else {
+        didEffectError = false
       }
 
       cb(res, isErr)
--- a/packages/core/src/internal/runSaga.js
+++ b/packages/core/src/internal/runSaga.js
@@ -75,6 +75,7 @@ export function runSaga(
     sagaMonitor,
     onError,
     finalizeRunEffect,
+    _eventStackDepth: 0,
   }
 
   return immediately(() => {


packages/core/src/internal/effectRunnerMap.js
```diff
@@ -122,6 +122,11 @@ function runCallEffect(env, { context, fn, args }, cb, { task }) {
 
     cb(result)
   } catch (error) {
+    // Remove again?
+    if (env._eventStackDepth > 1000) {
+      /* eslint-disable no-console */
+      console.warn('More than 1000 effects are currently processed recursively.')
+    }
```
philippotto, line 129: Theoretically, one could use the added `_eventStackDepth` property to warn proactively, but it would only be a heuristic (it highly depends on what the callstack looks like exactly and there is no linear correlation between the callstack size and `_eventStackDepth`).
Therefore, I would remove the code related to `_eventStackDepth` again, but I wanted to leave it here as food for thought for now.

philippotto:
@Andarist @mshustov Sorry for pinging you. I'm sure you have a lot on your plate, but maybe you have a few minutes look at this small PR? I think, it would be quite beneficial to merge it.

philippotto:
It's been two months since my latest ping. Sorry to annoy again, but I really think the PR would increase the robustness of redux-saga and should be merged in some form. @Andarist @mshustov

philippotto:
I'm trying my luck again @Andarist @mshustov

Andarist:
I have a better working fix for this locally - but I need to add more tests to it and I'm currently occupied with some other stuff. I plan to release it as soon as I can switch back to working on it. You could try it in your project and report back if you run into any issues with it:
```diff
diff --git a/packages/core/src/internal/proc.js b/packages/core/src/internal/proc.js
index dafbf59c..6446580a 100644
--- a/packages/core/src/internal/proc.js
+++ b/packages/core/src/internal/proc.js
@@ -7,6 +7,7 @@ import resolvePromise from './resolvePromise'
 import nextEffectId from './uid'
 import { asyncIteratorSymbol, noop, shouldCancel, shouldTerminate } from './utils'
 import newTask from './newTask'
+import { semaphore } from './scheduler'
 import * as sagaError from './sagaError'
 
 export default function proc(env, iterator, parentContext, parentEffectId, meta, isRoot, cont) {
@@ -55,6 +56,12 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
     cont.cancel = task.cancel
   }
 
+  let digesting = false
+  let digestingSemaphore = 0
+  let syncNextCall = false
+  let syncNextArg
+  let syncNextIsErr = false
+
   // kicks up the generator
   next()
 
@@ -71,47 +78,80 @@ export default function proc(env, iterator, parentContext, parentEffectId, meta,
    * receives either (command | effect result, false) or (any thrown thing, true)
    */
   function next(arg, isErr) {
+    if (digesting && (digestingSemaphore === semaphore || digestingSemaphore > 0)) {
+      syncNextCall = true
+      syncNextArg = arg
+      syncNextIsErr = isErr
+      return
+    }
     try {
       let result
-      if (isErr) {
-        result = iterator.throw(arg)
-        // user handled the error, we can clear bookkept values
-        sagaError.clear()
-      } else if (shouldCancel(arg)) {
-        /**
+
+      // eslint-disable-next-line no-constant-condition
+      while (true) {
+        if (isErr) {
+          result = iterator.throw(arg)
+          // user handled the error, we can clear bookkept values
+          sagaError.clear()
+        } else if (shouldCancel(arg)) {
+          /**
           getting TASK_CANCEL automatically cancels the main task
           We can get this value here
 
           - By cancelling the parent task manually
           - By joining a Cancelled task
         **/
-        mainTask.status = CANCELLED
-        /**
+          mainTask.status = CANCELLED
+          /**
           Cancels the current effect; this will propagate the cancellation down to any called tasks
         **/
-        next.cancel()
-        /**
+          next.cancel()
+          /**
           If this Generator has a `return` method then invokes it
           This will jump to the finally block
         **/
-        result = is.func(iterator.return) ? iterator.return(TASK_CANCEL) : { done: true, value: TASK_CANCEL }
-      } else if (shouldTerminate(arg)) {
-        // We get TERMINATE flag, i.e. by taking from a channel that ended using `take` (and not `takem` used to trap End of channels)
-        result = is.func(iterator.return) ? iterator.return() : { done: true }
-      } else {
-        result = iterator.next(arg)
-      }
+          result = is.func(iterator.return) ? iterator.return(TASK_CANCEL) : { done: true, value: TASK_CANCEL }
+        } else if (shouldTerminate(arg)) {
+          // We get TERMINATE flag, i.e. by taking from a channel that ended using `take` (and not `takem` used to trap End of channels)
+          result = is.func(iterator.return) ? iterator.return() : { done: true }
+        } else {
+          result = iterator.next(arg)
+        }
 
-      if (!result.done) {
-        digestEffect(result.value, parentEffectId, next)
-      } else {
-        /**
+        if (!result.done) {
+          let prevDigestingSemaphore = digestingSemaphore
+          try {
+            digesting = true
+            digestingSemaphore = semaphore
+            digestEffect(result.value, parentEffectId, next)
+          } finally {
+            digesting = false
+            digestingSemaphore = prevDigestingSemaphore
+          }
+
+          const syncCalled = syncNextCall
+          syncNextCall = false
+
+          if (!syncCalled) {
+            break
+          }
+
+          arg = syncNextArg
+          isErr = syncNextIsErr
+
+          syncNextCall = false
+          syncNextArg = undefined
+          syncNextIsErr = false
+        } else {
+          /**
           This Generator has ended, terminate the main task and notify the fork queue
         **/
-        if (mainTask.status !== CANCELLED) {
-          mainTask.status = DONE
+          if (mainTask.status !== CANCELLED) {
+            mainTask.status = DONE
+          }
+          mainTask.cont(result.value)
+          break
         }
-        mainTask.cont(result.value)
       }
     } catch (error) {
       if (mainTask.status === CANCELLED) {
diff --git a/packages/core/src/internal/scheduler.js b/packages/core/src/internal/scheduler.js
index 19f93855..cecde090 100644
--- a/packages/core/src/internal/scheduler.js
+++ b/packages/core/src/internal/scheduler.js
@@ -6,7 +6,7 @@ const queue = []
   - Decrementing releases a lock. Zero locks puts the scheduler in a `released` state. This
     triggers flushing the queued tasks.
 **/
-let semaphore = 0
+export let semaphore = 0
 
 /**
   Executes a task 'atomically'. Tasks scheduled during this execution will be queued
```

Andarist:
superseded by https://github.com/redux-saga/redux-saga/pull/2727

PR #2721, opened 2025-10-27
Bring back UMD build

### LINKED ISSUE #2720: No more UMD min file on v1.4.2
### Steps to reproduce

- Upgrade to redux-saga 1.4.2
- Check that there are no more min UMD file downloaded on node_modules

When upgrading to redux-saga 1.4.2 there are no more UMD min file on node_modules dist

Image

Andarist:
fixes https://github.com/redux-saga/redux-saga/issues/2720

### DIFF
--- a/packages/redux-saga/package.json
+++ b/packages/redux-saga/package.json
@@ -53,6 +53,7 @@
   },
   "homepage": "https://redux-saga.js.org/",
   "dependencies": {
+    "@babel/runtime": "^7.28.4",
     "@redux-saga/core": "^1.4.2"
   },
   "devDependencies": {
--- a/packages/redux-saga/package.json
+++ b/packages/redux-saga/package.json
@@ -4,7 +4,7 @@
   "description": "Saga middleware for Redux to handle Side Effects",
   "main": "dist/redux-saga.cjs.js",
   "module": "dist/redux-saga.esm.js",
-  "unpkg": "./dist/redux-saga.umd.min.js",
+  "umd:main": "dist/redux-saga.umd.min.js",
   "files": [
     "dist",
     "effects"
@@ -75,6 +75,7 @@
     "entrypoints": [
       "./index.js",
       "./effects.js"
-    ]
+    ],
+    "umdName": "ReduxSaga"
   }
 }
--- a/patches/@preconstruct+cli+2.8.13.patch
+++ b/patches/@preconstruct+cli+2.8.13.patch
@@ -1,8 +1,30 @@
 diff --git a/node_modules/@preconstruct/cli/cli/dist/preconstruct-cli-cli.cjs.js b/node_modules/@preconstruct/cli/cli/dist/preconstruct-cli-cli.cjs.js
-index b5d4350..42ab524 100644
+index b5d4350..2df24f1 100644
 --- a/node_modules/@preconstruct/cli/cli/dist/preconstruct-cli-cli.cjs.js
 +++ b/node_modules/@preconstruct/cli/cli/dist/preconstruct-cli-cli.cjs.js
-@@ -2772,7 +2772,8 @@ function getRollupConfigs(pkg) {
+@@ -2184,16 +2184,22 @@ let rollupPluginBabel = ({
+   return {
+     name: "babel",
+-    resolveId(id, parent) {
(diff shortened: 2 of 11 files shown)


guilhermewm:
@Andarist any news on this one?

DenVilden:
@Andarist ping

yelouafi:
hey — quick heads up, this doesn’t build on `5e18069`.

`yarn build` fails with:

```
ENOTDIR: packages/core/src/index.js/effects
```

looks like the UMD build for the `effects` entrypoint is the issue — preconstruct resolves `@redux-saga/core` to `index.js` and then incorrectly appends `/effects`.

i tried doing UMD only on the root, but preconstruct requires it on all entrypoints or none.

i’ll open a follow-up PR with a workaround

packages/redux-saga/package.json
```diff
@@ -53,6 +53,7 @@
   },
   "homepage": "https://redux-saga.js.org/",
   "dependencies": {
+    "@babel/runtime": "^7.28.4",
```
Andarist, line 56: this might be redundant now, it's worth checking out if we could remove this explicit dep line from here

patches/@preconstruct+cli+2.8.13.patch
```diff
@@ -0,0 +1,37 @@
...
+   };
+   return {
+     name: "babel",
+-    resolveId(id, parent) {
++    async resolveId(id, parent) {
+       const currentIsBabelHelper = id.startsWith(babelHelpersModuleStart);
+       if (!currentIsBabelHelper) {
+         if (parent && parent.startsWith(babelHelpersModuleStart)) {
+-          return resolveIdForBabelHelper(id);
++          return this.resolve(resolveIdForBabelHelper(id), parent, {
```
Andarist, line 15: I'll have to prepare another PR to Preconstruct to fix this one in there ;p

packages/redux-saga/package.json
yelouafi: tested locally yarn build fails, should probably keep for now

```text
redux-saga "rollupPluginBabelHelpers/extends" is imported by "../core/src/internal/io.js"
but the package is not specified in dependencies or peerDependencies
```

Andarist: that's probably fixable too - but more complicated and not worth it imho as it's only relevant to `@babel/helpers` usage in UMD builds that load other in-monorepo modules. Eh, bundling is *fun*

packages/redux-saga/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Saga middleware for Redux to handle Side Effects",
   "main": "dist/redux-saga.cjs.js",
   "module": "dist/redux-saga.esm.js",
-  "unpkg": "./dist/redux-saga.umd.min.js",
```
yelouafi, line 7: not sure if it still matters but just to confirm losing the unpkg field is intentional (for https://unpkg.com/redux-saga)

scripts/combine-umd-builds.js
Andarist, line 1: this is quirky... but it maintains the old shape of the UMD entrypoint: https://github.com/redux-saga/redux-saga/pull/2721/changes#diff-8522e3f434257cacc921a01a9a4c0e5fcdfca81f6fd3f89e6b685db6bbaac2b2

🤷‍♂️

packages/redux-saga/package.json
yelouafi: haha srry I dragged you into this 😅
I think for a legacy issue keep it simple no need to go for the cleanest solution as long as it works

packages/redux-saga/package.json
Andarist: It's semi-intentional. I thiiiink `unpkg` should be able to resolve using more general-purpose `umd:main` but we can recheck this after deployment and add yet another field into this package.json if it doesnt work. But then, it would be a shame if we'd have to add this for every other CDN~.

yelouafi: yeah I think you're right; we're not missing much from #2720 and if necessary direct url can be used; so this shouldn't be a blocker

yelouafi (approved):
looks good to me 👍; I think hacking our way here is fine as long as long as it works

PR #2725, opened 2026-05-05
[codex] Restore legacy UMD builds

yelouafi:
## Summary

Restores the legacy UMD artifacts for `redux-saga` and `redux-saga/effects` without switching to Preconstruct's UMD pipeline.

The build now keeps `preconstruct build` for the package outputs and runs a small Rollup sidecar script afterward to emit:

- `dist/redux-saga.umd.js`
- `dist/redux-saga.umd.min.js`
- `effects/dist/redux-saga-effects.umd.js`
- `effects/dist/redux-saga-effects.umd.min.js`
- sourcemaps for each artifact

## Why

`packages/redux-saga/package.json` still points `unpkg` at `./dist/redux-saga.umd.min.js`, but the current Preconstruct build no longer emits that file. This makes CDN/browser consumers miss the expected UMD bundle.

This approach preserves the old root UMD shape, including `ReduxSaga.effects`, while keeping the existing Preconstruct build setup intact.

## Notes

- Adds `terser` as a direct root devDependency so the UMD build script is not relying on Preconstruct's transitive dependency.
- Leaves the existing `unpkg` field unchanged.
- Uses the repo's existing Rollup 1 / `rollup-plugin-*` dependency family.

## Validation

- `yarn build`
- UMD artifact existence check
- `npm pack --dry-run --json ./packages/redux-saga`
- UMD smoke check for root and effects bundles
- `yarn prettier --check package.json scripts/build-redux-saga-umd.js .changeset/legacy-umd-builds.md`
- `yarn test`


scripts/build-redux-saga-umd.js
```diff
@@ -0,0 +1,115 @@
+const fs = require('fs').promises
```
Andarist: nit:
```suggestion
const fs = require('fs/promises')
```

package.json
```diff
@@ -9,7 +9,7 @@
     "format": "prettier --write \"{packages,website,examples}/**/*.{js,jsx}\"",
     "lint": "eslint packages/*/{src,test}",
     "test": "jest",
-    "build": "preconstruct build",
+    "build": "preconstruct build && node scripts/build-redux-saga-umd.js",
```
Andarist, line 12: note that preconstruct technically should be able to output those - but it has a bug in aliasing stuff that I didn't get to fixing 😬 A custom script is fine though as it delivers what is needed here

.changeset/legacy-umd-builds.md
Andarist, line 1: Note that you might run into some issues when trying to actually publish this right now. I'll prepare some workflow changes + set up trusted publishing

package.json
yelouafi: Codex actually went that way and patched preconstruct but I steered for a simpler solution

Andarist: I'm working on a fix in Preconstruct for this aliasing stuff - but I'm not sure if this will resolve the whole issue, let's see

Andarist: https://github.com/preconstruct/preconstruct/pull/624

yelouafi: nice!

yelouafi:
superseded by #2721

PR #2726, opened 2026-05-05
Setup trusted publishing

Andarist:


.github/workflows/publish.yml
```diff
@@ -0,0 +1,66 @@
...
+      id-token: write # to use OpenID Connect token for trusted publishing (changesets/action)
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Use Node.js 24
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 24
+          cache: yarn
+
```
yelouafi, line 58: dont we need to install deps & build here as well?

Andarist: good catch, those steps were not set up good between both jobs - I just fixed this.

yelouafi:
a potential issue, package manifests currently use GitHub tree URLs (`"repository": "https://github.com/redux-saga/redux-saga/tree/main/packages/core"`); [npm docs](https://docs.npmjs.com/trusted-publishers#troubleshooting) say 

> To publish from GitHub, your package's repository.url field in package.json must exactly match your GitHub repository


should we switch to repository url +directory (https://docs.npmjs.com/cli/v11/configuring-npm/package-json#repository)

Andarist:
I'm using this `package.json#repository` flavor in Changesets itself and it works fine: https://github.com/changesets/changesets/blob/6c250f58a128350c6905ed16691a25bf0c8bae0c/packages/cli/package.json#L8

That said, on the `next` branch we are going to use the flavor u suggest soon (see [here](https://github.com/beeequeue/changesets/blob/18b426af3bc0e5c3fe5aa5fdc3a78f397db5ec6e/packages/cli/package.json#L8-L12)) and that's supposed to work too.

yelouafi:
if you think we should just go with current one for now, it's okk for me

Andarist:
Either one is fine with me really, I just used this pattern because I know it works and I don't currently have time setup a different repo and test out a different pattern there before we could confidently say here it's fine to switch over ;p

yelouafi (approved):

PR #2727, opened 2026-05-06
Fix stack overflow from synchronous saga resumes

### LINKED ISSUE #1592: Saga stopped inside loop without throwing any error

yelouafi:
Fixes #1592

## Summary

This prevents deeply synchronous saga chains from recursively growing the JS call stack.

`proc` now detects synchronous resumes that happen while an effect is still being digested and stores the resume request for the outer `next()` loop to drain iteratively.

## Why

Some effects can complete synchronously and immediately call back into `next()`. With a long enough chain, that recursive resume path can overflow the stack.

A naive trampoline is too broad, though: it can delay parent continuation across scheduler boundaries and break ordering guarantees around `fork`/`put`/`take`.

The implementation therefore uses the scheduler semaphore to flatten safe same-stack resumes while preserving scheduler expectations.

A visual playground illustrating the scheduler-order cases and the naive trampoline failure mode is available here:
https://coruscating-snickerdoodle-3f49f8.netlify.app/

## Covered cases

- long synchronous `call(syncFn)` chains
- sync chains after an async boundary
- plain synchronous yielded values
- caught sync call errors
- cancellation during long sync chains
- child cancellation after a long sync prefix
- called sub-sagas with long sync chains
- repeated `all`/`race` sync completion paths
- deep fork chains
- fork/put ordering regressions

### DIFF
--- a/.changeset/sync-resume-stack-overflow.md
+++ b/.changeset/sync-resume-stack-overflow.md
@@ -1,0 +1,5 @@
+---
+'@redux-saga/core': patch
+---
+
+Fix stack overflow when sagas resume through deeply synchronous effect chains.
--- a/packages/core/__tests__/interpreter/stackoverflow.js
+++ b/packages/core/__tests__/interpreter/stackoverflow.js
@@ -1,0 +1,483 @@
+import { createStore, applyMiddleware } from 'redux'
+import { TASK_CANCEL } from '@redux-saga/symbols'
+import sagaMiddleware from '../../src'
+import * as io from '../../src/effects'
+
+test('saga handles many synchronous call effects without getting stuck', () => {
+  const callAttempts = 10000
+  let actualCallCount = 0
+  let timeoutId
+
+  const middleware = sagaMiddleware()
+  createStore(() => ({}), {}, applyMiddleware(middleware))
+
+  function syncCall() {
+    actualCallCount += 1
+  }
+
+  function* root() {
+    for (let i = 0; i < callAttempts; i++) {
+      yield io.call(syncCall)
+    }
+  }
+
+  const task = middleware.run(root)
+  const timeout = new Promise((_, reject) => {
+    timeoutId = setTimeout(() => {
+      reject(new Error(`Saga did not complete after ${actualCallCount} synchronous call effects`))
+      task.cancel()
+    }, 1000)
+  })
+
+  return Promise.race([task.toPromise(), timeout])
+    .then(() => {
+      expect(actualCallCount).toBe(callAttempts)
+    })
+    .finally(() => {
+      clearTimeout(timeoutId)
+    })
+})
+
+test('saga handles a long synchronous call chain after an asynchronous boundary', () => {
+  const callAttempts = 10000
+  let actualCallCount = 0
+  let timeoutId
+
+  const middleware = sagaMiddleware()
+  createStore(() => ({}), {}, applyMiddleware(middleware))
+
+  function syncCall() {
+    actualCallCount += 1
+  }
+
+  function* root() {
+    yield Promise.resolve()
+    for (let i = 0; i < callAttempts; i++) {
+      yield io.call(syncCall)
+    }
+  }
+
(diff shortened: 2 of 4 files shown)


Andarist:
Hah, this is basically the one local fix I had pending 😅 I put my version in [this PR](https://github.com/redux-saga/redux-saga/pull/2728) now for comparison purposes. It's just slightly different but also very-very similar. We should bash some agents against both implementations to see if there are any meaningful differences between both and pick one.

yelouafi:
Haha no way! even variable names!

I pointed codex at the PRs

I think cancellation point is very legit, the other are mostyl about hardening/avoiding surprises (I also merged tests in your PR with mine so we have broader coverage)

> The only concrete semantic difference I can point to right now is cancellation. In ours, a queued TASK_CANCEL immediately calls next.cancel() before storing the trampoline resume. That cancels the currently settling effect, so its normal sync completion can’t overwrite the cancellation. See proc.js (line 85).
> 
> In #2728, cancellation is just queued like any other resume. If task.cancel() happens inside the sync callback, the callback can then return normally and overwrite the queued cancel resume. That’s why Andarist’s test expects the loop to continue to 1499; our new test expects it to stop at the cancellation point.
> 

> What should happen semantically:
> 
> task.cancel() asks the running saga to resume with TASK_CANCEL.
> The saga should jump to finally.
> The currently running effect should be marked as cancelled/settled.
> When syncCall returns normally, that normal result should be ignored.
> 
> That last part is crucial.
> 
> In redux-saga, every effect has this invariant: completion and cancellation are mutually exclusive. That is guarded by effectSettled inside proc.js (line 210). When cancellation wins, next.cancel() marks the effect settled, so a later normal cb(result) is ignored.






> The other differences are not purely cosmetic, but I’d call them semantic boundaries / risk controls, not proven behavioral divergences yet:
> 
> 
> The semaphore guard: ours is narrower and documents the nested fork lock case; his broader digestingSemaphore > 0 may flatten more scheduler-locked situations.
> 
> The digest state restore: ours restores both digesting and digestingSemaphore; his always resets digesting = false. That feels like an invariant/robustness difference around nested re-entry.


> 
> Tests
> 
> His tests add explicit put, fork after async, and alternating put/call/fork cases.
> Ours has stronger stress coverage for sync call depth, raw yielded values, caught errors, generator throws, sub-sagas, all, race, cancellation, and fork/put ordering.
> I’d consider borrowing one or two of his happy-path mixed-effect tests, especially alternating put/call/fork, but I wouldn’t replace our stackoverflow.js coverage with his syncEffects.js.

Andarist (commented):
1. the agent feedback from [here](https://github.com/redux-saga/redux-saga/pull/2727#issuecomment-4387281021) has to be addressed (if needed)
2. this PR needs a small changeset to be properly released and attributed later on

yelouafi:
I already merged all tests from your PR not present in mine to ensure max coverage

Andarist (approved):

PR #2728, opened 2026-05-06
Fix stack overflow from synchronous saga resumes

Andarist:


Andarist:
superseded by https://github.com/redux-saga/redux-saga/pull/2727

PR #2732, opened 2026-05-11
Modernize ESLint configuration

yelouafi:
## Summary
- Migrate from legacy `.eslintrc.json` / `.eslintignore` to flat `eslint.config.js`.
- Upgrade ESLint from 4 to 10 and add flat-config helpers.
- Remove obsolete `babel-eslint` and React lint plugin usage from the current non-JSX lint target.

## Notes
- Keeps build/Babel/runtime behavior unchanged.
- Pins parser syntax support to `ecmaVersion: 2025`.

## Validation
- `yarn install --frozen-lockfile --non-interactive`
- `yarn lint`
- `yarn test`
- `yarn build`
- `git diff --check`


Andarist:
> Upgrade ESLint from 4 to 10 

😅

package.json
```diff
@@ -7,7 +7,7 @@
   ],
   "scripts": {
     "format": "prettier --write \"{packages,website,examples}/**/*.{js,jsx}\"",
-    "lint": "eslint packages/*/{src,test}",
+    "lint": "eslint packages/*/src packages/*/test",
```
Andarist, line 10: in the long run, it might be good to switch to oxlint - but certainly that doesn't have to be done now. ESLint is still solid

Andarist (approved):

package.json
yelouafi: yeah thought of oxlint, but for now just taking the obvious path

PR #2741, opened 2026-05-18
Modernize TypeScript type tests

yelouafi:
## Summary
- replace dtslint type checks with current TypeScript + NodeNext
- drop legacy pre-TS declaration branches
- allow `yield*` with `SagaIterator`

## Validation
- `yarn test:types`
- `yarn jest packages/testing-utils/__tests__/cloneableGenerator.js --runInBand`


packages/core/types/pre-ts3.6/channels.test.ts
Andarist, line 1: technically-ish this could be considered a breaking change, like dropping support for an old node runtime... but at this point those really are ancient so 🤷‍♂️

packages/core/types/middleware.test.ts
```diff
@@ -18,25 +18,25 @@ function testRun() {
 
   // TODO: https://github.com/Microsoft/TypeScript/issues/28803
   {
-    // // $ExpectError
+    // // @ts-expect-error
     // middleware.run(function* saga(a: 'a'): SagaIterator {})
```
Andarist: I wonder if this test could get enabled now

packages/core/types/middleware.test.ts
```diff
@@ -18,25 +18,25 @@ function testRun() {
...
+  // @ts-expect-error
   middleware.run(function* saga(a: 'a'): SagaIterator {}, 1)
 
   middleware.run(function* saga(a: 'a'): SagaIterator {}, 'a')
 
   // TODO: https://github.com/Microsoft/TypeScript/issues/28803
   {
-    // // $ExpectError
+    // // @ts-expect-error
     // middleware.run(function* saga(a: 'a', b: 'b'): SagaIterator {}, 'a')
```
Andarist: same here, could this be enabled now?

packages/core/types/runSaga.test.ts
```diff
@@ -68,25 +68,25 @@ function testRunSaga() {
 
   // TODO: https://github.com/Microsoft/TypeScript/issues/28803
   {
-    // // $ExpectError
+    // // @ts-expect-error
     // runSaga({}, function* saga(a: 'a'): SagaIterator {})
```
Andarist: and here ;p

packages/core/types/runSaga.test.ts
```diff
@@ -68,25 +68,25 @@ function testRunSaga() {
...
+  // @ts-expect-error
   runSaga({}, function* saga(a: 'a'): SagaIterator {}, 1)
 
   runSaga({}, function* saga(a: 'a'): SagaIterator {}, 'a')
 
   // TODO: https://github.com/Microsoft/TypeScript/issues/28803
   {
-    // // $ExpectError
+    // // @ts-expect-error
     // runSaga({}, function* saga(a: 'a', b: 'b'): SagaIterator {}, 'a')
```
Andarist: and here ;p

packages/core/types/tsconfig.json
```diff
@@ -1,13 +1,16 @@
...
-    "baseUrl": ".",
+    "types": [],
     "paths": {
       "redux-saga": ["./index.d.ts"],
-      "redux-saga/effects": ["./effects.d.ts"]
+      "redux-saga/effects": ["./effects.d.ts"],
+      "@redux-saga/core/effects": ["./effects.d.ts"],
+      "@redux-saga/types": ["../../types/types/index.d.ts"]
     }
   }
```
Andarist: hm, if you'd run TS tests *after* building package then it's likely those paths wouldn't be needed

packages/core/package.json
```diff
@@ -47,7 +47,7 @@
     }
   },
   "scripts": {
-    "test:types": "dtslint types"
```
Andarist, line 50: one cool advantage of dtslint was that it would test against multiple TS versions (including `@next`). We probably don't have to go as far here... but as a point of reference, this can also be achieved with a job matrix, like here: https://github.com/reduxjs/redux-toolkit/blob/b946e508db124de4d1faf68153b98bc12681d83c/.github/workflows/tests.yml#L129-L130

.changeset/modern-typescript-support.md
```diff
@@ -0,0 +1,7 @@
+---
+"@redux-saga/core": minor
+"@redux-saga/testing-utils": patch
+"@redux-saga/types": minor
+---
+
+Drop legacy TypeScript compatibility branches and validate declarations with the current TypeScript compiler using NodeNext module resolution. `SagaIterator` now uses `Generator` so sagas annotated with it can be composed with `yield*`, and cloneable generator test helpers now expose the iterable protocol.
```
Andarist: This mentions multiple things in a single changeset. Think of changesets as a simple substitution mechanism, those messages end up in CHANGELOGs of the described packages. 

It feels like not all of the described changes make sense from the perspective of all mentioned packages. The rules are: we can freely split this into multiple files (multiple changeset files are more than OK), when message reuse makes sense - we can also mention multiple packages in a single changeset file

packages/core/types/pre-ts3.6/channels.test.ts
yelouafi: yeah strictly should be major but felt too much for this kind of change

packages/core/package.json
yelouafi: maybe just define a a modern min ts version to support, and test against it; optionally addding next

packages/core/types/tsconfig.json
yelouafi: yah makes sense for ci; but locally it means locally we'd need to run build before test:types; which might be surprising?

Andarist: it feels it should just work without build locally (based on `preconstruct dev` alone) - but perhaps it doesn't work because our source code is not actually *authored* in TS. I'll look into improving this in Preconstruct

packages/core/package.json
Andarist: Yeah, our types are not that fancy - so we don't need to test them against multiple versions right now. 

FWIW, DefinitelyTyped defines this support window: https://github.com/DefinitelyTyped/DefinitelyTyped/#support-window

packages/core/types/tsconfig.json
Andarist: actually, Preconstruct already handles this quite OK based on my testing - those `paths` aliases feel redundant as thigns "just work" right now without them

Andarist (approved):

PR #2757, opened 2026-07-02
Fix babel plugin crash when yielding primitive values

### LINKED ISSUE #2088: babel-plugin-redux-saga crashes sagas - Object.defineProperty called on non-object
### Steps to reproduce
yield something that isn't an object.

 The plugin should check that that it transforming an object before calling defineProperty.

 TypeError: Object.defineProperty called on non-object

#### Example
```
// I know there is nothing to yield here, but this would normaly be fine without the plugin 
function* foo() {
  yield 20; //  crash: Object.defineProperty called on non-object
}
```

Zish19:
Fixes #2088.

## Summary

This PR fixes a runtime crash in `babel-plugin-redux-saga` when a yielded call or logical expression evaluates to a primitive value.

The fix introduces a shared hoisted helper that:

* Evaluates the yielded expression exactly once.
* Attaches location metadata only for non-null objects and functions.
* Returns the original value unchanged.
* Avoids duplicating the runtime guard at every yield site.

## Tests

* Added an `effect-primitive-yield` fixture.
* Added runtime tests covering numbers, strings, booleans, `null`, `undefined`, and objects.
* Regenerated Babel 6 and Babel 7 expected fixtures.

## Cross-platform

* Normalized generated paths to POSIX separators.
* Removed CRLF-related snapshot differences.
* Restored the `{{absolutePath}}` placeholder in fixtures.
* Regenerated snapshots for deterministic output.

## Release

Adds a patch changeset for `babel-plugin-redux-saga`.

### DIFF
--- a/packages/babel-plugin-redux-saga/src/index.js
+++ b/packages/babel-plugin-redux-saga/src/index.js
@@ -6,16 +6,16 @@ var symbolName = '@@redux-saga/LOCATION'
 function getSourceCode(path) {
   // use `toString` for babel v7, `getSource` for older versions
   const rawCode = Object.prototype.hasOwnProperty.call(path, 'toString') ? path.toString() : path.getSource()
-  return rawCode.replace(/^(yield\*?)\s+/, '')
+  return rawCode.replace(/^(yield\*?)\s+/, '').replace(/\r/g, '')
 }
 
 function getFilename(fileOptions, useAbsolutePath) {
   if (useAbsolutePath) {
-    return fileOptions.filename
+    return fileOptions.filename.replace(/\\/g, '/')
   }
   // babel v7 defines cwd. for v6 use fallback
   const cwd = fileOptions.cwd || fileOptions.sourceRoot || process.cwd()
-  return pathFS.relative(cwd, fileOptions.filename)
+  return pathFS.relative(cwd, fileOptions.filename).replace(/\\/g, '/')
 }
 
 function isSaga(path) {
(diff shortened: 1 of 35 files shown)


Zish19:
Pushed a cleanup commit to ensure deterministic snapshot generation across operating systems. This normalizes path separators to POSIX style, removes CRLF-related snapshot noise, and restores the absolute-path fixture placeholder. The remaining fixture updates are only due to the intended yield instrumentation change.

packages/babel-plugin-redux-saga/src/index.js
```diff
@@ -37,8 +37,25 @@ module.exports = function (babel) {
     });
   `)
 
+  // Safe template for yield expressions: wraps Object.defineProperty in
+  // a runtime guard so that primitive return values don't crash.
+  var safeExtendExpressionWithLocationTemplate = template(`
+    (function (value) {
```
Andarist: this should be a shared hoisted helper and not an IIFE inserted at each yield site

Zish19:
That makes sense, the current IIFE approach avoids repeated evaluation but does duplicate the guard logic at every yield site.
I’ll refactor this to use a shared hoisted helper instead, so the runtime guard is emitted once and yield sites call into that helper.

Zish19:
Hi @Andarist , I addressed the review feedback by replacing the per-yield IIFE with a shared hoisted helper and updated the tests accordingly. When you have a chance, I'd appreciate another look to see if this implementation matches what you had in mind. Thanks!

yelouafi (approved):

PR #2742, opened 2026-05-19
Switch to pnpm and harden CI

Andarist:

PR #2764, opened 2026-07-23
Migrate workspace from Yarn to pnpm

yelouafi:
> **Scope and relationship to #2742**
>
> PR #2742 combined a pnpm migration with broader CI and GitHub Actions hardening. This PR intentionally extracts the pnpm migration as the first independently reviewable step. It includes only the CI changes needed to support pnpm and the updated Node version; the remaining CI/security hardening will follow in separate, focused PRs.

## Summary

- migrate the monorepo from Yarn Classic to pnpm 11 with a single workspace lockfile
- move the existing Preconstruct patch to pnpm's native patch mechanism and declare dependencies that previously relied on Yarn hoisting
- migrate Husky 1 hooks to Husky 9's tracked, pnpm-compatible hook setup while preserving the existing lint-staged checks
- update contributor docs, examples, release scripts, and GitHub Actions for pnpm and Node 24
- add pnpm supply-chain controls for dependency scripts, release age, provenance downgrade, lockfile validation, and exotic transitive sources
- update Pages deployment to pinned `peaceiris/actions-gh-pages` v4.1.0 while deferring the artifact-based Pages migration to a follow-up hardening PR

## Why

Yarn Classic's hoisted install model and install-script defaults make dependency relationships and supply-chain policy less explicit. This migration gives the repository a deterministic workspace setup, exact package-manager pinning, native patch support, and explicit install-time security policy. Node 24 is used because Node 20 is EOL and pnpm 11 requires Node 22 or newer.

## Impact

Contributors need Node 22 or newer; CI uses Node 24. Common commands are now `pnpm install`, `pnpm run ...`, and `pnpm --dir ...`. Published package runtime compatibility is unchanged; the Node requirement applies to repository tooling.

## Validation

- clean Linux/Node 24 frozen install with pnpm 11.16.0
- pnpm supply-chain validation of 2,186 lockfile entries
- ESLint
- package build and TypeScript type tests
- Jest: 39 suites, 223 tests, 3 snapshots
- Docusaurus production build
- actionlint 1.7.12
- normal commit and push through the Husky 9 hook setup without bypass flags
- Prettier and `git diff --check`
- zizmor audit (remaining findings match the deferred broader Actions-hardening scope)


packages/core/package.json
```diff
@@ -92,6 +92,7 @@
     "jest": "^29.7.0",
     "mitt": "^1.1.2",
     "redux": "^4.0.4",
+    "redux-saga": "^1.5.0",
```
Andarist: this one looks like bringing circularity to the table, why do we need this? it feels like it should be avoidable

.github/workflows/publish.yml
```diff
@@ -23,14 +23,17 @@ jobs:
...
 
+      - name: Install pnpm
+        uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
+
       - name: Use Node.js 24
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
           node-version: 24
-          cache: yarn
+          cache: pnpm
```
Andarist, line 36: supposedly github already fixed the possibility of those cache poisoning attacks but perhaps we should continue skipping the cache in this publish workflow

.github/workflows/publish.yml
```diff
@@ -23,14 +23,17 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
+      - name: Install pnpm
```
Andarist, line 29: in the other PR, I created a shared action for the common setup: https://github.com/redux-saga/redux-saga/pull/2742/changes#diff-ffe2818c1fb7fe72e6f88b6bb008825e0ce4b42a6af745e4e68d6f7bbe33f998

i feel it's nicer in the long run to have such but I don't mind having this stuff duplicated either

pnpm-workspace.yaml
```diff
@@ -0,0 +1,45 @@
...
+  # Unlisted dependency scripts are blocked as unreviewed; false records that
+  # these scripts were reviewed and intentionally denied.
+  '@parcel/watcher': false
+  core-js: false
+  fsevents: false
+
+autoInstallPeers: false
+blockExoticSubdeps: true
+dedupePeers: false
+dedupePeerDependents: false
```
Andarist, line 16: any particular reaosn why those 2 are set to false and not true?

.github/workflows/publish.yml
Andarist, line 1: this workflow should use:
```yml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: false
  queue: max
```

Andarist (approved):

packages/core/package.json
yelouafi: yeah you're right; the agent added this because core type tests import from redux-saga, seems Yarn hoisting made it implictly available; will move to root dev dependencies

.github/workflows/publish.yml
yelouafi: I intentionally made PR about pnpm only, introducing minimum necessary changes to CI to make pnpm work; will introcuce CI in a followup PR

pnpm-workspace.yaml
yelouafi: mm, it was added because examples & website use different versions of react & webpack, and setting esp `dedupePeerDependents: true` mixes webpack contexts; `dedupePeers: true` is possible but per local test provides little in our case (about 489 changed lockfile lines)

Andarist: it's fine then

.github/workflows/publish.yml
yelouafi: fixed

